PR #65, opened 2026-05-29
fix(auth): trim bearer tokens in rate-limit keys

pekar9781:
## Summary

Use `extractBearerToken` when building rate-limit keys and audit actor hints so bearer credentials match private API authentication.

Closes #63

## Validation

- [x] `npm run test:ci` (auth unit tests)
- [ ] Changelog updated only if this is a release-prep change

## Safety

- [x] Backend-only change
- [x] No secrets, wallet details, user PATs, raw trust scores, or private rankings exposed
- [x] Public text avoids compensation-seeking or optimization-tactic language
- [ ] OpenAPI/MCP behavior updated where needed
- [ ] Public docs/changelogs updated where needed


JSONbored (changes requested):
@jony376 the underlying fix is valid, but this is not merge-ready yet.

Required:
- Strengthen the test so it exercises the actual rate-limit key/enforcement path with padded and unpadded `Authorization: Bearer ...` headers, not only `extractBearerToken()`/`hashToken()` directly.
- Sync/rebase with current `main` and fix the failing `validate` run. Current failure is in readiness/SLO integration assertions.
- Keep the fix scoped to `src/auth/rate-limit.ts` and auth/rate-limit tests.

pekar9781:
Hi, @JSONbored I fixed conflict and CI error. Please check this PR. Thanks.

JSONbored (changes requested):
@jony376 The implementation direction is right, but the test still does not cover the actual rate-limit identity/enforcement path. Please add a behavioral test proving padded and unpadded `Authorization: Bearer ...` headers produce the same limiter identity through the rate-limit path itself, not only through `extractBearerToken()`/`hashToken()` directly.

JSONbored (changes requested):
@jony376 the implementation direction is right: routing bearer parsing through `extractBearerToken()` removes the whitespace mismatch in both `rateLimitKey()` and `actorHint()`.

A few notes:
- The current added test proves trimmed helper output hashes the same.
- The maintainer concern is still the runtime path: `enforceRateLimit()` should produce the same rate-limit identity for padded and unpadded bearer headers.
- CI is green, so this is now a coverage/behavior proof issue rather than a build issue.

Required changes:
- Add a behavioral test through `enforceRateLimit()` using `Authorization: Bearer session-token` and `Authorization: Bearer  session-token  `.
- Assert they hit the same limiter key/bucket, for example by capturing the Durable Object name/key or by proving the second request is rate-limited when the limit is 1.

Validation expected:
- Re-run `npm run test:coverage`.
- Keep the auth tests green without weakening existing rate-limit coverage.

JSONbored:
@jony376 - this was a valid issue, however it was already resolved through a security advisory yesterday, so I'm going to close this. Getting it fixed was more important than waiting for an update on this PR, unfortunately, and the advisory was more extensive in its resolution.

A few notes:
- The original whitespace mismatch was valid, and your branch moved in the right direction by using the shared bearer extraction helper.
- Current `main` now contains the broader advisory fix: invalid bearer headers fall back to IP-based buckets, and valid bearer tokens are checked before token bucket identity is used.
- Keeping this branch active would reintroduce stale conflicts around the security fix that is already present on `main`.

PR #161, opened 2026-05-31
fix(labels): ensurePullRequestLabel misses labels beyond first page

### LINKED ISSUE #160: [Bug]: ensurePullRequestLabel misses labels beyond first page
# [Bug]: ensurePullRequestLabel fails when label exists beyond first 100 repo labels

## Summary

`ensurePullRequestLabel` lists only the first page of repository labels. If the target label exists on a later page, creation is attempted, GitHub returns 422 already exists, and the label is never applied to the PR.

## Steps to reproduce

1. Repository with >100 labels including `gittensor` not on page 1.
2. Run auto-label on a PR without the label applied.

## Expected behavior

Label applied via `POST .../issues/{n}/labels`.

## Actual behavior

Function errors or skips apply after failed create.

## Impact

Auto-labeling fails on large active repos.

glorysr1209-png:
## Summary

Closes #160.

Look up repository labels by name instead of scanning only the first labels page.

## Changes

- `GET /repos/{owner}/{repo}/labels/{name}` with 404 → create when allowed.
- Update label unit tests for direct lookup.

## Test plan

- [ ] `npm test -- test/unit/github-labels.test.ts`


JSONbored (changes requested):
@glorysr1209-png The direct label lookup is a good API-level fix, but the branch currently breaks webhook accounting.

A few notes:
- Switching to `GET /labels/{name}` avoids the first-page-only repository-label lookup problem.
- CI now fails because the queue webhook test expected `labelsCreated: 1` and received `labelsCreated: 0`.
- That means the missing-label creation path is no longer reflected correctly in the visible queue stats.

Required changes:
- Preserve direct lookup for existing labels.
- Preserve missing-label creation and `labelsCreated` accounting when the label does not exist.
- Keep both cases covered in tests.

Validation expected:
- `npm run test:coverage`

JSONbored (changes requested):
@glorysr1209-png this is still blocked by CI.

A few notes:
- Direct label lookup by name is a reasonable fix for first-page pagination misses.
- Current validate fails in `test/unit/queue.test.ts` because `labelsCreated` is `0` where the test expects `1`.
- The body needs current validation/safety evidence.

Required changes:
- Fix the label-create/lookup behavior or update the test only if the expected behavior has intentionally changed.
- Update the PR body to the current template.

Validation expected:
- Focused label tests
- Full validate pipeline

JSONbored:
This is now stale/have been awaiting changes for a few days with no response. Please feel free to try again based on current codebase status if interested.

PR #164, opened 2026-05-31
fix(signals): detect preflight duplicate work by symmetric term overlap

### LINKED ISSUE #163: [Bug]: Preflight `possible_duplicate_work` is missed because overlap uses a one-direction substring match
### Summary

For a normal contribution, the preflight duplicate/overlap check fails to fire because it tests overlap with `item.title.toLowerCase().includes(input.title.toLowerCase())` — true **only** when an existing issue/PR title contains the *entire* planned PR title.

A planned PR title is normally longer and more descriptive than the duplicate it overlaps, so the substring is almost never satisfied and `possible_duplicate_work` is **silently suppressed**. The check is also direction-asymmetric in the wrong direction: a very short planned title spuriously matches unrelated open work. The engine already exposes the correct primitive — `buildCollisionReport` clusters items by `termOverlap` (shared tokenized terms) — but `buildPreflightResult` does not use it for the planned contribution.

### Area

REST API

### Expected behavior

`possible_duplicate_work` is reported whenever the planned contribution meaningfully overlaps an existing open-work cluster, regardless of which title is longer.

### Actual behavior

Overlap is detected only when an existing item's title is a superstring of the whole planned title, so genuine duplicates are missed; a one-word planned title also produces false positives.

### Validation

_No response_

joaovictor712:
Fixes #163.

## Summary
`buildPreflightResult` detected duplicate/overlapping work with `item.title.toLowerCase().includes(input.title.toLowerCase())` - true only when an existing issue/PR title contains the **entire** planned PR title. A normal (longer, more descriptive) planned title therefore never matched a shorter duplicate, so `possible_duplicate_work` was silently suppressed across the MCP/API preflight, the public PR-intelligence comment, and the `@gittensory preflight` / `duplicate-check` commands; a one-word planned title spuriously matched unrelated work.

## Fix
Compare the planned contribution to each collision item with the same direction-independent term-overlap heuristic `buildCollisionReport` already uses between items (`termOverlap`, >=2 shared meaningful terms), via a new `plannedContributionTerms` helper that reuses `tokenize`. The `linkedIssues` short-circuit is kept.

## Tests
Two fail-on-revert regression tests in `test/unit/signals-coverage.test.ts`:
- a longer planned title overlapping a 2-item cluster -> `possible_duplicate_work` is present;
- a one-word planned title sharing a single word -> finding is absent.

Restoring the old `includes` check fails both.

## Verification
`npm run typecheck` clean; `npx vitest run test/unit/signals-coverage.test.ts` -> 10 passed; `test/unit` green aside from one unrelated `mcp-cli` cold-start timeout that passes on re-run.


JSONbored (approved):
@galuis116 This looks ready from my side.

A few notes:
- The symmetric term-overlap approach is the right fix for the previous one-way substring check.
- The two tests cover both sides of the regression: a longer planned title now matches overlapping open work, and a short one-word title no longer creates noisy duplicate warnings.
- Current `validate` is green and the merge-tree is clean.

No requested changes from me.

PR #181, opened 2026-06-01
Fix saturation-model contribution bonus capped at 5 instead of MAX_CONTRIBUTION_BONUS (25)

### LINKED ISSUE #180: [Bug]: Saturation-model `contributionBonus` is hard-capped at **5** instead of `MAX_CONTRIBUTION_BONUS` (25), understating every active-model score preview
## Summary

In `src/scoring/preview.ts`, the **active** scoring model (`pending_saturation_model`) computes its token-contribution bonus through `saturationContributionBonus`, which caps the bonus at a literal **5** — both via a wrong fallback default and via a redundant `Math.min(..., 5)`:

```ts
// src/scoring/preview.ts:561
function saturationContributionBonus(totalTokenScore: number, constants: Record<string, number>): number {
  const contributionBonusCap = Math.min(constant(constants, "MAX_CONTRIBUTION_BONUS", 5), 5);
  return clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) * contributionBonusCap;
}
```

The identical concept — the contribution bonus ramp `clamp(totalTokenScore / CONTRIBUTION_SCORE_FOR_FULL_BONUS, 0, 1) * MAX_CONTRIBUTION_BONUS` — is computed **correctly** in the density-model sibling a few lines above, scaling by the full `MAX_CONTRIBUTION_BONUS` (default **25**):

```ts
// src/scoring/preview.ts:214
const densityContributionBonus =
  clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) *
  constant(constants, "MAX_CONTRIBUTION_BONUS", 25);
```

Two compounding defects pin the saturation bonus to ≤ 5:

1. **Wrong fallback default.** `constant(constants, "MAX_CONTRIBUTION_BONUS", 5)` uses `5`, but the canonical default for this constant is **25** (`src/scoring/model.ts:14` → `MAX_CONTRIBUTION_BONUS: 25`, and the density twin at `preview.ts:216` uses `25`).
2. **Redundant hard cap.** Even when the real constant `25` is present (which it is by default), `Math.min(25, 5)` forces it back down to `5`. There is no separate "saturation bonus cap" constant in the model — `MAX_CONTRIBUTION_BONUS` *is* the cap, so re-clamping it against a literal `5` is nonsensical.

`MAX_CONTRIBUTION_BONUS` is the **only** contribution-bonus constant in `DEFAULT_SCORING_CONSTANTS` (`src/scoring/model.ts:9-33`); nothing else defines a `5`-point ceiling, so the literal `5` has no config backing.

### Why this is the *active* path, not a dormant branch

The saturation model is the default active model. `detectActiveModel` returns `pending_saturation_model` whenever `SRC_TOK_SATURATION_SCALE` is finite:

```ts
// src/scoring/model.ts:107
export function detectActiveModel(constants) {
  if (hasSaturationConstants(constants)) return "pending_saturation_model";  // SRC_TOK_SATURATION_SCALE finite
  ...
}
// src/scoring/model.ts:32 — SRC_TOK_SATURATION_SCALE: 58 is in DEFAULT_SCORING_CONSTANTS
```

So `baseScore = saturationBaseScore` (`preview.ts:224-225`) and `contributionBonus = saturationContributionBonusValue` (`preview.ts:227`) for every preview under the default constants. The bug is in the live scoring path, not a future/disabled model.

Additionally, `pendingSaturationScore` / `underlyingPotentialScore` is **always** derived from the saturation function regardless of the active model (`preview.ts:243` `pendingSaturationScore = roundScore(saturationBaseScore)`, `preview.ts:128` `underlyingPotentialScore = current.scoreEstimate.pendingSaturationScore`), so even density-model previews surface an understated "underlying potential" projection.

A high-token contribution at full bonus: `sourceTokenScore = 58`, `totalTokenScore = 1500` (= `CONTRIBUTION_SCORE_FOR_FULL_BONUS`, so the ramp is `1.0`), `credibility = 1`, `openPrCount = 0`, with the canonical constants (`MERGED_PR_BASE_SCORE = 25`, `SRC_TOK_SATURATION_SCALE = 58`, `MAX_CONTRIBUTION_BONUS = 25`).

The exponential term is `25 * (1 - e^(-58/58)) = 25 * 0.63212 = 15.803`. The contribution bonus ramp is `1.0`, so:

| Quantity | Current (capped at 5) | Correct (× `MAX_CONTRIBUTION_BONUS` = 25) |
|---|---|---|
| `contributionBonus` | `1.0 × min(25, 5) = `**`5`** | `1.0 × 25 = `**`25`** |
| `saturationBaseScore` (`baseScore`) | `15.803 + 5 = `**`20.803`** | `15.803 + 25 = `**`40.803`** |
| `estimatedMergedScore` (labels `["bug"]`→1.2, issue `standard`→1.33) | `20.803 × 1.596 = `**`33.20`** | `40.803 × 1.596 = `**`65.12`** |
| `pendingSaturationScore` / `underlyingPotentialScore` | **`20.803`** | **`40.803`** |

The bonus is understated by **20 points at the base-score level** (`5` vs `25`), which propagates multiplicatively into `estimatedMergedScore` — here roughly **half** of the rules-correct value.

### Recommendation-level threshold crossing

With label/issue multipliers at their defaults (`1.0` each), the 20-point base-score gap lands directly on `effectiveEstimatedScore`, which gates the fit recommendation (`preview.ts:166-168`): `>= 30 → strong_fit`, `>= 15 → reasonable_fit`. The same full-bonus contribution above (no label/issue boost) yields:

- **Correct:** `effectiveEstimatedScore = 40.803` → **`strong_fit`**.
- **Current:** `effectiveEstimatedScore = 20.803` → **`reasonable_fit`**.

A high-value, high-token contribution that should read as a strong fit is silently downgraded.

## Downstream effects

- **`estimatedMergedScore` / `effectiveEstimatedScore`** are understated for every active (saturation) preview — these feed the public score preview, `@gittensor` score commands, and MCP responses.
- **`recommendation.level`** (`preview.ts:166-168`) is biased toward weaker tiers, exactly for the largest contributions (which sit near the full-bonus ceiling where the missing 20 points matters most).
- **Opportunity ranking** in `src/mcp/server.ts:771` sorts by `effectiveEstimatedScore`; a uniform-but-nonlinear understatement reorders ranked opportunities.
- **`underlyingPotentialScore`** (shown in the scoreability summary at `preview.ts:521-522` and in `src/services/agent-orchestrator.ts:406`) is understated even under the density model.

## Steps to reproduce

1. Build a `ScoringModelSnapshotRecord` with `activeModel: "pending_saturation_model"` and `MAX_CONTRIBUTION_BONUS: 25`, `SRC_TOK_SATURATION_SCALE: 58` (the defaults).
2. Call `buildScorePreview` with `sourceTokenScore: 58`, `totalTokenScore: 1500`, `credibility: 1`, `openPrCount: 0`.
3. Read `preview.scoreEstimate.contributionBonus`.
4. Observe **`5`** (capped) instead of the rules-correct **`25`**, and `baseScore` `20.803` instead of `40.803`.

## Expected behavior

The saturation-model contribution bonus scales by the full `MAX_CONTRIBUTION_BONUS` constant (default 25), identical to the density-model bonus at `preview.ts:214-216`. For a full-bonus contribution the bonus is `25`, not `5`.

## Actual behavior

`saturationContributionBonus` caps the bonus at a literal `5` — via both a `5` fallback default and a redundant `Math.min(..., 5)` — so the bonus is understated by up to 20 points (and the whole estimate proportionally) for every preview under the active saturation model.

## Suggested fix

- In `saturationContributionBonus` (`preview.ts:561-563`), drop the literal `5` cap and the wrong default; scale by `MAX_CONTRIBUTION_BONUS` exactly as the density twin does:

  ```ts
  function saturationContributionBonus(totalTokenScore: number, constants: Record<string, number>): number {
    return (
      clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) *
      constant(constants, "MAX_CONTRIBUTION_BONUS", 25)
    );
  }
  ```

  (Optionally factor the shared ramp into one helper used by both `densityContributionBonus` and `saturationContributionBonus` so the two models cannot drift again.)

- Fix the test that currently **ratifies** the bug. `test/unit/scoring.test.ts:132` asserts `contributionBonus` `toBe(5)` for the saturation model while the sibling density test at `:162` asserts `toBe(25)` for the *same* inputs and the *same* `MAX_CONTRIBUTION_BONUS: 25` — so the suite encodes the inversion rather than catching it. Update the saturation expectations to the corrected values (`contributionBonus` `25`, `baseScore` `40.803`, `estimatedMergedScore` ≈ `65.12`, `pendingSaturationScore` `40.803`).

- Add fail-on-revert coverage: for identical inputs and the same `MAX_CONTRIBUTION_BONUS`, the saturation-model `contributionBonus` must equal the density-model `contributionBonus` (both `25` at full ramp), and a full-bonus saturation preview must not fall below the `strong_fit` threshold solely because of the contribution bonus.

joaovictor712:
Closes #180.

## Problem
`saturationContributionBonus` (the active `pending_saturation_model` path) capped the token-contribution bonus at a literal `5`, via both a wrong fallback default (`MAX_CONTRIBUTION_BONUS` defaulted to `5`) and a redundant `Math.min(..., 5)`. Its density-model twin correctly scales by the full `MAX_CONTRIBUTION_BONUS` (25), so identical inputs produced a bonus of `5` under saturation vs `25` under density — understating `baseScore`, `estimatedMergedScore`, and `underlyingPotentialScore` for every active-model preview (e.g. a full-bonus contribution scored 20.8 instead of 40.8, dropping from `strong_fit` to `reasonable_fit`).

## Fix
- Factor the contribution-bonus ramp into a single shared `contributionBonusRamp` helper used by both models, so they cannot drift.
- Scale by `MAX_CONTRIBUTION_BONUS` (default 25) in both paths; drop the `5` cap and wrong default.

## Tests
- Correct the saturation-model test that ratified the bug (`contributionBonus` 5 -> 25, `baseScore` 20.803 -> 40.803, `estimatedMergedScore` 33.2016 -> 65.1216).
- Update the density-era projection test (`pendingSaturationScore`/`underlyingPotentialScore` now 40.803).
- Add coverage asserting the saturation and density bonuses are equal for identical inputs, and that a full-bonus contribution stays at/above the `strong_fit` threshold.

`vitest run test/unit/scoring.test.ts` passes (13/13); `tsc --noEmit` clean.

### DIFF
--- a/src/scoring/preview.ts
+++ b/src/scoring/preview.ts
@@ -211,9 +211,7 @@ function computeScoreCore(
   const densityMultiplier = clamp(rawDensity || 0, 0, constant(constants, "MAX_CODE_DENSITY_MULTIPLIER", 1.15));
   const densityTokenGatePassed = sourceTokenScore >= constant(constants, "MIN_TOKEN_SCORE_FOR_BASE_SCORE", 5);
   const baseTokenGatePassed = snapshot.activeModel === "pending_saturation_model" ? sourceTokenScore > 0 : densityTokenGatePassed;
-  const densityContributionBonus =
-    clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) *
-    constant(constants, "MAX_CONTRIBUTION_BONUS", 25);
+  const densityContributionBonus = contributionBonusRamp(totalTokenScore, constants);
   const saturationContributionBonusValue = saturationContributionBonus(totalTokenScore, constants);
   const saturationBaseScore = saturationScore(sourceTokenScore, totalTokenScore, constants);
   const densityBaseScore =
@@ -559,8 +557,17 @@ function saturationScore(sourceTokenScore: number, totalTokenScore: number, cons
 }
 
 function saturationContributionBonus(totalTokenScore: number, constants: Record<string, number>): number {
-  const contributionBonusCap = Math.min(constant(constants, "MAX_CONTRIBUTION_BONUS", 5), 5);
-  return clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) * contributionBonusCap;
+  return contributionBonusRamp(totalTokenScore, constants);
+}
+
+// Shared contribution-bonus ramp used by both scoring models so the saturation
+// and density bonuses cannot drift: clamp(totalTokenScore / FULL_BONUS, 0, 1)
+// scaled by MAX_CONTRIBUTION_BONUS (default 25).
+function contributionBonusRamp(totalTokenScore: number, constants: Record<string, number>): number {
+  return (
+    clamp(totalTokenScore / constant(constants, "CONTRIBUTION_SCORE_FOR_FULL_BONUS", 1500), 0, 1) *
+    constant(constants, "MAX_CONTRIBUTION_BONUS", 25)
+  );
 }
 
 function nonNegative(value: number | undefined): number {
--- a/test/unit/scoring.test.ts
+++ b/test/unit/scoring.test.ts
@@ -143,15 +143,15 @@ MAX_CODE_DENSITY_MULTIPLIER = 1.15
     });
 
     expect(preview.activeModel).toBe("pending_saturation_model");
-    expect(preview.scoreEstimate.baseScore).toBeCloseTo(20.803, 3);
-    expect(preview.scoreEstimate.contributionBonus).toBe(5);
+    expect(preview.scoreEstimate.baseScore).toBeCloseTo(40.803, 3);
+    expect(preview.scoreEstimate.contributionBonus).toBe(25);
     expect(preview.scoreEstimate.pendingSaturationScore).toBe(preview.scoreEstimate.baseScore);
-    expect(preview.scoreEstimate.estimatedMergedScore).toBeCloseTo(33.2016, 3);
+    expect(preview.scoreEstimate.estimatedMergedScore).toBeCloseTo(65.1216, 3);
     expect(preview.gates.baseTokenGatePassed).toBe(true);
     expect(JSON.stringify(preview.scoreEstimate)).not.toMatch(/reward estimate|wallet|hotkey|farming|payout/i);
   });
 
-  it("keeps pending saturation projection bonus capped for density-era snapshots", () => {
+  it("projects the saturation-model score with the full contribution bonus for density-era snapshots", () => {
     const densitySnapshot: ScoringModelSnapshotRecord = {
       ...snapshot,
       activeModel: "current_density_model",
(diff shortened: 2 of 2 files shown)


JSONbored (approved):
@galuis116 This is ready from my side. The fix is small and targeted: saturation contribution bonus now uses the configured max instead of the lower saturation value, and the regression coverage locks the intended ceiling. Current checks are green and the merge tree is clean.

PR #394, opened 2026-06-04
fix(signals): rank contributor opportunities before the per-repo cap

### LINKED ISSUE #393: [Bug]: Contributor opportunities cap each repo to its first 5 issues in raw DB order, discarding the issue-quality ranking — high-quality "ready" issues past index 5 are never surfaced
## Summary

`buildContributorOpportunities` (`src/signals/engine.ts:1219-1309`) selects, per registered repo, the issues to surface as contribution opportunities by taking the **first 5** of an **unsorted** list:

```ts
// src/signals/engine.ts:1240-1249
const availableIssues = repoIssues.filter((issue) => issue.linkedPrs.length === 0 && !linkedIssueNumbers.has(issue.number));
// ...
const rankable = qualityByIssue
  ? availableIssues.filter((issue) => qualityByIssue.get(issue.number)?.status !== "do_not_use")
  : availableIssues;
for (const issue of rankable.slice(0, 5)) {           // <- first 5 in raw DB order, NOT ranked
  const quality = qualityByIssue?.get(issue.number);
  // ...computes a per-issue opportunity `score` (lane fit + label fit + qualityAdjustment ...)
  opportunities.push({ ..., score, ... });
}
```

`availableIssues` is `repoIssues.filter(...)`, i.e. whatever order the issues arrive from the DB (effectively by ingestion / number, not by relevance). The variable is named `rankable`, but it is **never ranked before `.slice(0, 5)`**. Only the ~5 issues that happen to sit first in DB order are ever scored and pushed; issues at index 5+ are silently dropped before the per-issue score is ever computed.

## Why this is wrong

The issue-quality producer goes to real effort to rank issues. `buildIssueQualityReport` (`engine.ts:2671-2744`) computes a per-issue `status`/`score` and returns the issues **sorted best-first**:

```ts
// src/signals/engine.ts:2723, 2736
const score = clamp(100 - warnings.length * 18 + reasons.length * 5 - (age > 180 ? 15 : 0), 0, 100);
// ...
.sort((left, right) => right.score - left.score || left.number - right.number);   // ready/high-score issues first
```

The consumer throws that ordering away: it keys the report by issue number (`engine.ts:1244`) for per-issue lookups, but selects which issues to surface by raw DB order, then caps to 5. So the quality signal influences the *score of whichever 5 issues were first*, but not *which 5 issues are chosen*. A "ready" issue that the producer ranked #1 is invisible if it happens to be the 6th `availableIssue` in DB order.

The per-issue `score` the consumer itself computes (`engine.ts:1265-1277`, combining lane fit, label-history overlap, queue/bounty penalties, and the quality adjustment) is the value the whole pipeline ultimately ranks on — opportunities are globally sorted by it at the end (`engine.ts:1308`). But that score is only ever computed for the arbitrary first-5 per repo, so the global ranking is drawn from a biased sample: the genuinely best-fit issue for a contributor in a repo never enters the candidate set unless it is in the repo's first 5 by DB order.

A registered repo has 30 open, unlinked issues. By DB order the first 5 are all thin/stale (`status: "hold"`/`"needs_proof"`), while issue `#742` at array index 12 is `status: "ready"` with strong label overlap to the contributor's history.

- **Current:** the loop scores only the first 5; `#742` is never turned into an opportunity. The contributor's opportunity list (and the decision pack built on top of it) surfaces 5 weak issues and omits the best match. `reasons` like "Issue quality report rates this issue as ready." can never appear for `#742`.
- **Correct:** the repo's best-scoring issues (including the "ready", high-label-fit `#742`) are the ones surfaced and ranked.

This directly steers miners toward lower-fit issues and hides the highest-value work the system actually identified.

## Steps to reproduce

1. Build contributor opportunities (`buildContributorOpportunities`, reached via `buildContributorFit` → `buildAndPersistContributorDecisionPack` and `queue/processors.ts:370`) for a registered repo with more than 5 open unlinked issues whose DB order does not match quality/fit order.
2. Put a high-quality (`ready`) issue beyond array index 5.
3. Observe it is absent from the returned `opportunities`, while lower-scoring issues from the first 5 are present.

## Expected behavior

The per-repo cap should select the **best** issues, not the **first** ones: rank `rankable` by the opportunity score (which already incorporates lane fit, label history, and the quality adjustment) and keep the top 5 per repo, consistent with the producer's score-descending ordering and with the global top-25 sort applied afterward.

## Actual behavior

`rankable.slice(0, 5)` takes the first 5 issues in raw DB order, so only those are scored and considered; higher-scoring / quality-`ready` issues beyond index 5 are dropped before scoring and never surface.

## Suggested fix

Score all `rankable` issues per repo, then keep that repo's top 5 by the computed opportunity score, instead of capping the raw list to 5 before scoring:

- Build the opportunity objects (with their `score`) for every `rankable` issue (still skipping historical-bounty issues via the existing `continue`).
- Sort that repo's opportunities by `score` descending (tie-break on issue number, mirroring the producer) and take the top 5.
- Push those into the global list, which is then globally sorted and capped to 25 as today.

This makes the cap select the best-fit issues, restores the value of the issue-quality ranking, and also improves the no-quality-report path (which currently also samples the first 5 by DB order rather than by lane/label fit). Add fail-on-revert coverage: a repo whose highest-scoring / `ready` issue sits beyond index 5 must appear in the returned opportunities and outrank the weaker first-5 issues.

joaovictor712:
## Summary
`buildContributorOpportunities` (`src/signals/engine.ts`) selected the issues to surface per repo by taking the **first 5** of an **unsorted** list (`rankable.slice(0, 5)`, raw DB order), then scoring only those. The list is named `rankable` but was never ranked before the cap, so issues at index 5+ were dropped before their opportunity score was ever computed.

This discards the ranking the issue-quality producer goes to effort to build: `buildIssueQualityReport` scores each issue and returns them **sorted best-first**, but the consumer keyed the report by issue number for per-issue lookups and ignored the ordering. The quality signal therefore only adjusted the *score of whichever 5 issues came first*, never *which 5 were chosen*. A "ready", high-fit issue that the system itself ranked #1 was invisible if it happened to be the 6th available issue in DB order. Closes #393.

This is not cosmetic: these opportunities feed `buildContributorFit` and the contributor decision pack (`queue/processors.ts`, `services/decision-pack.ts`), and the global top-25 ranking was drawn from a biased per-repo sample — steering miners toward lower-fit issues while hiding the highest-value work.

## Scope
- `src/signals/engine.ts` — score every eligible (`rankable`) issue per repo, then keep that repo's top 5 by computed opportunity score (tie-break on issue number, mirroring the producer's order), instead of capping the raw list to 5 before scoring. The existing global sort + top-25 cap are unchanged.
- `test/unit/issue-quality.test.ts` — fail-on-revert: a "ready" issue sitting 6th in DB order (behind five "hold" issues) must now surface and rank first.

## Validation
- `npx tsc --noEmit` — clean.
- `npx vitest run` (full suite) — 1066 passed, 1 skipped; no other assertions affected.
- Branch coverage 97.15% (above the 97% gate); engine.ts 98.02% branch.

## Safety
- Behavior is unchanged for any repo with <= 5 eligible issues (scoring all then taking top 5 is identical to scoring the same set). The only change is which issues survive the cap when a repo has more than 5 — now the best-scoring ones, as intended.
- No public schema / API surface change; the opportunity object shape, scoring formula, and global ranking are untouched. The no-quality-report path also benefits (it now caps by lane/label fit rather than DB order).

## Notes
The fix selects by the consumer's own opportunity score (lane fit + label-history overlap + quality adjustment − queue/bounty penalties), which is the value the pipeline ultimately ranks on, so per-repo selection and global ranking are now consistent.

### DIFF
--- a/src/signals/engine.ts
+++ b/src/signals/engine.ts
@@ -1246,7 +1246,11 @@ export function buildContributorOpportunities(
     const rankable = qualityByIssue
       ? availableIssues.filter((issue) => qualityByIssue.get(issue.number)?.status !== "do_not_use")
       : availableIssues;
-    for (const issue of rankable.slice(0, 5)) {
+    // Score every eligible issue, then keep this repo's best 5 by score -- the cap must select the
+    // strongest-fit issues (mirroring the issue-quality report's score-descending order), not the
+    // arbitrary first 5 in DB order.
+    const repoOpportunities: ContributorOpportunity[] = [];
+    for (const issue of rankable) {
       const quality = qualityByIssue?.get(issue.number);
       const bounty = bountyByIssue.get(bountyIssueKey(repo.fullName, issue.number)) ?? null;
       const bountyLifecycle = bounty ? classifyBountyLifecycle(bounty, issue) : null;
@@ -1277,7 +1281,7 @@ export function buildContributorOpportunities(
       );
       const baseFit = score >= 70 ? "good" : score >= 40 ? "caution" : "hold";
       const downgradeToCaution = (bountyPenalty > 0 || quality?.status === "needs_proof") && baseFit === "good";
-      opportunities.push({
+      repoOpportunities.push({
         repoFullName: repo.fullName,
         issueNumber: issue.number,
         title: issue.title,
@@ -1302,6 +1306,8 @@ export function buildContributorOpportunities(
         ],
       });
     }
+    repoOpportunities.sort((left, right) => right.score - left.score || (left.issueNumber ?? 0) - (right.issueNumber ?? 0));
+    opportunities.push(...repoOpportunities.slice(0, 5));
   }
 
   /* v8 ignore next -- Repo-name tie ordering is deterministic presentation fallback after scored opportunity ranking. */
--- a/test/unit/issue-quality.test.ts
+++ b/test/unit/issue-quality.test.ts
@@ -238,6 +238,28 @@ describe("buildContributorOpportunities x issue quality", () => {
     expect(opportunities[0]?.reasons).toEqual(expect.arrayContaining(["Issue quality report rates this issue as ready."]));
   });
 
+  it("surfaces the highest-quality issue even when it sits beyond the first 5 in DB order", () => {
+    const repo = issueDiscoveryRepo("owner/deep-ready");
+    // Six available issues in DB order; the only "ready" one is last (index 5), the rest are "hold".
+    const issues = [1, 2, 3, 4, 5, 6].map((n) => issue(repo.fullName, n, `Issue ${n}`, { body: "x".repeat(220), labels: ["bug"] }));
+    const quality: IssueQualityReport = {
+      repoFullName: repo.fullName,
+      generatedAt: now(),
+      lane: { repoFullName: repo.fullName, lane: "issue_discovery", issueDiscoveryShare: 1, directPrShare: 0, summary: "", contributorGuidance: "", maintainerGuidance: "" },
+      issues: [
+        ...[1, 2, 3, 4, 5].map((n) => ({ number: n, title: `Issue ${n}`, status: "hold" as const, score: 30, reasons: [], warnings: [] })),
+        { number: 6, title: "Issue 6", status: "ready" as const, score: 92, reasons: [], warnings: [] },
+      ],
+      summary: "",
+    };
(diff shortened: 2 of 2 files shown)


JSONbored (approved):
@galuis116 this is good to merge.

A few notes:
- The ranking logic fits the signal layer and does not add a separate review path.
- The unit test gives enough coverage for the new evidence ordering.
- No code changes requested from me.

Validation expected:
- Current green CI is enough.

PR #399, opened 2026-06-04
fix(api): count cached open PRs across all in-scope repos on the maintainer dashboard

### LINKED ISSUE #398: [Bug]: Maintainer-dashboard "Open PRs cached" metric only counts the first 12 repositories, undercounting the global total (and its sparkline mixes PR-count over repo-count)
## Summary

The `/v1/app/maintainer-dashboard` handler builds a row of headline metrics. Three of the four are global counts across the full (scoped) data set, but **"Open PRs cached" is computed from a list that was deliberately capped at the first 12 repositories**, so it under-reports the real number whenever there are more than 12 repos in scope:

```ts
// src/api/routes.ts (maintainer-dashboard handler)
const openPullRequests = (
  await Promise.all(repositories.slice(0, 12).map((repo) =>          // <- capped at 12 repos (for the reviewability preview)
    listOpenPullRequests(c.env, repo.fullName).then((rows) => rows.map((pull) => ({ repoFullName: repo.fullName, pull }))),
  ))
).flat();
return c.json({
  // ...
  metrics: [
    { label: "Installations",     value: installations.length, spark: sparklineFromCounts(installations.length, Math.max(installations.length, 1)) },
    { label: "Open PRs cached",   value: openPullRequests.length, spark: sparklineFromCounts(openPullRequests.length, Math.max(repositories.length, 1)) }, // <- BUG
    { label: "Install issues",    value: health.filter((r) => r.status !== "healthy").length, /* ... */ },
    { label: "Rate-limit events", value: rateLimits.length, /* ... */ },
  ],
  reviewability: openPullRequests.slice(0, 20).map(/* ... */),       // the 12-repo cap exists to bound THIS preview list
  // ...
});
```

`openPullRequests` is intentionally limited to `repositories.slice(0, 12)` because it feeds the `reviewability` preview (sliced again to 20). Reusing that same perf-capped list for the **"Open PRs cached" metric value** means the metric reflects only the first 12 repositories, not all of them.

## Why this is wrong

The metric is presented as a global headline count alongside its siblings, all of which span the full scoped set:

- `Installations` → `installations.length` (all installations)
- `Install issues` → `health.filter(...)` (all health records)
- `Rate-limit events` → `rateLimits.length` (all rate-limit observations)
- `Open PRs cached` → `openPullRequests.length` — **only the first 12 repositories**

So for an operator (no scope → all repositories) or any maintainer/owner whose scope includes more than 12 repos, "Open PRs cached" silently undercounts. The repo ordering from `listRepositories` is arbitrary, so which 12 repos contribute is effectively non-deterministic from the operator's point of view; adding/removing repos can change the number with no actual change in open PRs.

Secondarily, the sparkline for this metric is dimensionally inconsistent:

```ts
spark: sparklineFromCounts(openPullRequests.length, Math.max(repositories.length, 1))
```

The numerator is a **PR count** (from 12 repos) and the denominator is a **repository count** (all repos) — two different units. The intended `value / max` ratio is meaningless here (the numerator can exceed or fall arbitrarily short of the denominator depending on PR density), so the rendered bar does not represent any real proportion.

## Reachability

- `GET /v1/app/maintainer-dashboard`, gated to `maintainer`/`owner`/`operator` roles.
- Operators see all repositories (`scope === null`), so any deployment with >12 registered/installed repos shows an undercounted value. Scoped maintainers with >12 repos in scope are likewise affected.

A deployment has 30 registered repos; the first 12 (arbitrary order) hold 8 open cached PRs, the other 18 hold 25 more.

- **Current:** `Open PRs cached = 8` (only the first 12 repos), with a sparkline of `8 / 30`.
- **Correct:** `Open PRs cached = 33` (all repos), with a sparkline proportional to a meaningful denominator.

An operator reading the dashboard concludes there are 8 open PRs cached when there are 33 — a ~4x undercount that worsens as repo count grows.

## Steps to reproduce

1. As an operator (or a maintainer scoped to >12 repos), call `GET /v1/app/maintainer-dashboard` with more than 12 repositories that have cached open PRs.
2. Compare the `Open PRs cached` metric value to the actual total number of cached open PRs across all in-scope repos.
3. Observe the metric equals only the open-PR count of the first 12 repositories.

## Expected behavior

"Open PRs cached" reports the total cached open PRs across all in-scope repositories (consistent with the other global metrics), and its sparkline uses a same-unit denominator.

## Actual behavior

The metric counts open PRs for only the first 12 repositories (the slice that exists to bound the reviewability preview), and the sparkline divides a 12-repo PR count by the all-repo count.

## Suggested fix

Decouple the metric's count from the perf-capped preview fetch. The cached open-PR count per repo is already available cheaply via `listRepoSyncStates` (`RepoSyncStateRecord.openPullRequestsCount`, the same source `buildRepoDecision` uses), so the metric can sum across all in-scope repos in a single query without fetching every repo's PRs:

- Load `listRepoSyncStates(c.env)` (already a single query), scope it to `repositories` (by `repoFullName`), and compute `totalOpenPrs = sum(openPullRequestsCount)`.
- Use `totalOpenPrs` as the `Open PRs cached` value, and a same-unit denominator for the sparkline (e.g. a sensible max such as the number of in-scope repos times an expected-per-repo bound, or simply `Math.max(totalOpenPrs, 1)` to render a filled bar like the `Installations` metric does).
- Keep the existing `repositories.slice(0, 12)` fetch solely for the `reviewability` preview list, where the cap is intended.

Add fail-on-revert coverage: a dashboard built from >12 repos whose later repos hold open PRs must report the full total, not the first-12 subtotal.

philluiz2323:
## Summary
The `/v1/app/maintainer-dashboard` "Open PRs cached" headline metric was computed from `openPullRequests`, a list deliberately limited to `repositories.slice(0, 12)` (that cap exists to bound the `reviewability` preview). Its three sibling metrics (Installations, Install issues, Rate-limit events) are global counts across the full scoped set, so "Open PRs cached" silently undercounted whenever more than 12 in-scope repos held open PRs. Because repo ordering from `listRepositories` is arbitrary, which 12 repos contributed was effectively non-deterministic. The metric's sparkline was also dimensionally inconsistent -- a PR-count numerator (`openPullRequests.length`) over a repo-count denominator (`repositories.length`). Closes #398.

Reachable for operators (no scope -> all repos) and any maintainer/owner scoped to >12 repos. Example: 30 repos with 8 open PRs in the first 12 and 25 in the rest reported 8 instead of 33.

## Scope
- `src/api/routes.ts` -- the cached open-PR count is now summed across all in-scope repos from sync state (`listRepoSyncStates`, added to the existing `Promise.all`; `RepoSyncStateRecord.openPullRequestsCount`, the same source `buildRepoDecision` uses), a single query with no extra per-repo fetches. The sparkline now uses a same-unit ratio (`reposWithOpenPullRequests / repositories.length`). The `repositories.slice(0, 12)` PR fetch is unchanged and still feeds only the `reviewability` preview list, where the cap is intended.
- `test/integration/api.test.ts` -- fail-on-revert: two repos carry cached open-PR counts in sync state (5 + 3) but have no open PR records, so the old per-repo-fetch metric reported 0 while the global count reports 8.

## Validation
- `npx tsc --noEmit` -- clean.
- `npx vitest run test/integration/api.test.ts` -- 26/26 pass (including the new test).
- Full suite -- 1029 pass, 1 skipped (excluding the two unparseable upstream test files and one unrelated `mcp-cli` timeout that passes in isolation).
- Branch coverage 97.15% (above the 97% gate).

## Safety
- No response-shape change: the `metrics` array keeps the same labels/fields; only the (now correct, larger-or-equal) "Open PRs cached" value and its sparkline change.
- Scope handling is unchanged -- the new sum reuses the existing scoped `repositories` set, so scoped maintainers still see only their repos.
- Uses already-cached sync-state counts; no additional GitHub or per-repo PR queries.

## Notes
The sibling metrics remain as-is; only "Open PRs cached" diverged from the global-count convention.

### DIFF
--- a/src/api/routes.ts
+++ b/src/api/routes.ts
@@ -824,11 +824,12 @@ export function createApp() {
     const summary = await getRoleSummaryForIdentity(c.env, identity);
     if (!summary.roles.some((role) => ["maintainer", "owner", "operator"].includes(role))) return c.json({ error: "insufficient_role" }, 403);
 
-    const [allRepositories, allInstallations, allHealth, allRateLimits] = await Promise.all([
+    const [allRepositories, allInstallations, allHealth, allRateLimits, allSyncStates] = await Promise.all([
       listRepositories(c.env),
       listInstallations(c.env),
       listInstallationHealth(c.env),
       listLatestGitHubRateLimitObservations(c.env, 20),
+      listRepoSyncStates(c.env),
     ]);
     const scope = identity.kind === "session" && !summary.roles.includes("operator") ? await loadControlPanelAccessScope(c.env, identity.actor) : null;
     const scopedRepoNames = new Set(scope?.repositoryFullNames.map((repo) => repo.toLowerCase()) ?? []);
@@ -842,6 +843,13 @@ export function createApp() {
       ? allHealth.filter((record) => scopedInstallationIds.has(record.installationId) || scopedAccountLogins.has(record.accountLogin.toLowerCase()))
       : allHealth;
     const rateLimits = scope ? allRateLimits.filter((record) => record.repoFullName !== undefined && record.repoFullName !== null && scopedRepoNames.has(record.repoFullName.toLowerCase())) : allRateLimits;
+    // Cached open-PR count is summed across ALL in-scope repos from sync state (a single query) so the
+    // headline metric is a true global count like its siblings. The per-repo PR fetch below is capped at
+    // 12 only to bound the `reviewability` preview list, not the metric.
+    const scopedRepoNameSet = new Set(repositories.map((repo) => repo.fullName.toLowerCase()));
+    const scopedSyncStates = allSyncStates.filter((state) => scopedRepoNameSet.has(state.repoFullName.toLowerCase()));
+    const totalOpenPullRequestsCached = scopedSyncStates.reduce((sum, state) => sum + Math.max(0, state.openPullRequestsCount), 0);
+    const reposWithOpenPullRequests = scopedSyncStates.filter((state) => state.openPullRequestsCount > 0).length;
     const openPullRequests = (
       await Promise.all(repositories.slice(0, 12).map((repo) => listOpenPullRequests(c.env, repo.fullName).then((rows) => rows.map((pull) => ({ repoFullName: repo.fullName, pull })))))
     ).flat();
@@ -851,7 +859,7 @@ export function createApp() {
       health: health.map(enrichInstallationHealth),
       metrics: [
         { label: "Installations", value: installations.length, spark: sparklineFromCounts(installations.length, Math.max(installations.length, 1)) },
-        { label: "Open PRs cached", value: openPullRequests.length, spark: sparklineFromCounts(openPullRequests.length, Math.max(repositories.length, 1)) },
+        { label: "Open PRs cached", value: totalOpenPullRequestsCached, spark: sparklineFromCounts(reposWithOpenPullRequests, Math.max(repositories.length, 1)) },
         { label: "Install issues", value: health.filter((record) => record.status !== "healthy").length, spark: sparklineFromCounts(health.filter((record) => record.status === "healthy").length, Math.max(health.length, 1)) },
         { label: "Rate-limit events", value: rateLimits.length, spark: sparklineFromCounts(rateLimits.filter((record) => (record.remaining ?? 0) > 0).length, Math.max(rateLimits.length, 1)) },
       ],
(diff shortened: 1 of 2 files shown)


JSONbored (approved):
@philluiz2323 this looks ready from my side.

A few notes:
- The stale readiness count is wired into the maintainer dashboard path instead of living as an isolated helper.
- The integration test covers the behavior that matters for the API consumer.
- No code changes requested from me; merge after required review clears.

Validation expected:
- Current green CI is sufficient.

PR #438, opened 2026-06-06
fix(backfill): stop PR detail re-queue loop when a batch makes no progress

joaovictor712:
## Summary
`backfillOpenPullRequestDetails` (`src/github/backfill.ts`) processes open PRs' detail data (files, reviews, checks) in fixed-size batches, always starting from the front of the *incomplete* list (`cursor = 0` by design), and re-queues itself while the incomplete list is larger than one batch:

```ts
const nextCursor = batch.length < incompleteOpenPullRequests.length ? 0 : undefined;   // no progress check
```

`nextCursor` was set to `0` (continue) whenever `incompleteOpenPullRequests.length > batchSize`, **without checking whether the batch completed any PR**. A PR stays incomplete ("partial") whenever its detail fetch warns -- e.g. file sync fails on both REST and GraphQL (`backfill.ts`: `File sync failed for #N`), which happens for real PRs (changed-files API limits, deleted source branch). If at least `batchSize` such PRs sit at the front of the number-sorted incomplete list, the batch is always those same PRs, they never complete to drop out, the incomplete count never shrinks, and `nextCursor` stays `0` forever: the job re-queues itself every 20s indefinitely, the `pull_request_files`/`pull_request_reviews`/`check_summaries` segments stay `"running"`, and **every PR past the stuck front is never synced**. `PR_DETAIL_BATCH_SIZE` is `{ light: 12, full: 40, resume: 40 }`; once triggered the stall never self-heals.

Downstream consumers that rely on PR detail data (changed-files / review / check evidence -> reviewability, preflight `missingTests`, scoring) then silently operate on missing data for most of the repo's PRs, while the queue does unbounded busy-work.

## Scope
- `src/github/backfill.ts` -- add a no-progress guard: compute `incompleteAfter` from the refreshed detail states and `madeProgress = incompleteAfter < incompleteOpenPullRequests.length`; only keep cursoring (`nextCursor = 0`) when the batch actually shrank the incomplete list. When a full batch makes no progress, terminate as `"partial"` instead of re-queuing the identical failing batch. A later backfill run can retry once transient failures clear.
- `test/unit/backfill.test.ts` -- fail-on-revert: a repo whose front-12 PRs all fail file sync (no batch completes) now returns a terminal result (`status: "partial"`, no `nextCursor`, no re-queue) instead of `status: "running"` + `nextCursor: 0` + a `backfill-pr-details` re-queue.

## Validation
- `npx tsc --noEmit` -- clean.
- `npx vitest run test/unit/backfill.test.ts` -- 56/56 (the existing normal-case light-mode test, where each round completes some PRs, is unchanged -- progress is still made so cursoring continues as before).
- Full suite -- 1159 passed, 1 skipped (excluding the two unparseable upstream test files and one unrelated `mcp-cli` timeout that passes in isolation).
- Branch coverage 97.02% (above the 97% gate); backfill.ts 100% line.

## Safety
- Behavior is unchanged for the normal case (each round completes at least one PR, so `madeProgress` is true and cursoring continues exactly as before, terminating when the incomplete list reaches one batch).
- Only the pathological case changes: a full batch of persistently-"partial" PRs now stops as `"partial"` rather than looping forever. No persisted-data clobbering; the same segments and counts are written, just with a terminal status.

## Notes
A complementary hardening (left out to keep this focused) would be a per-PR retry cap so a chronically-failing PR is eventually dropped from the incomplete set.

### DIFF
--- a/src/github/backfill.ts
+++ b/src/github/backfill.ts
@@ -523,8 +523,16 @@ export async function backfillOpenPullRequestDetails(
     });
   });
   const refreshedDetailStates = await listPullRequestDetailSyncStates(env, repo.fullName);
+  const refreshedStateByPull = new Map(refreshedDetailStates.map((state) => [state.pullNumber, state.status]));
   const completedCount = refreshedDetailStates.filter((state) => openPullNumbers.has(state.pullNumber) && state.status === "complete").length;
-  const nextCursor = batch.length < incompleteOpenPullRequests.length ? 0 : undefined;
+  // Only keep cursoring over the oldest incomplete PRs if this batch actually shrank the incomplete
+  // list. The cursor is fixed at 0, so a full batch of persistently-"partial" PRs (e.g. repeated
+  // detail-fetch failures) never completes and never drops out; without this guard `nextCursor` would
+  // stay 0 and re-queue the identical failing batch forever, starving every later PR. Stop as "partial"
+  // when no progress is made; a later backfill run can retry once the transient failures clear.
+  const incompleteAfter = openPullRequests.filter((pr) => refreshedStateByPull.get(pr.number) !== "complete").length;
+  const madeProgress = incompleteAfter < incompleteOpenPullRequests.length;
+  const nextCursor = batch.length < incompleteOpenPullRequests.length && madeProgress ? 0 : undefined;
   const status: RepoSyncSegmentRecord["status"] = nextCursor !== undefined ? "running" : completedCount >= openPullRequests.length ? "complete" : "partial";
   await Promise.all(
     (["pull_request_files", "pull_request_reviews", "check_summaries"] as const).map((segment) =>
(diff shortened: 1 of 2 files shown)


ghost:
> [!NOTE]
> ## Gittensory Gate skipped
>
> | --- | --- | --- | --- |
> | Gate result | ⚠️ Skipped | JSONbored/gittensory#438 is no longer open. | No action. |

---
Checked by [Gittensory](https://github.com/JSONbored/gittensory), a quiet PR intelligence layer for OSS maintainers.

JSONbored (approved):
@galuis116

A few notes:
- This is the right shape for the stuck PR-detail backfill loop. The important part is that the fixed cursor only continues when the incomplete set actually shrinks.
- The regression test covers the failure mode that matters: a full front batch staying partial should stop as partial instead of re-queueing forever.
- I am not asking for code changes. The missing linked issue is not blocking here because the body explains the operational bug and scope clearly.

Validation expected:
- Keep the current green `validate` state through merge.

PR #595, opened 2026-06-11
feat(ui): self-hosted Umami analytics

JSONbored:
Adds the privacy-friendly self-hosted Umami tracking tag (tasty.aethereal.dev) as a deferred head script in __root.tsx. Cookieless, no PII.

PR #599, opened 2026-06-11
fix(ui): strip cookies from the analytics proxy request path

### LINKED ISSUE #597: [Bug]: Umami analytics reverse-proxy forwards first-party request cookies to the upstream, breaking its "cookieless / strips cookies" guarantee
## Summary

`handleAnalyticsProxy` (the first-party reverse proxy for the self-hosted Umami
tracker, added in #595) strips the **response** `set-cookie` header but does
**not** strip the **request** `cookie` header. So the browser's first-party
cookies for the proxy origin (`gittensory.aethereal.dev`) — including any
session / auth / OAuth-state cookie whose path matches `/stats/*` — are forwarded
verbatim to the Umami upstream (`tasty.aethereal.dev`) on every
`GET /stats/script.js` and `POST /stats/api/send`.

This directly contradicts the design as documented in the same change.

## The stated invariant

The #595 commit message:

> The proxy is a strict path allowlist … **strips cookies**/hop-by-hop headers,
> forwards the real client IP for geolocation, and fails closed (502)…

and the file header comment ("cookieless analytics"). The response side honors
it; the request side does not.

## Evidence

```ts
// apps/gittensory-ui/src/lib/analytics-proxy.ts:26
const STRIP_REQUEST_HEADERS = new Set([
  "host", "connection", "keep-alive", "transfer-encoding", "upgrade",
  "cf-connecting-ip", "cf-ipcountry", "cf-ray", "cf-visitor",
  "x-forwarded-host", "x-forwarded-proto",
]);                                   // <-- no "cookie"

// :43  (response side, correctly stripped)
const STRIP_RESPONSE_HEADERS = new Set([
  "connection", "keep-alive", "transfer-encoding", "content-encoding",
  "content-length",
  "set-cookie", // cookieless analytics: never relay cookies to the client
]);

// :71  forwards every request header except the strip-list -> cookie is forwarded
const headers = new Headers();
request.headers.forEach((value, key) => {
  if (!STRIP_REQUEST_HEADERS.has(key.toLowerCase())) headers.set(key, value);
});
```

## Reachability

Live in production. `apps/gittensory-ui/src/server.ts` runs the proxy ahead of
SSR for every request:

```ts
const analytics = await handleAnalyticsProxy(request);
if (analytics) return analytics;
```

When a logged-in visitor loads any page, the browser fetches
`/stats/script.js` and the tracker `POST`s `/stats/api/send` — both same-origin
requests that the browser decorates with the origin's cookies. Those cookies are
then forwarded to `tasty.aethereal.dev`.

`gittensory.aethereal.dev` does set first-party cookies (e.g. the GitHub OAuth
state cookie via `buildGitHubOAuthStateCookie`, and the authenticated session),
so this is not hypothetical.

## Impact

- **Privacy guarantee broken.** Umami is deployed precisely for "cookieless, no
  PII" analytics; forwarding the visitor's cookies to the analytics host
  undermines that and lands those cookies in the analytics server's request
  logs.
- **Expanded credential exposure.** Session / auth cookies now traverse to, and
  may be logged by, a second host. A compromise or log leak of the analytics
  host would expose live `gittensory` session cookies — a surface that should
  not exist for an analytics beacon.

## Test status

Not locked in — there is no test for `handleAnalyticsProxy` (no `analytics-proxy`
test file). No test asserts the request header strip-list.

## Suggested fix

Strip `cookie` from forwarded requests, mirroring the response-side `set-cookie`
strip:

```ts
const STRIP_REQUEST_HEADERS = new Set([
  "host", "connection", "keep-alive", "transfer-encoding", "upgrade",
  "cookie", // cookieless analytics: never forward first-party cookies upstream
  "cf-connecting-ip", "cf-ipcountry", "cf-ray", "cf-visitor",
  "x-forwarded-host", "x-forwarded-proto",
]);
```

Add a unit test for `handleAnalyticsProxy` asserting that a request with a
`Cookie` header reaches the upstream `fetch` with **no** `cookie` header (and,
for parity, that `set-cookie` is absent from the returned response).

## Secondary note (same function, lower severity)

`x-forwarded-for` is **not** in the strip-list, and the proxy appends the real
`cf-connecting-ip` *after* any client-supplied `x-forwarded-for`
(`existing ? `${existing}, ${clientIp}`` at :78–79). A client can therefore
prepend a spoofed IP that Umami may treat as the visitor's address for
geolocation. Setting `x-forwarded-for` to the trusted `cf-connecting-ip` only
(overwrite, not append) closes this and is consistent with the cookie fix.

## Distinct from prior reports

Not a duplicate of #461 (that restricted the public **GitHub stats** API proxy
to known repos). This is the **UI Umami tracker** reverse-proxy and a
request-header forwarding/strip defect, in newly-added #595 code.

joaovictor712:
Closes #597.

`handleAnalyticsProxy` stripped the **response** `set-cookie` but did **not** strip the **request** `cookie` header, so the browser's first-party cookies for `gittensory.aethereal.dev` (session / GitHub OAuth state) were forwarded to the Umami upstream (`tasty.aethereal.dev`) on every `/stats/script.js` and `/stats/api/send` — breaking the documented "cookieless / strips cookies" guarantee (the #595 commit message explicitly claims it "strips cookies").

## Changes
- Add `cookie` to `STRIP_REQUEST_HEADERS` so first-party cookies are never forwarded upstream.
- Add `x-forwarded-for` to the strip-list and set it to the trusted `cf-connecting-ip` only (overwrite, not append) so a client-supplied `x-forwarded-for` can't spoof the visitor's geolocation.
- Use `null` (not `undefined`) for the no-body `fetch` so the file typechecks under the root tsconfig once a test imports it.
- Add `test/unit/analytics-proxy.test.ts` (first test for this proxy): cookies are not forwarded, `set-cookie` is not relayed back, a spoofed `x-forwarded-for` is ignored in favor of `cf-connecting-ip`, and the path allowlist / 405 behavior holds.

## Verification
- `tsc --noEmit` clean; `analytics-proxy.test.ts` 3/3 (the cookie and x-forwarded-for assertions fail on the old code). `apps/**` is excluded from coverage, so the gate is unaffected.

### DIFF
--- a/apps/gittensory-ui/src/lib/analytics-proxy.ts
+++ b/apps/gittensory-ui/src/lib/analytics-proxy.ts
@@ -29,12 +29,16 @@ const STRIP_REQUEST_HEADERS = new Set([
   "keep-alive",
   "transfer-encoding",
   "upgrade",
+  // Cookieless analytics: never forward the visitor's first-party cookies upstream.
+  "cookie",
   "cf-connecting-ip",
   "cf-ipcountry",
   "cf-ray",
   "cf-visitor",
   "x-forwarded-host",
   "x-forwarded-proto",
+  // Re-derived below from the trusted cf-connecting-ip; never trust a client-supplied value.
+  "x-forwarded-for",
 ]);
 
 // Response headers we never relay back to the browser. content-encoding/-length
@@ -72,12 +76,11 @@ export async function handleAnalyticsProxy(request: Request): Promise<Response |
   request.headers.forEach((value, key) => {
     if (!STRIP_REQUEST_HEADERS.has(key.toLowerCase())) headers.set(key, value);
   });
-  // Preserve the real client IP so Umami geolocates the visitor, not the Worker.
+  // Preserve the real client IP so Umami geolocates the visitor, not the Worker. Set it to the
+  // trusted cf-connecting-ip only -- the client-supplied x-forwarded-for is stripped above so a
+  // visitor cannot spoof their geolocation.
   const clientIp = request.headers.get("cf-connecting-ip");
-  if (clientIp) {
-    const existing = request.headers.get("x-forwarded-for");
-    headers.set("x-forwarded-for", existing ? `${existing}, ${clientIp}` : clientIp);
-  }
+  if (clientIp) headers.set("x-forwarded-for", clientIp);
 
   const hasBody = request.method !== "GET" && request.method !== "HEAD";
 
@@ -87,7 +90,7 @@ export async function handleAnalyticsProxy(request: Request): Promise<Response |
       method: request.method,
       headers,
       // Buffer the (tiny) collect payload so we don't need a streaming/duplex body.
-      body: hasBody ? await request.arrayBuffer() : undefined,
+      body: hasBody ? await request.arrayBuffer() : null,
     });
   } catch {
     // Analytics must never take the page down — fail quietly.
--- a/test/unit/analytics-proxy.test.ts
+++ b/test/unit/analytics-proxy.test.ts
@@ -1,0 +1,60 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { handleAnalyticsProxy } from "../../apps/gittensory-ui/src/lib/analytics-proxy";
+
+const UPSTREAM = "https://tasty.aethereal.dev";
+
+function captureUpstream() {
+  const calls: Array<{ url: string; init: RequestInit; headers: Headers }> = [];
+  vi.stubGlobal("fetch", async (url: RequestInfo | URL, init: RequestInit = {}) => {
+    calls.push({ url: url.toString(), init, headers: new Headers(init.headers) });
+    return new Response("ok", { status: 200, headers: { "set-cookie": "umami=1", "content-type": "application/javascript" } });
+  });
+  return calls;
+}
+
+describe("handleAnalyticsProxy", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("does not forward the visitor's cookies to the analytics upstream", async () => {
+    const calls = captureUpstream();
+    const response = await handleAnalyticsProxy(
+      new Request("https://gittensory.aethereal.dev/stats/script.js", {
+        method: "GET",
+        headers: { cookie: "gittensory_session=secret; gh_oauth_state=abc", "cf-connecting-ip": "203.0.113.7" },
+      }),
+    );
+
+    expect(response?.status).toBe(200);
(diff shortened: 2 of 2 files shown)


ghost:
> [!NOTE]
> ## Gittensory Gate skipped
>
> | --- | --- | --- | --- |
> | Gate result | ⚠️ Skipped | JSONbored/gittensory#599 is no longer open. | No action. |

---
Checked by [Gittensory](https://github.com/JSONbored/gittensory), a quiet PR intelligence layer for OSS maintainers.

ghost:
**reviewbot** · advisory review

Reviewed **2** changed file(s) — two independent AI reviewers.

The PR tightens privacy by stripping cookies and client‑supplied x-forwarded-for headers from analytics proxy requests, and simplifies the forwarding of the trusted client IP. It adds comprehensive unit tests to verify the new behavior. The changes are small, well‑documented, and appear safe.

_Suggestions_
- Consider adding a brief comment explaining why `null` is used for the request body instead of `undefined`.
- Verify that the TypeScript type for `RequestInit.body` accepts `null` in the project’s tsconfig (or cast as `null as any` if needed).
- Optionally add a test case ensuring other stripped request headers (e.g., "keep-alive") are also omitted.

_Worth double-checking_
- Make sure the runtime environment (Cloudflare Workers) correctly handles `body: null` for GET/HEAD requests.
- Confirm that any downstream services consuming the proxied request do not rely on the original `x-forwarded-for` header.

The PR strips the Cookie header and client‑supplied X-Forwarded-For from analytics proxy requests, forwarding only the trusted CF‑Connecting-IP to prevent cookie leakage and IP spoofing. It adds corresponding unit tests that verify cookies are not forwarded, X-Forwarded-For is replaced with the trusted IP, and non‑allowlisted paths/methods are handled correctly. The changes are small, focused, and well‑tested.

_Suggestions_
- Consider adding a brief comment explaining why `body: null` is used instead of `undefined` (though both are functionally equivalent).
- If the project has a convention for header‑set constants (e.g., all‑uppercase names), ensure `STRIP_REQUEST_HEADERS` follows it; otherwise it's fine as is.

_Worth double-checking_
- Removing X-Forwarded-For may discard legitimate proxy chained IPs if the deployment ever relies on more than just CF‑Connecting-IP; verify that the analytics setup does not need the full chain.
- Ensure the response‑header stripping logic (not shown in the diff) still removes Set‑Cookie from upstream responses so the test expectation holds.

_Before = production · After = this PR's preview deploy._

| Route | Before | After |
| --- | --- | --- |
| `/` | before | after |

JSONbored (approved):

PR #652, opened 2026-06-13
feat(ai-review): dual-AI maintainer review + BYOK (Phase C)

JSONbored:
## Phase C — AI agent review (Workers AI default + BYOK)

Adds a gittensor-aware AI maintainer review to the public PR path. It is **opt-in and dormant by default** — gated by per-repo `aiReviewMode` (default `off`) AND the existing `AI_SUMMARIES_ENABLED` / `AI_PUBLIC_COMMENTS_ENABLED` flags (both default `false`). Nothing changes in production until a maintainer enables it.

### What it does
Two layers, both fail-safe (no AI / error / over-budget / unsafe output → no public text and no gate finding — gittensory never blocks because a model spoke):

- **Advisory notes** — a concise maintainer-style write-up (assessment / suggestions / risks) rendered as an advisory `Gittensory AI review` section in the PR panel. Free Cloudflare Workers AI by default; the maintainer's frontier model when BYOK is configured.
- **Consensus defect (opt-in blocker)** — a defect becomes a gate blocker only when `aiReview: block` is set **and** the two free Workers-AI models (`gpt-oss-120b` + `nemotron`, adapted from reviewbot's proven pair) **independently** agree on a high-confidence (≥0.9) critical defect. Still confirmed-contributor gated — only confirmed Gittensor contributors can ever be hard-blocked. BYOK never drives this path, so it never changes *who* can be blocked.

### Config-as-code
Everything is controllable from `.gittensory.yml` (precedence `yml > DB > defaults`):
```yaml
gate:
  aiReview:
    mode: advisory   # off | advisory | block
    byok: false
```
Also settable via the dashboard settings route and the generic `settings:` override.

### BYOK (encrypted at rest)
- New `encryptSecret`/`decryptSecret` (AES-256-GCM, PBKDF2-derived key, random per-write IV) — gittensory's first reversible-encryption primitive — keyed by a new `TOKEN_ENCRYPTION_SECRET` worker secret.
- Isolated `repository_ai_keys` table (migration 0027) so ciphertext is never serialized by the settings surface. Stores ciphertext + iv + key_version + a display-only `last4`; the plaintext key is never persisted.
- Write-only internal routes `POST`/`GET`/`DELETE /v1/internal/repos/:owner/:repo/ai-key` — `GET` returns only `{ configured, provider, last4, model }`, never the key. The key is never logged and never reaches a public surface.

### Privacy
[...]

PR #650, opened 2026-06-13
feat(mcp): gittensory_remediation_plan

kiannidev:
## Summary

- Adds `buildRemediationPlan` to consolidate `branchQualityBlockers`, `accountStateBlockers`, and `scoreBlockers` from local branch analysis into an ordered, deduplicated public-safe checklist.
- Each remediation item includes an impact level and rerun condition; account-state blockers are prioritized ahead of branch-quality and score blockers.
- Registers the MCP tool `gittensory_remediation_plan` in the HTTP MCP server and local MCP package, plus `POST /v1/local/remediation-plan` for the bin client.

Closes #548

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build:mcp`
- [x] `npm run test:coverage -- test/unit/remediation-plan.test.ts`
- [ ] Full CI (`npm run validate`)

Made with [Cursor](https://cursor.com)

PR #655, opened 2026-06-13
fix(gate): never leave the Gittensory Gate check stuck in_progress

JSONbored:
### Root cause
`maybePublishPrPublicSurface` posts the **"Gittensory Gate is evaluating"** check (`status: in_progress`) early, then does a block of D1 + GitHub + Gittensor work before the completing PATCH — with **no request timeouts** and **no failure finalization**. If anything in that window hangs (a bare `fetch` to a slow/half-open upstream) or throws, the completing PATCH never runs and the check is orphaned `in_progress` forever; the caller's `.catch` only logs it.

Confirmed from prod (`audit_events` / `webhook_events`): miner detection `cache_hit` 13:35:08 → pending posted 13:35:11 → then **nothing** (no completion, **no error row** → a hang, not a caught throw), while prod D1 was intermittently overloaded. Since miner detection was cached, the hang was in the completion's own un-timed GitHub calls.

### Fix (3 commits)
1. **Bound every external call** in the gate window so a hang becomes a catchable error: `AbortSignal.timeout` on the GitHub App fetches + the Octokit instance (`src/github/app.ts`, 12s) and the Gittensor API client (`src/gittensor/api.ts`, 10s).
2. **Always finalize the check.** Wrap the pending-post → completion window in `try/catch`; on any failure, finalize the **same** check run to a **neutral, non-blocking** `"Gittensory Gate — could not finish evaluating — will re-run"` state (`createOrUpdateErroredGateCheckRun`) + a `github_app.gate_finalized_on_error` audit. Guarded so it never clobbers a real `success`/`failure` verdict.
3. **Dead-letter the queue.** The `gittensory-jobs` consumer had no `dead_letter_queue`, so a webhook job that exhausts its retries was silently dropped. Route exhausted jobs to a new `gittensory-jobs-dlq` landing queue (already provisioned on the account; no consumer, matching the house pattern) and set `max_retries` explicitly. Validated with `wrangler deploy --dry-run`.

Net effect: a transient blip leaves a completed (harmless) check that re-runs on the next push, and a persistently-failing job lands in the DLQ for inspection instead of vanishing.

### Test
A queue test drives a confirmed `in_progress` post, fails the completion PATCH (500), and asserts the **same** check id is then PATCHed to `status: completed, conclusion: neutral, title: "Gittensory Gate — could not finish evaluating"`, with the audit row recorded. Full suite green; coverage holds above the 97% gate; Workers-runtime tests pass.

### Immediate remediation for PR #650
Re-trigger the gate (push any commit, or tick **"Re-run Gittensory review"** in the panel) — the orphaned run is owned by `gittensory[bot]`, so only a fresh run from the app supersedes it.

### Related
Separate from #652 (AI review + BYOK) and #656 (timestamp-default fix); all came out of the same gate-hang investigation.

### DIFF
--- a/src/github/app.ts
+++ b/src/github/app.ts
@@ -26,9 +26,20 @@ export const GITTENSORY_GATE_CHECK_NAME = "Gittensory Gate";
 type GitHubCheckConclusion = Advisory["conclusion"] | GateCheckConclusion | "skipped";
 type GitHubCheckStatus = "queued" | "in_progress" | "completed";
 
+/** Hard cap on a single GitHub API request. Without it a slow/half-open GitHub connection can hang the
+ *  Worker — e.g. the Gate's own completing PATCH stalling after the pending check was posted, which leaves
+ *  the check in_progress forever. A bounded timeout turns a hang into a catchable error the caller can
+ *  finalize. Applied to every raw fetch here and to the Octokit instances (via a timeout-injecting fetch). */
+const GITHUB_FETCH_TIMEOUT_MS = 12_000;
+
+function timeoutFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+  if (init?.signal) return fetch(input, init);
+  return fetch(input, { ...(init ?? {}), signal: AbortSignal.timeout(GITHUB_FETCH_TIMEOUT_MS) });
+}
+
 export async function createInstallationToken(env: Env, installationId: number): Promise<string> {
   const jwt = await createAppJwt(env);
-  const response = await fetch(`https://api.github.com/app/installations/${installationId}/access_tokens`, {
+  const response = await timeoutFetch(`https://api.github.com/app/installations/${installationId}/access_tokens`, {
     method: "POST",
     headers: githubHeaders(`Bearer ${jwt}`),
   });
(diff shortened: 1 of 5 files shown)


ghost:

Reviewed **5** changed file(s) — two independent AI reviewers.

The PR adds robust timeout handling for GitHub and Gittensor API calls and introduces a safe fallback that finalizes a pending Gate check to a neutral state on errors, preventing in_progress hangs. The changes are well‑scoped, include comprehensive tests, and respect the public/private output boundaries.

_Suggestions_
- Document the new GITHUB_FETCH_TIMEOUT_MS and GITTENSOR_FETCH_TIMEOUT_MS constants in the developer guide.
- Add a fallback for environments where AbortSignal.timeout may not be available, e.g., a manual AbortController with setTimeout.
- Consider exporting the timeout constants from a central module for easier configuration.
- Add a unit test ensuring that non‑timeout fetches (quick calls) still succeed without being aborted.

_Worth double-checking_
- AbortSignal.timeout is relatively new; verify runtime support on all target Workers environments.
- The new timeout may cause legitimate long‑running API calls to be aborted earlier than expected; ensure this does not affect other features.

The PR adds timeouts to GitHub and Gittensor API calls, introduces a function to finalize stuck Gate checks to a neutral state on error, and wraps gate evaluation in try/catch to prevent in_progress hangs. It also adds a dead letter queue for observable failures and includes a test verifying the fix. The changes are focused, well-tested, and address the core issue without violating public/private boundaries.

_Suggestions_
- Consider adding a comment explaining why 12s and 10s were chosen for timeouts (e.g., based on observed latency percentiles).
- The audit event detail uses `errorMessage(error)`; ensure this function doesn't leak sensitive info (though it's internal).
- In the try/catch block, `gateFinalized` is set only when `gateCheckResult?.kind === 'published'`; verify this aligns with all possible successful gate check outcomes.

_Worth double-checking_
- Timeout values might need tuning if legitimate API calls occasionally exceed these limits (monitor for false-positive timeouts).
- Dead letter queue requires operational awareness to inspect failed jobs; ensure team is aware of this new visibility mechanism.
- The try/catch now catches *all* errors during gate evaluation; confirm no unintended errors are silently converted to neutral checks (though audit logs and re-throw help).

PR #820, opened 2026-06-17
chore(ci): guard migration numbering against silent duplicates

JSONbored:
Addresses a real, recurring latent risk: two PRs that each grab the same next migration number make **different files** (`0038_foo.sql` + `0038_bar.sql`), so git reports **no conflict** and both merge — then `wrangler d1 migrations apply` runs both and, if they touch the same column, **errors mid-deploy**.

**This already happened twice** — `0015` and `0017` are each duplicated in the repo right now (two PRs silently grabbed each number). The guard found them on first run.

## What this adds
- `scripts/check-migrations.mjs` — fails on any duplicate number, gap, or malformed filename; reports the next free number.
- Wired into the **CI lint job** + `test:ci` (local parity).
- `migrations/**` added to the CI path filter so migration-only PRs actually trigger it.

## The two pre-existing duplicates
`0015` and `0017` are **already applied in production** (D1 records applied migrations by filename, so renaming either would make wrangler try to re-apply it). They're **grandfathered** via an explicit allowlist. Any *new* duplicate now fails CI and must be renumbered before merge — which also surfaces the conflict the next contributor would otherwise miss.

Net: the next migration is unambiguously `0038`, and this class of silent deploy-breaker can't merge again.


ghost:

Reviewed **3** changed file(s) — two independent AI reviewers.

**Changed files:** `.github/workflows/ci.yml`, `package.json`, `scripts/check-migrations.mjs`

**Suggested action:** 🛠️ **Request changes.** _(reviewers split: merge / request changes)_ Address the suggestions below before merging.

The PR adds a Node script that validates D1 migration filenames for proper numbering, duplicate detection, and gaps, integrates it into the CI workflow, and exposes an npm script for local use. The changes are straightforward, safe, and improve CI reliability.

_Suggestions_
- Add a unit test for `scripts/check-migrations.mjs` to ensure future modifications don’t break its validation logic.
- Consider handling nested directories or non‑SQL files gracefully in case the migrations folder structure changes.
- Document the script’s purpose and usage in the repository README or contributing guide.

_Worth double-checking_
- If the current migrations contain any unexpected gaps or duplicate patterns not accounted for, CI will now fail; verify the repository passes the new check before merging.
- The script assumes all migration files are directly under `migrations/`; adding subfolders later would require updating the script.

**Reviewer B · `nemotron-3-120b-a12b`** — recommends 🛠️ `request changes`
The PR adds a CI guard to prevent silent duplicate migrations and gaps in migration numbering. It introduces a new validation script, updates the CI workflow to run it on migration changes, and adds an npm script. The implementation correctly handles the two pre-existing grandfathered duplicates (0015, 0017) and enforces contiguous numbering. The change is safe and addresses a real deployment risk.

_Suggestions_
- Add unit tests for scripts/check-migrations.mjs to verify behavior (valid migrations, duplicates, gaps, malformed filenames, and grandfathered cases).
- Consider adding a test case for the scenario where migrations directory is empty (though currently unlikely given existing migrations).

_Worth double-checking_
- The gap detection logic assumes numbers are contiguous starting from the lowest migration number; verify this aligns with D1's expectations.
- Ensure the grandfathered duplicate logic cannot be accidentally bypassed (e.g., by adding a third file for 0015 or 0017).

PR #1223, opened 2026-06-24
fix(gate): route draft-dodge close through agent executor

JSONbored:
### Motivation
- The converted-to-draft "draft-dodge" handler previously posted a comment and directly closed PRs, which bypassed the agent action executor's safety stack (global pause, dry-run, per-action autonomy, approval staging, and permission readiness), creating an authorization/behavior mismatch.
- The change ensures draft-conversion close attempts obey the same safety and audit semantics as other automated maintainer actions.

### Description
- Replace the direct `createIssueComment` + `closePullRequest` path with a planned `close` action routed through `executeAgentMaintenanceActions` so close-specific autonomy, approval staging, dry-run, global/per-repo pause, and permission checks run before any GitHub mutation (`src/queue/processors.ts`).
- Use `getInstallation` to supply `installationPermissions` to the executor and build a `close` `PlannedAgentAction` including `requiresApproval`, `closeKind`, and `closeComment` so the existing approval/flagging logic applies.
- Only record the `github_app.draft_dodge_closed` audit event when the executor reports a completed close, avoiding misleading audit records when the executor denies, queues, dry-runs, or errors the action.
- Add and update unit tests to cover executor error recording, label-only autonomy (close denied when only label is enabled), dry-run/global pause behavior, and staging when `close` requires approval (`test/unit/queue.test.ts`).

### Testing
- Ran the focused unit suite for the converted-to-draft handler: `npx vitest run test/unit/queue.test.ts -t "converted_to_draft gate-close"`, which passed (tests for the block/close branches and new regression cases succeeded).
- Type-check verified with `npm run typecheck` and succeeded.
- Lint/checks verified locally with `git diff --check` (no whitespace/conflict marker issues).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3badfcfc6c832cb4185c8b7370706c)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -1584,39 +1584,50 @@ async function processGitHubWebhook(env: Env, deliveryId: string, eventName: str
       await persistAdvisory(env, advisory);
       // Draft-dodge guard (#converted-to-draft): a contributor converting an OPEN PR to draft cannot use
       // draft state to keep a gate-rejected PR alive. When a prior gate failure exists for the PR's current
-      // headSha (and the block has not been maintainer-overridden), close the PR immediately — the gate
-      // verdict stands and does not reset on draft conversion. Skipped when the agent is unconfigured or
-      // paused (the gate doesn't act on paused repos) and for owner / automation PRs.
-      if (
-        payload.action === "converted_to_draft" &&
-        installationId &&
-        pr.headSha &&
-        pr.state === "open" &&
-        isAgentConfigured(settings.autonomy) &&
-        !settings.agentPaused &&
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded by #1247, which closes the same gap (the draft-dodge close now respects per-repo pause, the global freeze, and `agent_dry_run`) **and** also covers the `maybeRecloseDisallowedReopen` close path, which this PR did not. Your executor-routing approach is cleaner architecturally — routing both direct-close paths through `executeAgentMaintenanceActions` so they inherit close-autonomy + approval enforcement too is a worthwhile follow-up refactor; filing that separately. Thanks for catching the gap.

PR #1226, opened 2026-06-24
fix(github): continue paginated issue-events when Link has rel="next" but no rel="last"

JSONbored:
### Motivation
- A regression treated a Link header lacking `rel="last"` as a single-page result, which can skip later issue-event pages when the header contains `rel="next"`, causing `getLastCloserLogin` to miss a later maintainer/bot close.

### Description
- Change `issueEventsLastPage` to return `number | null` and treat the absence of `rel="last"` with a present `rel="next"` as an unknown-last sentinel (null).
- Add `scanIssueEventsWithoutLastPage(requestPage, firstEvents)` and `issueEventsHasNextPage` to perform a bounded forward scan when the last page cannot be determined, preserving the most-recent closer seen across the scanned pages.
- Wire `getLastCloserLogin` to invoke the bounded forward scan when the sentinel is returned, while keeping the existing newest-window behavior when `rel="last"` is present.
- Add unit tests that cover the regression (page-2 maintainer close found when page 1 only had `rel="next"`) and a single-page case when the header contains neither `next` nor `last`.

### Testing
- `npm run typecheck` succeeded with no errors.
- `npx vitest run test/unit/github-pr-actions.test.ts` succeeded (all tests in that file passed).
- `npm run test:coverage` was attempted but coverage report generation failed in this environment due to a V8/coverage tooling error (`TypeError: jsTokens is not a function`), so full coverage verification is currently blocked.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3badfe03ec832c9d2c696ae35028d9)

### DIFF
--- a/src/github/pr-actions.ts
+++ b/src/github/pr-actions.ts
@@ -127,6 +127,7 @@ export async function getLastCloserLogin(env: Env, installationId: number, repoF
     const firstResponse = await requestPage(1);
     const firstEvents = firstResponse.data as Array<{ event?: string; actor?: { login?: string | null } | null }>;
     const lastPage = issueEventsLastPage(firstResponse.headers.link);
+    if (lastPage === null) return scanIssueEventsWithoutLastPage(requestPage, firstEvents);
     if (lastPage <= 1) return latestCloserInPage(firstEvents) ?? null;
 
     // GitHub returns issue-events oldest-first. Use the Link header to inspect the newest bounded window instead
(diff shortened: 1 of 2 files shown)


JSONbored:
Re-implemented fresh on current `main` as #1250. Your diagnosis was right — `issueEventsLastPage` returning 1 when `rel="last"` is absent makes a paginated-without-`rel="last"` timeline read only page 1 while claiming full coverage. #1250 integrates the `rel="next"` forward-scan with the `coveredAllPages` fail-closed contract that landed after this PR was branched (so an un-terminated `rel="next"` now fails closed rather than silently passing). Thanks for catching it.

Issue #561, opened 2026-06-10
feat(signals): slop signal — generated/vendored/dependency/docs matchers


JSONbored:
Parent roadmap: #525
Parent epic: #530

## Goal
Implement path-matcher signals: generated/vendored/lockfile/min edits, dependency-bump-only, docs-only padding.

## Reuse
New pure path matchers (sibling to `isTestFile`).

## Acceptance
Each matcher classifies representative paths correctly.

## Testing
`npm run test:ci`, 97%+ coverage. Public/private output boundary preserved; deterministic where applicable.

PR #1261, opened 2026-06-24
feat(signals): classify automation and CI toolchain config in path-matchers

jason020818:
## Summary

Extends `isConfigFile` / `classifyChangedFile` so slop signals stop miscounting common automation and hosted-CI config as uncategorized source effort (#561).

**New config recognition:**
- Dependency automation: `renovate.json`, `dependabot.yml`
- Toolchain pins: `.tool-versions`, `mise.toml`
- Git hooks: `lefthook.yml` / `.lefthook.yaml`, `.pre-commit-config.yaml`
- Hosted CI: `.gitlab-ci.yml`, `azure-pipelines.yml`, `.github/workflows/*.yml`, `.circleci/config.yml`
- Protobuf codegen: `buf.yaml`, `buf.gen.yaml`

Path-only, deterministic, full unit coverage + negative cases.

## Test plan

- [x] `npm test -- test/unit/path-matchers.test.ts`


JSONbored (approved):
This is the earlier of the two path-matcher PRs and it's correct — workflow/circleci/dependabot regexes plus the common toolchain basenames, with a real "renovate-like source names aren't config" guard. It wins the tiebreak over the later #1422 by being the original, not because either is "better." It's a jaso0n0818 fork so I won't touch it — your call to take it; if you do, #1422's extra lockfile/config entries can be folded in after.

PR #1046, opened 2026-06-22
feat(signals): add test-coverage classification to test-evidence

jason020818:
Add `classifyTestCoverage()` which maps a list of changed paths to a `TestCoverageClassification` band — `strong` (≥40% test ratio), `adequate` (≥20%), `weak` (>0% but below 20%), or `absent` (no test files at all).

Slop signals currently detect the presence of any test file but have no lightweight way to express how well-covered a change is relative to its source footprint. This helper fills that gap: callers pass the full list of changed paths and receive a single band they can include in advisory output or weight against the contribution score.

PR #1262, opened 2026-06-24
feat(signals): recognize e2e, integration, and snapshot test paths

jason020818:
## Summary

Extends `isTestPath` / `classifyTestCoverage` so slop and contributor evidence recognize modern e2e and snapshot test layouts (#1046 follow-up).

**New test path recognition:**
- `e2e/`, `integration/`, `playwright/`, `cypress/` directories
- Cypress-style `*.cy.ts` / `*.cy.js` and `*.e2e.ts` suffixes
- Jest/Vitest `__snapshots__/` trees and `*.snap` files

Pure path-only matchers with unit coverage.

## Test plan

- [x] `npm test -- test/unit/test-evidence.test.ts`


JSONbored (approved):
Extends the test-path matcher; composes with #1306 rather than duplicating it (different function). Keying snapshots on the __snapshots__/ dir not a bare .snap is right. Hands-off — owner executes.

PR #1108, opened 2026-06-23
feat(github-app): add reviewer routing auto-request mode

statxc:
## Summary

- Implements #830 by adding CODEOWNERS-based reviewer routing with `reviewerRoutingMode: off | advisory | auto_request`
- Surfaces reviewer suggestions in the PR review output
- Auto-requests top-ranked individual reviewers only when `reviewerRoutingMode` is `auto_request`, while skipping teams, already-requested reviewers, and first-time external contributors
- Adds the required settings, API/OpenAPI plumbing, migration, and focused tests for routing, idempotency, newcomer guards, and gate-override safety
- Includes the follow-up CODEOWNERS matcher fix so directory and shallow-wildcard rules align with GitHub's documented semantics

## Scope

- [x] The PR title follows `type(scope): short summary` Conventional Commit format, for example `fix(api): restore profile access checks`.
- [x] This PR is focused and does not mix unrelated backend, UI, MCP, docs, dependency, and deploy changes.
- [x] This follows `CONTRIBUTING.md` and does not reintroduce GitHub Pages, VitePress, `site/`, or `CNAME`.
- [x] I linked an issue, or this is small enough that the summary explains why an issue is not needed.

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage`
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:openapi:check`
- [x] `npm run ui:version-audit`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:test`
- [x] `npm run ui:build`
- [x] `npm audit --audit-level=moderate`
- [x] New or changed behavior has unit/integration tests for new branches, fallback paths, and sanitizer boundaries

If any required check was skipped or failed, explain why:

- I ran `npm run test:ci` locally against the final PR tree.
- `ui:lint` passes, but it emits existing `react-refresh/only-export-components` warnings in route files outside this PR's scope.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed.
- [x] Public GitHub text stays sanitized, low-noise, and does not imply compensation guarantees or optimization tactics.
- [x] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests.
[...]

Issue #830, opened 2026-06-17
feat(github-app): reviewer-routing auto_request action (#540 follow-up)


JSONbored:
Follow-up to #540 (advisory reviewer routing, shipped in #829). #540's advisory core surfaces ranked CODEOWNERS reviewer suggestions in the PR panel; this issue adds the **outward-facing `auto_request` action** that was intentionally deferred because it actually assigns reviewers on real PRs.

## Why split out
#829 shipped `reviewerRoutingMode` as `off | advisory`. Actually requesting reviewers via the GitHub API is outward-facing and warrants its own focused review (and the strict opt-in guards the parent issue calls for). The setting column is **TEXT**, so this adds the enum value + action with **no migration**.

## Deliverables
- [ ] Add `auto_request` to the `reviewerRoutingMode` union (`src/types.ts`, `parseReviewerRoutingMode` in `src/db/repositories.ts`, the openapi enum, settings-preview, and the `.gittensory.yml` settings-block enum in `focus-manifest.ts`).
- [ ] When `reviewerRoutingMode === "auto_request"`, call GitHub's request-reviewers API for the top-ranked `buildReviewerRouting` suggestion(s) — using the existing installation-token GitHub client (this is the first reviewer-routing call needing the App token, vs the public raw fetch the advisory core uses).
- [ ] **Strict opt-in guard (acceptance criterion):** never auto-request for a **first-time external contributor** without explicit opt-in. Reuse the newcomer detection already threaded for #552 (0 merged PRs in this repo) so a genuine newcomer's PR is never auto-assigned.
- [ ] Skip teams / already-requested reviewers; idempotent (don't re-request on every webhook).
- [ ] Tests: the opt-in guard (newcomer → no auto-request), the action fires only under `auto_request`, idempotency.

## Context
`src/signals/reviewer-routing.ts` (the ranked suggestions), `src/queue/processors.ts` `maybePublishPrPublicSurface` (where routing is built), the newcomer detection at the gate-policy threading (#552). Relates #540, #829, #525/#528.

JSONbored:
2026-07-01 issue audit: Off-thesis: reviewer-routing auto-request targets large teams; low value for solo/small self-host maintainers, and advisory base #829 is unbuilt.

PR #829, opened 2026-06-17
feat(signals): advisory reviewer routing via CODEOWNERS (#540)

JSONbored:
Part of #540 (advisory core; `auto_request` deferred — see follow-up).

## What
Suggest reviewers from a repo's CODEOWNERS for a PR's changed files, ranked by ownership and de-weighted by each owner's current load — **advisory only**, surfaced in the maintainer-private PR panel. Opt-in via `reviewerRoutingMode` (`off|advisory`, default off).

## How
- **`src/github/codeowners.ts`** — pure `parseCodeowners` + `matchCodeowners` implementing GitHub's CODEOWNERS semantics: **last-match-wins**; a pattern with **any** slash (leading/internal/trailing) is root-anchored, a slash-free pattern matches anywhere; `dir/` owns files under the dir (not a file named `dir`); `*` stays within a segment, `**` crosses segments. `loadRepoCodeowners` fetches `.github/CODEOWNERS` → `CODEOWNERS` → `docs/CODEOWNERS` via `raw.githubusercontent.com/.../HEAD/...` (mirrors the focus-manifest fetch), best-effort, `[]` on any failure.
- **`src/signals/reviewer-routing.ts`** — pure `buildReviewerRouting` ranks individual owners by matched-file count, then **load band** (busy = ≥ 3 authored open PRs — the only available proxy, since PR records carry no requested-reviewer data), then login. **Excludes the PR author**, collects `@org/team` handles separately, caps at 5. **Public-safe**: output is login + file count + band + reason only — no trust/score/reward fields (asserted in tests).
- **Setting** `reviewerRoutingMode` wired like `mergeReadinessGateMode` (types/schema/**migration 0042**/repositories/openapi/settings-preview) + `.gittensory.yml` `settings:` block parity. The gate path **reuses the shared changed-file load** (no double fetch); routing is advisory and **never feeds the gate**.
- **Panel** gains a "Suggested reviewers (advisory)" section via `sanitizePanelText`.

## Deferred (follow-up)
`auto_request` — actually assigning reviewers on GitHub — is **outward-facing** and intentionally split out for its own focused review (with the strict opt-in + first-time-external-contributor guard the issue calls for). The column is TEXT, so adding the value + action later needs **no migration**.

## Tests
[...]

PR #1281, opened 2026-06-24
feat(github): add CODEOWNERS parser and matcher

statxc:
## Summary
- add pure CODEOWNERS parsing and matching helpers
- cover GitHub-documented wildcard and escaping behavior, including `*`, `?`, `**`, escaped literals, and escaped whitespace
- keep this PR limited to the parser/matcher foundation only

## Why
This is the first PR in the reviewer-routing split for #830.

The original implementation was opened as #1108, but that PR mixed multiple concerns together. After maintainer feedback, it was split into a smaller stack so the pure CODEOWNERS foundation can be reviewed independently before layering routing, settings/UI, and the live auto-request write path.

This PR intentionally contains only the parser/matcher base.

## Scope
Included here:
- `src/github/codeowners.ts`
- `test/unit/codeowners.test.ts`

Intentionally not included here:
- reviewer ranking
- settings, schema, migration, or UI wiring
- GitHub reviewer-request writes

## Validation
- `npm run typecheck`
- `npx vitest run test/unit/codeowners.test.ts`

## Follow-ups
- ranking logic
- advisory settings and PR-panel rendering
- focused `auto_request` PR with the dry-run-safe write path


JSONbored (commented):
Closing this — and I want to be clear it's a direction call, not a knock on the work. The parser is clean, fully covered, and is exactly the foundation-first split we asked for after #1108, so you did the right thing here.

The reason is scope, above this PR: we're not pursuing reviewer-routing right now. The advisory core this builds toward (#829) was closed without landing, the full implementation (#1108) was closed, and nothing in that line is on `main` — so #830's `auto_request` would mean rebuilding the whole stack from the core up, and that's not where the project is focused (anti-gaming, contributor governance, Orb/self-host). The value of an outward-facing auto-request write path doesn't outweigh that maintenance surface for us today.

Feel free to focus on any other areas you notice being worked on, plenty of bugs/optimizations/etc to dig through in the codebase 👍🏻

Issue #1255, opened 2026-06-24
Orb: fleet calibration collector + analytics (retire per-instance App; feed from review_audit)


JSONbored:
## Summary / decision

Orb stops being a **per-instance GitHub App** and becomes a **central, anonymized fleet-calibration collector + analytics layer**. Gittensory cloud stays the owner's review app for our own repos; Orb aggregates de-noised calibration telemetry **up** from maintainers' self-hosted instances.

The key correction: every self-hosted instance **already** records rich, de-noised ground truth in `review_audit` (migration `0049`) via `processGitHubWebhook` → `recordPrOutcome` + `recordReversalSignals`:
- `gate_decision` — the prediction (`merge`/`close`/`hold`)
- `pr_outcome` — ground truth (`merged`/`closed`), author-self-close-filtered
- `reversal_reopened` / `reversal_reverted` — the human-override signals

Orb's current `orb_events` model is a **noisier reinvention** of this (raw merged/closed, no reversals, broken `review_targets` verdict lookup, wrong `head_sha` dedup key) and the `/orb/setup` per-instance App wizard is exactly the per-instance custom work we want to avoid. So we **retire** that, **repoint** the exporter at `review_audit`, **extend** the collector schema + analytics, and keep the anonymized fleet-export (one central store, instances contribute, no shared secret — the `das-github-mirror` principle).

PR #1254 (merged) already shipped the live collector URL (`https://gittensory-api.aethereal.dev/v1/orb/ingest`), the `strict` per-IP rate-limit on `/v1/orb/ingest`, and the no-shared-secret `.env.example` framing. This work builds on it.

## 1. What Orb collects (one anonymized row per resolved PR)

Sourced from `review_audit` (+ `review_targets` for open-time), **not** `orb_events`:

| Field | Source | Anonymization |
|---|---|---|
| `instance_id` | `SHA256(ORB_APP_ID).slice(0,16)` | one-way hash, no PII |
| `repo_hash` | `HMAC(repo_full_name, ORB_WEBHOOK_SECRET).slice(0,24)` | keyed by the instance's **own** secret → collector cannot de-anonymize |
| `pr_hash` | `HMAC("${repo}#${pr}", secret).slice(0,24)` | same |
| `gate_verdict` | `gate_decision.decision` (`merge`/`close`/`hold`) | none (non-PII) |
| `outcome` | `pr_outcome.decision` (`merged`/`closed`) | none |
[...]

PR #1289, opened 2026-06-24
fix(orb): gate fleet cycle percentiles

JSONbored:
### Motivation

- Prevent untrusted or unregistered senders from poisoning fleet cycle-time percentiles reported by Orb analytics.  The ingest endpoint is intentionally unauthenticated and previously stored clamped but attacker-controlled `time_to_close_ms` for any `instance_id`, which could skew `cycleP50Ms`/`cycleP95Ms`.
- Align cycle-time percentile computation with the existing registration + volume eligibility model used for other fleet metrics so only operator-registered, eligible instances influence fleet calibration.

### Description

- Change the cycle-time query to only select rows from registered instances by joining `orb_signals` to `orb_instances` with `i.registered = 1` and return `instance_id` + `ms` rows. 
- Introduce a `CycleTime` interface and collect `cycleRows`, then filter those rows to the same eligible instance set (`decided >= MIN_DECIDED` and registered) before computing percentiles. 
- Compute `cycleP50Ms` and `cycleP95Ms` from the filtered cycle array instead of the unfiltered store. 
- Add a regression unit test (`test/unit/orb-analytics.test.ts`) that demonstrates unregistered high-cycle signals and registered-but-below-volume senders cannot move fleet cycle percentiles.

### Testing

- Ran `npx vitest run test/unit/orb-analytics.test.ts` and the unit test suite for `orb-analytics` passed. 
- Ran `npm run typecheck` and `git diff --check` which passed. 
- Attempted `npm run test:coverage` and `npm run test:ci` locally but coverage remapping failed with `TypeError: jsTokens is not a function` due to an invalid local `js-tokens@4.0.0` in `node_modules` (dependency `ast-v8-to-istanbul` expects `^10.0.0`), and `npm ci` / `npm audit` could not repair the install because registry fetches returned `403 Forbidden` in this environment, so the full repo gate did not complete here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c64cc7a24832ca0aeab31d99ba86c)


JSONbored:
Verdict: MERGE · gittensor:bug. Gates cycle percentiles to registered + decided instances (the one metric a stranger could still poison after #1255); dual SQL+JS gate, adversarial test, green. On-direction (Orb).

PR #1356, opened 2026-06-25
fix(agent): preserve maintainer-configured AI-review hard blockers (do not auto-refute on green CI)

JSONbored:
### Motivation
- Prevent green CI from automatically downgrading a maintainer-configured AI hard-block (e.g. `ai_consensus_defect` / `ai_review_split`) into a mergeable success, which permitted an authorization/policy bypass in the automated merge planner.
- Keep the public PR comment and the disposition in lock-step so a maintainer-configured gate failure remains visible and actionable.
- Maintain conservative, defense-in-depth behavior: CI is necessary for merges but not a security oracle that overrides maintainer policy about AI blockers.

### Description
- Use the gate's configured `conclusion` directly in the auto-maintenance planner so AI-only failures remain blocking instead of being auto-refuted by green CI; removed the downgrade path. (`src/settings/agent-actions.ts`)
- Stop forwarding the CI-refutation inputs into the planner and stop calling the repo-level refutation gate from the processor so the disposition no longer treats green CI as an automatic refutation. (`src/queue/processors.ts`)
- Keep the public comment rendering aligned with the raw gate verdict by making `reconcileGateEvaluationForGreenCi` a no-op that preserves the original evaluation instead of clearing AI blockers when CI is green. (`src/rules/advisory.ts`)
- Update unit tests to assert that `ai_consensus_defect` and `ai_review_split` remain blocking on green CI and that public comment behavior preserves the raw failure; adjusted imports/wiring to reflect removed refutation wiring. (tests modified: `test/unit/agent-actions.test.ts`, `test/unit/rules.test.ts`)
- Files changed: `src/settings/agent-actions.ts`, `src/rules/advisory.ts`, `src/queue/processors.ts`, `test/unit/agent-actions.test.ts`, `test/unit/rules.test.ts`.

### Testing
- Ran targeted unit tests: `npm test -- --run test/unit/agent-actions.test.ts test/unit/rules.test.ts test/unit/grounding-wiring.test.ts`, and the modified test cases passed (the three target test files all passed).
- Ran full typecheck with `npm run typecheck` which succeeded.
- Ran `git diff --check` (hygiene) which succeeded.
- Attempted full coverage with `npm run test:coverage`; the run exercised many suites but the coverage remapping step failed with an upstream coverage remap error (`TypeError: jsTokens is not a function`), so the full coverage job did not complete successfully in this environment.
- `npm audit --audit-level=moderate` could not complete due to the npm registry audit endpoint returning `403 Forbidden` in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cd795a624832ca2b774faaba20e3e)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -179,7 +179,7 @@ import { resolveRepositorySettings } from "../settings/repository-settings";
 import type { LocalBranchAnalysisInput } from "../signals/local-branch";
 import { runGittensoryAiReview } from "../services/ai-review";
 import { secretLeakFinding } from "../review/safety";
-import { aiCiRefutationActive, buildReviewGroundingText, checkSummaryText as checkFailureSummaryText, isGroundingEnabled } from "../review/grounding-wire";
+import { buildReviewGroundingText, checkSummaryText as checkFailureSummaryText, isGroundingEnabled } from "../review/grounding-wire";
 import { buildReviewRagContext, isRagEnabled } from "../review/rag-wire";
 import { indexRepo, reindexChangedPaths } from "../review/rag-index";
 import { isReputationEnabled, recordReputationOutcome, shouldSkipAiForReputation } from "../review/reputation-wire";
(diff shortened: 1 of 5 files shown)


JSONbored:
Closing — not the right shape to merge.

This PR removes the AI-CI-refutation entirely (both the disposition path in `planAgentMaintenanceActions` and the public-comment `reconcileGateEvaluationForGreenCi`), so an AI-judgment-only defect would close a green-CI PR. That refutation is the deterministic guard against the calibration's headline problem — AI false-closes on a high share of green PRs — so a blanket removal regresses precisely what we're trying to fix.

The valid kernel here is narrower: when a maintainer *explicitly* opts into `aiReviewMode: block`, green-CI refutation arguably shouldn't override their choice. If that case proves real, it should ship as a **scoped exception** (refute by default; skip refutation only for an explicit maintainer AI-block opt-in), not a wholesale removal. Re-open as a fresh, scoped PR if/when we want that.

Also superseded by drift: the surrounding disposition logic has moved since this was opened (held-vs-closed parity, etc.), so it no longer rebases cleanly.

PR #1382, opened 2026-06-25
fix(review): close trailing dot and *.localhost SSRF host bypasses

Dexterity104:
## Summary

The SSRF host guard in `src/review/content-lane/safe-url.ts` returned a public verdict for several named loopback hosts, because its checks only matched the bare forms. Two gaps let an untrusted URL pass `isSafeHttpUrl` while still resolving to loopback or an internal address:

- Trailing FQDN root dot. `new URL("https://localhost./").hostname` keeps the dot as `localhost.`, which never equals `localhost` and never ends with `.local`, so the guard read it as public. The parser strips the dot for IP literals, so only named hosts leaked. The same held for `foo.local.` and `bar.internal.`.
- The RFC 6761 `*.localhost` namespace. The guard matched bare `localhost` but not `db.localhost` or `foo.bar.localhost`, which systemd-resolved and browsers (including the visual lane headless Chrome) resolve to loopback.

The fix normalizes the host once at the shared chokepoint `hostIsPrivateOrLocal`: lowercase, strip the trailing dot run, then reject `localhost`, the `*.localhost` namespace, `.local`, and `.internal`. Both `isSafeHttpUrl` and `isSafeEndpointUrl` inherit the hardening through that single function, and the duplicate `.toLowerCase()` at both call sites was removed so normalization lives in one place.

Scope note: the trailing dot case was the reported gap. The `*.localhost` case is the same class of named loopback bypass on the same line, so both are closed together. The behavior was confirmed in Node and in the Workers runtime, and public hosts such as `example.com.` stay allowed.

## Scope

- [x] Linked the tracking issue, or this fix is self-contained and the summary explains why a separate issue is not required.

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage`
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`

- No checks were skipped. The whole gate ran green through `npm run test:ci` plus `npm audit --audit-level=moderate`. `npm run test:coverage` reports 100% lines and branches on `safe-url.ts`, with every changed line and branch arm exercised. Two negative path regression tests were added (the `*.localhost` namespace and the trailing dot forms); each fails before the fix and passes after.

## Safety

## UI Evidence

Not applicable. This change is backend only and has no visible UI surface, so no screenshots are required.

## Notes

- The guard feeds server side fetches in `source-evidence.ts` (rechecked per redirect hop) and the screenshot navigation guard in `visual/shot.ts`, so the hardening covers both the HTTPS fetch path and the secure WebSocket endpoint path.
- Normalization was consolidated into `hostIsPrivateOrLocal`, which keeps a single defensive point for the lowercase plus trailing dot strip and avoids the duplicate host normalization that existed at the two call sites.

### DIFF
--- a/src/review/content-lane/safe-url.ts
+++ b/src/review/content-lane/safe-url.ts
@@ -1,12 +1,14 @@
 // SSRF-safe URL guard (content-lane primitive).
 //
-// SELF-CONTAINED NATIVE PORT (reviewbot→gittensory convergence). Byte-faithful to reviewbot's
+// SELF-CONTAINED NATIVE PORT (reviewbot→gittensory convergence). Ported from reviewbot's
 // core/source-url.ts isSafeHttpUrl + isSafeEndpointUrl (the host/IP guard, including the encoded-IP
-// decoding that a dotted-quad regex misses). PURE — no imports, no I/O.
+// decoding that a dotted-quad regex misses), hardened so a trailing-dot or `*.localhost` host can't
+// dodge the loopback check. PURE — no imports, no I/O.
 //
-// Rejects non-HTTPS (isSafeHttpUrl), localhost/.local/.internal, and private/loopback/link-local
-// IPs in any literal notation (decimal `2130706433`, hex `0x7f000001`, octal, short `127.1`, and the
-// IPv6 forms). isSafeEndpointUrl additionally permits wss:/ws: for base-layer chain endpoints.
+// Rejects non-HTTPS (isSafeHttpUrl), localhost / `*.localhost` / .local / .internal, and private/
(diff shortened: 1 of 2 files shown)


JSONbored (approved):
Verdict: MERGE · gittensor:bug (security). Strips the FQDN root dot + adds *.localhost, closing localhost./db.localhost loopback bypasses at the one chokepoint, green.

PR #1391, opened 2026-06-25
fix(review): isolate review.profile from consensus prompts

JSONbored:
### Motivation
- A recent change threaded `.gittensory.yml` `review.profile` into the single system prompt used for both advisory write-ups and block-mode consensus reviewers, allowing a maintainer presentation knob to influence gate-producing consensus outputs.
- `review.profile` is intended to be PRESENTATION ONLY and must never affect the consensus-defect path that the gate consumes.

### Description
- Split system-prompt construction by adding an `options.includeProfile` flag to `buildSystemPrompt` so callers can opt out of appending the `review.profile` suffix.
- Compute two prompts in `runGittensoryAiReview`: `advisorySystem` (includes `review.profile`) and `consensusSystem` (explicitly excludes the profile), and route advisory/BYOK calls to `advisorySystem` and block-mode Workers-AI consensus calls to `consensusSystem`.
- Adjust neuron estimation to account for the separate advisory/consensus prompts so grounding/path context is still billed without letting `review.profile` change consensus accounting.
- Add a regression unit test that proves an `assertive` profile affects the advisory prompt but not the two block-mode consensus prompts.

### Testing
- Ran `npx vitest run test/unit/ai-review.test.ts --reporter=verbose`, which passed (1 file, 41 tests, all green) including the new regression test.
- Ran `npm run typecheck`, which completed successfully with no type errors.
- Ran `git diff --check` to ensure no whitespace/conflict markers; it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d0c3c515c832cae85e65baaaee403)

### DIFF
--- a/src/services/ai-review.ts
+++ b/src/services/ai-review.ts
@@ -438,15 +438,21 @@ const INLINE_FINDINGS_SUFFIX =
   '\n\nINLINE FINDINGS: ALSO include an additional top-level field "inlineFindings" in the SAME JSON object — an array (possibly empty) of your most important findings, each anchored to a specific changed line, for inline PR comments. Each item: {"path": the changed file path EXACTLY as shown in the diff, "line": the 1-based line number in the NEW file (count forward from the "+" start in the nearest "@@ -old +new @@" hunk header) of an ADDED ("+") line you are commenting on, "severity": "blocker" or "nit", "body": the one-sentence finding}. Include ONLY findings you can place on a specific added line; OMIT any you cannot anchor precisely (a wrong line is worse than none). At most ~10 items.';
 
 /** The effective reviewer SYSTEM prompt. Appends the grounding-discipline suffix when the caller supplied one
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this guards a path the prompt already prevents. The `assertive` profile raises extra items as **nits** (`ai-review.ts:51,55`: "defensive or speculative hardening … is a NIT, not a blocker"), and the consensus gate only fires when **both** reviewers populate `blockers[]` (`ai-review.ts:672`, `consensusDefectOf` returns `null` otherwise). So `review.profile` can't leak into the consensus verdict. The fix also abandons the #1347 advisory-leg reuse (it pays for an extra free Workers-AI consensus call whenever a profile is set), and the regression test fabricates a model that promotes a nit into `blockers[]`, which the prompt explicitly forbids. Net cost, no real protection.

PR #1395, opened 2026-06-25
fix(orb): revalidate relay DNS before forwarding

JSONbored:
### Motivation
- The relay registration guard only checked URL syntax and the textual hostname, which permits DNS-based rebinding to private/loopback addresses at fetch time and creates a blind SSRF primitive when `forwardOrbEvent` posts stored relay URLs.
- Forwarding must therefore revalidate the resolved addresses at send time to prevent attacker-controlled hostnames from resolving to internal addresses.
- The change preserves the existing best-effort/fail-safe behavior so a forwarding failure never breaks the Orb webhook 202 path.

### Description
- Add forward-time DNS resolution for relay targets using a pluggable resolver and DNS-over-HTTPS lookups to obtain A/AAAA answers (`resolveRelayHostname`, `relayDestinationIsSafe`) in `src/orb/relay.ts`.
- Validate each resolved address with the existing `isSafeHttpUrl` logic (via `resolvedAddressIsSafe`) and reject destinations with no answers or any private/loopback/link-local addresses before attempting the POST.
- Keep `forwardOrbEvent` fail-safe by returning `"failed"` when a destination is unsafe or unresolvable instead of throwing, and keep the HMAC-signing/POST flow unchanged for safe destinations.
- Add regression tests in `test/integration/orb-relay.test.ts` to cover DNS revalidation, rejection of loopback/private-address resolutions, no-DNS-answer behavior, and a positive DNS-checked path; make the resolver pluggable for testing.

### Testing
- Ran `npx vitest run test/integration/orb-relay.test.ts` and the integration suite passed (21 tests passed).
- Ran `npm run typecheck` and TypeScript checked clean with no errors.
- Attempted coverage with `npx vitest run test/integration/orb-relay.test.ts --coverage.enabled true --coverage.reporter text`, but V8 coverage remapping failed after the tests passed with `TypeError: jsTokens is not a function` in `ast-v8-to-istanbul`; the functional tests themselves passed despite the coverage tooling error.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d065d33c0832c84b865bee259b610)

### DIFF
--- a/src/orb/relay.ts
+++ b/src/orb/relay.ts
@@ -8,6 +8,36 @@ import { hashToken } from "../auth/security";
 import { isSafeHttpUrl } from "../review/content-lane/safe-url";
 import { decryptSecret, encryptSecret } from "../utils/crypto";
 
+type RelayDnsResolver = (hostname: string) => Promise<string[]>;
+
+function addressAsUrlHost(address: string): string {
+  return address.includes(":") ? `[${address.replace(/^\[|\]$/g, "")}]` : address;
+}
+
+function resolvedAddressIsSafe(address: string): boolean {
+  return isSafeHttpUrl(`https://${addressAsUrlHost(address)}/`);
+}
+
+async function resolveRelayHostname(hostname: string): Promise<string[]> {
+  if (/^\d+\.\d+\.\d+\.\d+$/.test(hostname) || hostname.includes(":")) return [hostname];
+  const answers: string[] = [];
+  for (const type of ["A", "AAAA"]) {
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — `codecov/patch` is red, and the fix is only partial. `forwardOrbEvent` still POSTs to `row.relay_url` (the hostname, `relay.ts:167`), not the address it just validated, so a fast DNS rebind between the DoH check and the Workers `fetch()` re-resolution leaves the TOCTOU open — while adding a DNS round-trip to every forwarded event. The gap is real but low-severity (a relay can only target its own enrollment's forward path). If we want to close it properly, the fix is resolve-and-pin (or a private-target-blocking fetch), not re-resolve-then-POST-the-hostname.

PR #1416, opened 2026-06-26
fix(orb): fall back to default expiry when broker token expiresAt is unparseable

nickmopen:
## Summary

`fetchBrokeredInstallationToken` (`src/orb/broker-client.ts`) returns `expiresAtMs`, which the installation-token cache in `src/github/app.ts` uses to decide freshness:

```ts
if (cached && cached.expiresAtMs - TOKEN_SAFETY_MARGIN_MS > Date.now()) return cached.token;
```

The exchange defaulted a **missing** `expiresAt` to a ~50 min expiry, but a **present-but-malformed** `expiresAt` fell through to `Date.parse(...)` → `NaN`. For `NaN`, `cached.expiresAtMs - margin > Date.now()` is always `false`, so a brokered self-host **re-mints a token on every GitHub API call** (broker `/v1/orb/token` thrash) instead of caching it for ~an hour.

## Problem

The broker token comes from an external JSON response (`POST /v1/orb/token`), so a malformed-but-present `expiresAt` is a realistic input. Only the *absent* case was defaulted; the *unparseable* case poisoned the cache key with `NaN`.

## Fix

Guard with `Number.isFinite` so a malformed expiry takes the same ~50 min fallback as an absent one. The change errs safe — a bad expiry causes a re-mint, never reuse of a stale token.

## Scope note

`src/github/app.ts:75` has the same `payload.expires_at ? Date.parse(...) : default` shape, but that value is GitHub's own installation-token response (reliable ISO 8601), not an external broker body — left untouched to keep this fix narrow to the genuinely-external input.

## No linked issue

No tracked issue exists for this; it's a self-contained correctness fix surfaced while reading the broker path. (`gate.linkedIssue` is advisory.)

## Validation

Run from repo root (Node 24):

```sh
npm run test:ci   # GATE_EXIT=0 — typecheck, full coverage, workers, MCP build/pack, UI openapi/lint/typecheck/test/build all green
```

Targeted unit + coverage on the changed file:

```sh
npx vitest run test/unit/orb-broker-client.test.ts --coverage --coverage.include='src/orb/broker-client.ts'
# 14/14 passed — broker-client.ts 100% statements/branches/functions/lines
```

Added a regression test: a present-but-unparseable `expiresAt` now yields a finite `expiresAtMs` at the ~50 min fallback (asserting no `NaN` reaches the cache).


JSONbored (approved):
Verdict: MERGE · gittensor:bug. NaN expiry poisoned the install-token cache (re-mint every call); Number.isFinite guard mirrors the GitHub-direct path, green.

PR #1426, opened 2026-06-26
fix(signals): preserve segment boundaries for globstar paths

JSONbored:
### Motivation
- Prevent `**/` rewrite from removing the path-separator and turning segment globs (e.g. `**/safe.ts`) into broad suffix globs that match attacker-controlled filenames like `unsafe.ts`.
- Retain the linear-time matching and root-level `**/` semantics while avoiding ReDoS and avoiding overbroad exclusions used by `review.exclude_paths`.

### Description
- Replaced the previous global `**/`→`*` rewrite with an `expandGlobstarSlash` routine that produces explicit alternatives that preserve the `/` segment when appropriate. (`src/signals/focus-manifest.ts`)
- Compile-time matchers are now built as a set of alternatives and test each alternative with the existing linear matcher; exact alternatives are tested with a simple equality predicate for efficiency. (`src/signals/focus-manifest.ts`)
- Added a protective cap (`MAX_GLOBSTAR_SLASH_ALTERNATIVES`) to bound repeated `**/` expansion so matcher expansion remains safe and performant. (`src/signals/focus-manifest.ts`)
- Added regression tests that assert correct `**/` semantics (basename-only matches do not match suffix collisions), ensure `excludeReviewPaths` no longer drops attacker-named files, and exercise bounded repeated `**/` expansion. (`test/unit/focus-manifest.test.ts`)

### Testing
- Ran the focused unit suite with `npm test -- --run test/unit/focus-manifest.test.ts`, and all tests in that file passed.
- Ran static type checks with `npm run typecheck`, and `tsc --noEmit` passed.
- Ran `git diff --check` and it reported no issues.
- Attempted coverage with `npm run test:coverage -- --run test/unit/focus-manifest.test.ts`; the tests passed but coverage remapping failed with `TypeError: jsTokens is not a function` during coverage conversion (investigation required); this prevented completing the full `test:coverage` remap step in CI locally.
- `npm run test:ci` could not be completed locally due to the coverage remapping issue and external-network-dependent steps; `npm audit --audit-level=moderate` returned a `403` from the registry in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3dc27c4aa4832ca48973becb3b280c)


JSONbored:
Verdict: MERGE · gittensor:bug (security). **/safe.ts was matching unsafe.ts in the manifest matcher (backs review.exclude_paths); boundary preserved, blow-up bounded, both-branch tested, green.

Issue #16, opened 2026-05-28
feat(agent): make next-action recommendations repo-specific


JSONbored:
## Problem

Agent recommendations can still be too generic. A miner should not just hear “pick one narrow change”; Gittensory should explain which repo, lane, work type, or cleanup action is most useful and why.

## Desired behavior

Generate repo-specific next actions using contributor history, lane fit, queue pressure, duplicate risk, labels, language/path fit, maintainer-lane context, and prior outcome patterns.

## Implementation notes

- Start in `src/services/agent-orchestrator.ts`, `src/services/decision-pack.ts`, and the reward/risk signal builders.
- Use existing stored contributor evidence and repo signal snapshots before adding new storage.
- Keep sorting deterministic.
- Separate maintainer-owned repos from normal contributor/miner evidence.
- Make recommendations concrete: repo, lane, action type, reason, blocker, and rerun condition.

## Public/private boundaries

Private MCP/API output can explain scoreability, risk, and blocker impact. Public GitHub output must stay sanitized and avoid score estimates or private reviewability details.

## Acceptance criteria

- Agent plan names specific repos and work types to pursue or avoid.
- Maintainer-owned repo history does not inflate normal contributor-lane evidence.
- Cleanup-first actions outrank new work when open PR pressure blocks scoreability.
- Direct-PR, issue-discovery, split, and maintainer lanes produce different reasoning.
- Recommendations are deterministic for the same input data.

## Tests

- Direct-PR repo fixture.
- Issue-discovery-only repo fixture.
- Maintainer-lane fixture.
- High open-PR pressure fixture.
- Duplicate-risk fixture.
- Deterministic sorting regression.

PR #1427, opened 2026-06-26
fix(queue): reject stale gate overrides

JSONbored:
### Motivation
- Prevent a race where a maintainer's `@gittensory gate-override` could be applied to a newer, unreviewed commit pushed after the comment was issued.
- Preserve the existing fail-open fallback (use cached head when live fetch is unreadable or missing) while ensuring overrides only target the PR head the maintainer actually approved.

### Description
- Change `resolveOverrideHeadSha` to return a structured `OverrideHeadResolution` with `headSha`, `stale`, and `liveHeadSha` so callers can detect a live-vs-cached mismatch. 
- In the gate-override handler (`maybeProcessGateOverrideCommand`) skip the override when `stale === true`, record a `stale_pr_head` skip audit event, and only build the advisory / create the neutral Gate check when the cached head is still current. 
- Update unit tests to cover matching live/cached heads, the stale-head path (skip), and fail-open fallback cases. 
- Small type/test adjustments to accommodate the new return shape.

### Testing
- Ran the focused unit tests with `npm test -- --run test/unit/resolve-override-head-sha.test.ts test/unit/queue.test.ts -t "gate-override|resolveOverrideHeadSha"`, and the targeted suites passed. 
- Ran type checking with `npm run typecheck` which completed successfully. 
- Attempted coverage with `npm run test:coverage`; tests passed but coverage remapping failed in this environment with `TypeError: jsTokens is not a function` (environmental/tooling issue during Vitest V8 remapping). 
- `npm audit --audit-level=moderate` failed in this environment due to the registry audit endpoint returning `403 Forbidden` (environment limitation, not related to code changes).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3ddb0e851c832c8f1032c449b292ef)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -3249,8 +3249,9 @@ async function maybeProcessGateOverrideCommand(env: Env, deliveryId: string, pay
     await recordGateOverrideSkip(env, deliveryId, repoFullName, `${repoFullName}#${pr.number}`, actor, "stale_pr_head");
     return true;
   }
-  const prAtResolvedHead = headResolution.headSha === pr.headSha ? pr : { ...pr, headSha: headResolution.headSha };
-  const { advisory } = await buildAuthorizedPrActionAdvisory(env, repoFullName, prAtResolvedHead, settings);
+  // Not stale (the stale case returned above), and resolveOverrideHeadSha returns the cached head (pr.headSha)
+  // in that case — so the override acts on `pr` directly; there is no other head to substitute.
+  const { advisory } = await buildAuthorizedPrActionAdvisory(env, repoFullName, pr, settings);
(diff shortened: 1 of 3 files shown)


JSONbored:
Closing — this silently reverses a recently-audited security decision rather than fixing a bug. The current "override the Gate on the live PR head, not the stale cached SHA" behavior was shipped deliberately as #1408 (`3ec6bfc5`, audit item #16) because the neutral check is per-commit and posting on the stale SHA is a silent no-op. This PR flips that to "reject as stale when live≠cached." Both are defensible postures, but reversing #1408 is an explicit policy call to make in the open, not an auto-merge — and the branch is conflicting with stale CI besides. If you want stale-rejection over live-head-override, let's decide that against #1408 directly and re-cut.

PR #1560, opened 2026-06-26
fix(signals): preserve optional globstar cap semantics

JSONbored:
### Motivation
- Fix a regression where the capped expansion for repeated `**/` patterns dropped the zero-depth (root) alternative once the alternatives count hit the cap, producing false negatives for manifest path matching and allowing path-gated checks to be skipped.

### Description
- Adjust `expandGlobstarSlash` so the rewrite to `${prefix}*/` and the push of the root `prefix` occur only when `canKeepRootAlternatives` is true, preserving the optional zero-segment branch; also update the targeted regression test in `test/unit/focus-manifest.test.ts` to assert that a bounded repeated `**/` pattern still matches a valid nested path.

### Testing
- Ran the focused unit suite: `npx vitest run test/unit/focus-manifest.test.ts` and the updated "bounds repeated **/ expansion" test passed; a local `tsc --noEmit` run reported unrelated missing `@sentry/node` types in this environment and a coverage remapping step failed due to a local `js-tokens` dependency mismatch, both of which are environmental and unrelated to the logic change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3eefc76a78832cbc8d3e715c497e57)

### DIFF
--- a/src/signals/focus-manifest.ts
+++ b/src/signals/focus-manifest.ts
@@ -933,8 +933,10 @@ function expandGlobstarSlash(pattern: string): string[] {
       const canKeepRootAlternatives = count * 2 <= MAX_GLOBSTAR_SLASH_ALTERNATIVES;
       for (let altIdx = count - 1; altIdx >= 0; altIdx -= 1) {
         const prefix = alternatives[altIdx]!;
-        alternatives[altIdx] = `${prefix}*/`;
-        if (canKeepRootAlternatives) alternatives.push(prefix);
+        if (canKeepRootAlternatives) {
+          alternatives[altIdx] = `${prefix}*/`;
+          alternatives.push(prefix);
+        }
       }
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this is a reversal of intentional same-day work, not a fix. The globstar cap branch (`focus-manifest.ts:1000-1005`) and its test were authored hours earlier by #1426 (`c732bea3`, "preserve segment boundaries for globstar paths"), which deliberately fixed the over-cap result to `false`. Under the cap the two are byte-identical; they diverge only in the pathological >64-alternative case (no real-world pattern), and the PR flips #1426's test assertion `false→true` to lock in the opposite arm. Rewriting a fresh, deliberate spec on a can't-happen-in-practice edge isn't a bug fix.

PR #1582, opened 2026-06-27
fix(review): scope private review context by owner

JSONbored:
### Motivation
- The previous per-repo review-context resolver fell back to bare `{repo}/review` folders, allowing cross-owner collisions and leakage of private review prompt material to other repos.
- Review guide and skill text is sensitive operator-provided material and must be bound to the repository owner to prevent cross-tenant disclosure.

### Description
- Removed the bare repo-name fallback so `reviewContextFolders` only returns owner-qualified `
  {owner}__{repo}/review` candidates, ensuring review prompts are owner-scoped.
- Clarified the `makeLocalReviewContextReader` documentation to state that review prompt material is owner-qualified only.
- Added a regression unit test in `test/unit/private-config.test.ts` that verifies bare repo-name `review` folders are ignored for a different owner and that missing/invalid repos degrade to an empty review context.

### Testing
- Ran unit tests with `npm test -- --run test/unit/private-config.test.ts`, which passed (20 tests).
- Ran static type checking with `npm run typecheck`, which passed.
- Attempted coverage with `npm run test:coverage -- --run test/unit/private-config.test.ts`, where tests passed but coverage generation failed with `TypeError: jsTokens is not a function` from the local coverage provider.
- Attempted `npm audit --audit-level=moderate`, which was blocked by the registry audit endpoint returning `403 Forbidden`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3f785ce774832cb7368045d48353a6)

### DIFF
--- a/src/selfhost/private-config.ts
+++ b/src/selfhost/private-config.ts
@@ -84,15 +84,16 @@ export function makeLocalManifestReader(dir: string | undefined): RepoFocusManif
   };
 }
 
-/** Per-repo review-context candidate FOLDERS (relative to GITTENSORY_REPO_CONFIG_DIR): `{owner}__{repo}/review` then
- *  `{repo}/review`. Same owner/repo validation as localConfigCandidates; an invalid full name yields none. (#review-skills) */
+/** Per-repo review-context candidate FOLDERS (relative to GITTENSORY_REPO_CONFIG_DIR): only
+ *  `{owner}__{repo}/review`. Review prompt material is private and must stay owner-scoped; an invalid full name
+ *  yields none. (#review-skills) */
 function reviewContextFolders(repoFullName: string): string[] {
   const slash = repoFullName.indexOf("/");
   if (slash <= 0 || slash === repoFullName.length - 1 || slash !== repoFullName.lastIndexOf("/")) return [];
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this isn't an attacker-reachable leak, and the fix reverses recently-shipped back-compat one-sidedly. The review-context folder names come from operator-set `GITTENSORY_REPO_CONFIG_DIR` + a webhook-validated `repoFullName` (no PR-author input), and the owner-qualified `{owner}__{repo}/review` folder is already tried **first** (`makeLocalReviewContextReader`), so it always wins when present. The bare `{repo}/review` fallback you're removing was deliberately added one day earlier by #1580 (`00cff73c`) as a documented layout. And the change is self-inconsistent: `localConfigCandidates` (`private-config.ts:57`) STILL keeps the identical bare `{repo}` fallback for `.gittensory.yml`, which carries the same `review.instructions` material — so this doesn't actually close the claimed exposure, it just diverges two parallel resolvers. If we want to enforce owner-scoping, that's a deliberate policy change applied to **both** resolvers, not a one-sided patch that breaks #1580's layout.

PR #1587, opened 2026-06-27
fix(enrichment): fail closed on dependency scan truncation

JSONbored:
### Motivation
- The dependency-scanner silently truncated attacker-controlled inputs (manifest count, patch lines, dependency queries) which could hide vulnerable packages by processing only the first items and returning an apparently-successful empty result. This leads to a fail-open security-analysis bypass. 
- The orchestrator treats any normal return as `ok`, so a silent truncation produced misleading analyzer status and an incomplete brief; truncation must be explicit and surfaced as a degraded analyzer.

### Description
- Added `DependencyScanTruncatedError` and made the scanner fail closed by throwing when the manifest-file cap is exceeded, when a single file exceeds the per-file patch-line cap, or when the number of extracted dependency changes exceeds the dependency-query cap instead of silently slicing them. 
- Replaced silent early `break`/slice behavior in `extractDependencyChanges` and `scanDependencies` with explicit checks that throw `DependencyScanTruncatedError` on over-cap inputs. 
- Kept orchestrator behavior unchanged so thrown truncation errors are caught and result in `analyzerStatus = "degraded"` and `partial = true` in the brief. 
- Added regression tests in `review-enrichment/test/enrichment.test.ts` covering dependency-query bypass, manifest/patch-line truncation, that truncated scans cause degraded briefs, and that abort-signal forwarding still works when not truncated.

### Testing
- Ran unit tests for the review-enrichment package with `cd review-enrichment && npm test`, and all tests passed (`41` tests, `0` failures). 
- Ran the local gate `npm run test:ci`; it was invoked but failed early due to external tooling setup (actionlint fallback required network/DNS access), so full gate could not complete in this environment. 
- Ran `npm audit --audit-level=moderate` in this environment and the audit endpoint returned a `403` (external registry limitation), so that check could not complete here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3f7b842b68832c8c35fdc955fa9bb7)

### DIFF
--- a/review-enrichment/src/analyzers/dependency-scan.ts
+++ b/review-enrichment/src/analyzers/dependency-scan.ts
@@ -25,6 +25,13 @@ interface ScanOptions {
   limits?: ScanLimits;
 }
 
+export class DependencyScanTruncatedError extends Error {
+  constructor(reason: string) {
+    super(`dependency_scan_truncated:${reason}`);
+    this.name = "DependencyScanTruncatedError";
+  }
+}
+
 // Per-manifest line parsers. Each returns [name, version] for a `+`/`-` diff line, or null. Heuristic (line-based,
 // not a full manifest parse) — good enough to flag the deps a PR adds/bumps without resolving the whole tree.
 const NPM_RE = /^"([^"]+)"\s*:\s*"([\^~>=<\s]*[0-9][^"]*)"/;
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this reverses a deliberate policy I shipped hours earlier in #1537 (`104ac4b0`), and it's a net availability regression. #1537 intentionally bifurcated the dependency scan: a **timeout** → `degraded`/`partial`, but an over-cap input → **bounded best-effort** (`split("\n", maxPatchLinesPerFile)` scans the first N lines; `.slice(maxDependencyQueries)` caps the OSV fan-out). This PR throws `DependencyScanTruncatedError` on any over-cap input, so a single oversized-but-benign manifest patch (>500 lines — entirely plausible for a big lockfile or dep-bump diff) now discards the **entire** dependency scan and drops every CVE finding, instead of scanning the first 500 lines. It also rewrites #1537's own pinning tests (the "caps manifest files and patch lines" assertion of `["first"]` becomes `assert.throws`) with no issue link and no mention of #1537. Fail-open-bounded vs fail-closed-throw is a real maintainer tradeoff — and ideally a *configurable* one — but it's a considered decision to make against #1537, not a same-day one-shot reversal.

PR #1598, opened 2026-06-27
fix(review): drop stale files on resync failure

JSONbored:
### Motivation
- The live-head resync path could advance the stored PR `headSha` before its dependent pull-request file refresh completed, allowing stale `pull_request_files` rows from the old head to be trusted for review and merge planning.
- This created a TOCTOU where a missed `synchronize` webhook + a partial/failed file-sync could let unreviewed live heads be auto-approved/merged.
- The change prevents stale file rows from surviving a failed detail refresh so subsequent review logic must re-fetch live files or degrade safely.

### Description
- Import and use `deletePullRequestFiles` and, during resync, only treat the detail refresh as authoritative when it reports `status: "complete"`.
- Change `reReviewStoredPullRequest` so that after upserting the live PR we attempt `refreshPullRequestDetails`, and when that refresh is not `complete` we delete the `pull_request_files` rows for the PR to avoid trusting stale file data.
- Add a regression test `#sweep-resync: failed live-head file refresh drops stale cached files before review` that seeds an old-head file row, simulates live head drift and a failing `/pulls/:n/files` response, and asserts the stored head advances while cached files are cleared.

### Testing
- Ran targeted unit tests in `test/unit/queue.test.ts -t "sweep-resync"` and the new regression test passed. 
- Ran the same file with the specific `-t "failed live-head"` case and it passed (Vitest). 
- Typecheck (`npm run typecheck`) completed with no errors. 
- `npm audit --audit-level=moderate` could not complete in this environment due to the registry returning `403 Forbidden` (external audit endpoint), not a code failure.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3fb6fdad4c832cbcc9433a614641fe)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -28,6 +28,7 @@ import {
   listOpenPullRequests,
   listPullRequests,
   listPullRequestFiles,
+  deletePullRequestFiles,
   listRecentMergedPullRequests,
   updatePullRequestSlopAssessment,
   listRepoLabels,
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing: `codecov/patch` is red and the fix is incomplete. Dropping stale files on a resync failure leaves the review running against ZERO files (worse than stale), with no retry/defer and no observability for that failure. The underlying TOCTOU (stale `pull_request_files` after a failed head-resync refresh) is real — a reworked fix should defer the re-review to the next sweep (or retry the file fetch) instead of reviewing an empty set, and cover both branches to the 97% bar.

PR #1776, opened 2026-06-29
fix(signals): rank issue-quality reports before capping to the top 100

### LINKED ISSUE #1773: [Bug]: Issue-quality report caps to 100 issues before scoring and ranking
### Summary

`buildIssueQualityReport` (in `src/signals/engine.ts`) caps the open-issue set with `.slice(0, 100)` **before** it scores each issue and sorts the results by score. The cap is applied to the input order (callers pass open issues ordered by `updatedAt DESC`), so on a repo with more than 100 open issues only the 100 most-recently-updated issues are ever scored. A high-quality but less-recently-touched issue can never appear in the report, and the `ready` count used by downstream guidance (decision pack / reward-risk) is understated.

### Area

Signal logic

### Expected behavior

All open issues should be scored first, ranked by quality score (then issue number), and only then capped to the top 100. The report should surface the highest-quality issues regardless of how recently they were updated, matching the rank-before-cap behavior already used elsewhere in the signal pipeline.

### Actual behavior

The pipeline is `filter(open) -> slice(0, 100) -> map(score) -> sort(by score)`. The `.slice(0, 100)` truncates the issue list in `updatedAt DESC` order before any scoring, so the top-ranked issues that fall outside the 100 most-recently-updated are discarded. The final sort only reorders the arbitrary first 100.

### Reproduction

- A registered repo with ~150 open issues, ordered by `updatedAt DESC` as the caller provides them.
- A genuinely strong issue (rich body, no warnings, no linked/duplicate work) sits at position ~150 because it has not been touched recently.
- Request the issue-quality report for that repo.
- Expected: the strong issue appears in `issues` with status `ready`.
- Actual: it is dropped by the pre-score `.slice(0, 100)`; `readyCount` and the report omit it.

### Validation

Reproduced by reasoning through `buildIssueQualityReport` and the caller ordering. A regression test with more than 100 open issues (a high-quality issue placed after the first 100) confirms the issue is excluded today and included after reordering `slice` to run after the score+sort.

### Public-safety check

jason020818:
Closes #1773.

## The bug

`buildIssueQualityReport` (src/signals/engine.ts) caps the open-issue set with `.slice(0, 100)` **before** it scores each issue and sorts by score. Callers pass open issues ordered `updatedAt DESC`, so on a repo with more than 100 open issues only the 100 most-recently-updated issues are ever scored. A high-quality but less-recently-touched issue can never appear in the report, and the `ready` count used by downstream guidance (decision pack / reward-risk) is understated.

## The fix

Score every open issue first, rank by quality score (then issue number) — the order the report's own `.sort()` already implies — and only **then** cap to the top 100. The per-issue indexes were already built once over the full set, so scoring all open issues is O(issues + PRs), not a regression.

## Tests

Adds a regression test: 100 thin (low-score) issues followed by one detailed (high-score) issue placed **last** in `updatedAt` order (so the old cap-before-score dropped it). After the fix the high-quality issue is surfaced — it appears in the capped report and ranks at the top. Existing issue-quality / signals tests stay green; full local gate green, codecov patch covers the changed lines, OpenAPI unchanged.


JSONbored (commented):
Review notes blockers, closing.

PR #1794, opened 2026-06-29
feat(enrichment): doc-comment-vs-signature drift analyzer

### LINKED ISSUE #1519: feat(enrichment): Doc-comment-vs-signature drift
## Context
A REES (review-enrichment service) analyzer. Tier: **solid**.

**Detects:** The PR changed a function's signature (params/return/throws) but the adjacent JSDoc/docstring still describes the old shape, or references a parameter that no longer exists — a verifiable mismatch.

**Data source:** Fetch the full changed file at headSha via contents API (the hunk often splits doc-block from signature), parse doc + signature with a light TS/Python parser, diff the sets. Pure compute after one authed fetch.

This is heavy/external/historical analysis the no-checkout headless `claude --print` reviewer cannot do; the REES returns it as a brief block the engine splices into the review (additive + fail-safe).

## Implementation (established pattern, all inside `review-enrichment/`)
1. Finding type + `BriefFindings` key in `src/types.ts`
2. `src/analyzers/<name>.ts` — pure, inject `fetch` for tests
3. Register in `src/brief.ts` ANALYZERS registry
4. Render a public-safe block in `src/render.ts`
5. `node:test` units against `dist/` + a live smoke against the real data source

## Deliverables
- The analyzer + wiring + tests + a verifiable brief block (file:line or package@version)
- Clean PR off main (zero engine conflict; outside the engine tsc/vitest/codecov scope)

Parent: #1499

nickmopen:
## Summary

Adds a REES analyzer (#1519) that flags a **verifiable** doc-vs-code mismatch the PR introduced: a JSDoc/TSDoc `@param` tag naming a parameter the adjacent function **no longer declares** (the signature changed but the doc was left stale). The hunk often splits the doc block from the signature, so it fetches the full changed file at `headSha` (one authed contents fetch) and parses it. Additive + fail-safe — fills its own `docCommentDrift` findings key.

**Deliberately conservative for precision (no fuzzy guessing):**
- Only **named `function` declarations** whose parameters are **confidently enumerable** — any destructuring (`{ }`/`[ ]`) or arrow-default param makes the set ambiguous, so the whole function is skipped.
- Only functions whose **signature the PR actually changed** (a signature line is in the patch's added set) — pre-existing drift elsewhere is not reported.
- Only **top-level `@param` names**; nested `@param opts.x` tags reference an existing param and are ignored.
- Non-source and test files are skipped; a missing token/`headSha` or any fetch error fails safe (no finding).

## Implementation (established `review-enrichment/` pattern)
1. `DocCommentDriftFinding` type + `docCommentDrift?` key in `src/types.ts`
2. `src/analyzers/doc-comment-drift.ts` — pure helpers (`addedLineNumbers`, `parseDocParams`, `parseFunctionParams`, `findDocCommentDrift`) + `scanDocCommentDrift(req, fetch, opts)`; reuses the codeowners-style `contents` fetch
3. Registered in the `src/brief.ts` ANALYZERS registry
4. Public-safe block in `src/render.ts` (file:line + symbol + stale param names only)
5. `node:test` units in a **separate `test/doc-comment-drift.test.ts`**

## Validation
From `review-enrichment/` (Node 24):

```sh
npm test   # build + node --test: 235 pass / 0 fail (16 new doc-comment-drift units)
```

Covers added-line parsing, doc/param extraction (typed/optional/nested/rest/`this`/generics-with-comma), destructuring→skip, multi-line signatures, untouched-function exclusion, fail-safe fetch (missing token/headSha, non-ok, throw, abort), non-source skip, and the rendered block.

Closes #1519

### DIFF
--- a/review-enrichment/src/analyzers/doc-comment-drift.ts
+++ b/review-enrichment/src/analyzers/doc-comment-drift.ts
@@ -1,0 +1,312 @@
+// Doc-comment-vs-signature drift analyzer (#1519). Flags a JSDoc/TSDoc `@param` tag that names a parameter the PR
+// REMOVED or RENAMED while leaving the doc stale. It fetches the full changed file at headSha (one authed contents
+// fetch), reverse-applies the patch to reconstruct the pre-PR file, and compares each function's OLD parameter set
+// against its NEW one: a `@param` is drift only when it was a real parameter before and is gone now. This makes a
+// non-parameter signature edit (return type, name, modifier, parameter type) over PRE-EXISTING stale docs a
+// non-finding. Deliberately conservative: only NAMED `function` declarations whose parameters are confidently
+// enumerable (any destructuring / non-identifier param → skip the function). Reports symbol + stale params + line.
+import type { EnrichRequest, DocCommentDriftFinding } from "../types.js";
+
+const MAX_FILES = 20;
+const MAX_FINDINGS = 50;
+const MAX_SIGNATURE_LINES = 40;
+const SOURCE_RE = /\.(?:ts|tsx|js|jsx|mjs|cjs)$/;
+const SKIP_RE = /(?:\.d\.ts$|\.min\.|\.test\.|\.spec\.|__tests__\/|(?:^|\/)tests?\/)/;
+const SLUG_RE = /^[A-Za-z0-9._-]+$/;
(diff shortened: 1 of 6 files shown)


JSONbored (approved):
The most important remaining risk is a real false negative around optional JSDoc parameter syntax with defaults, where stale docs can be silently missed.

Please create a followup PR to resolve this as well. Approving.

PR #1698, opened 2026-06-28
fix(upstream): fail-closed when ruleset generatedAt is unparseable

kiannidev:
## Summary
- Add `isUpstreamRulesetStale()` helper that treats unparseable `generatedAt` values as stale (fail-closed)
- Wire `loadUpstreamStatus` through the helper instead of raw `Date.parse` comparison
- Regression tests for garbage timestamps, empty strings, and valid recent timestamps

Fixes #1695

## Validation
```sh
npx vitest run test/unit/upstream-ruleset.test.ts -t "1695|stale and unavailable"
```

## Test plan
- [x] Malformed `generatedAt` resolves to `status: "stale"`
- [x] Valid recent timestamp resolves to `status: "current"`
- [x] Existing stale-window behavior unchanged

PR #1848, opened 2026-06-30
fix(upstream): fail-closed when ruleset generatedAt is unparseable

### LINKED ISSUE #1695: fix(upstream): fail-closed when ruleset generatedAt is unparseable
## Problem

`loadUpstreamStatus` treats a ruleset snapshot as **current** when `generatedAt` is missing or malformed. `Date.parse` returns `NaN`, the stale comparison is false, and operators see a fresh upstream status on corrupted snapshot data.

## Expected behavior

Unparseable or missing `generatedAt` should resolve to **stale** (fail-closed), matching the guard pattern already used in `src/github/backfill.ts` (`Number.isFinite(ageMs)`).

## Scope

- Add a shared staleness helper in `src/upstream/ruleset.ts`
- Update `loadUpstreamStatus` status resolution
- Regression tests in `test/unit/upstream-ruleset.test.ts` for garbage ISO, empty string, and valid timestamps

## Acceptance criteria

- [ ] Malformed `generatedAt` → `status: "stale"` (when no drift reports)
- [ ] Valid recent timestamp → `status: "current"`
- [ ] Valid old timestamp beyond `UPSTREAM_STALE_MS` → `status: "stale"`
- [ ] Full `npm run test:ci` green

kiannidev:
## Summary

- Add `isUpstreamRulesetStale()` helper that treats missing/unparseable `generatedAt` values as stale (fail-closed)
- Wire `loadUpstreamStatus` through the helper instead of raw `Date.parse` comparison
- Merge `main` and resolve conflicts with the inline fix that landed upstream — helper centralizes the logic
- Regression tests for garbage timestamps, empty strings, null, and valid recent timestamps

Fixes #1695

Supersedes #1698 (closed due to merge conflicts; conflicts resolved on this branch).

## Validation

```
npx vitest run test/unit/upstream-ruleset.test.ts -t "1695|stale and unavailable"
```

## Test plan

- [x] Malformed `generatedAt` resolves to `status: "stale"`
- [x] Valid recent timestamp resolves to `status: "current"`
- [x] Existing stale-window behavior unchanged
- [x] Branch merges cleanly with `main`


JSONbored (commented):
Still seems you're missing something: 

```
The remaining gap is coverage depth: empty/null cases are asserted only at the helper level, not through the status-loading path described by the PR.
```

Please, get this fixed - you should be adding invariants + regressions everywhere possible so we can extensively test everything. It helps us catch problems way earlier, so it's important for my projects.

PR #1862, opened 2026-06-30
feat(enrichment): forward PR body to REES enrichment request

kiannidev:
## Summary

Forward `pr.body` through `runAiReviewForAdvisory` → `buildReviewEnrichment` so REES `history` can parse `Fixes #N` from the PR description.

## Issue

> **Action:** Open upstream issue `feat(enrichment): forward PR body to REES enrichment request` and link `Fixes #…` before gate review.

## Validation

```
npx vitest run test/unit/enrichment-wire.test.ts test/unit/enrichment-wiring.test.ts -t "body|FLAG-ON"
```

## Test plan

- [x] Non-empty PR body appears in REES POST
- [x] Missing body omits the field
- [x] Processor wiring test covers advisory → enrichment path

PR #1863, opened 2026-06-30
feat(enrichment): forward linked issue envelope to REES

kiannidev:
## Summary

When a PR has `linkedIssues`, resolve the primary issue from the local cache and forward a compact `linkedIssue` envelope (`number`, `title`, `body`) to REES so the `history` analyzer can correlate PR work with issue context.

## Issue

> **Action:** Open upstream issue `feat(enrichment): forward linked issue envelope to REES` and link `Fixes #…` before gate review.

## Depends on

Stacks cleanly after #1862 (PR body passthrough) — rebase if both touch `processors.ts`.

## Validation

```
npx vitest run test/unit/enrichment-wire.test.ts test/unit/enrichment-wiring.test.ts
```

## Test plan

- [x] `resolveEnrichmentLinkedIssue` returns cached envelope or number-only fallback
- [x] `buildReviewEnrichment` includes `linkedIssue` in POST body
- [x] Processor wiring test seeds issue + asserts REES payload

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -335,6 +335,7 @@ import {
   buildReviewEnrichment,
   isEnrichmentEnabled,
   isReesGithubTokenForwardingEnabled,
+  resolveEnrichmentLinkedIssue,
 } from "../review/enrichment-wire";
 import { captureReviewFailure } from "../selfhost/sentry";
 import { evaluateWithSurfaceLane } from "../review/content-lane-wire";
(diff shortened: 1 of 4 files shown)


JSONbored (commented):
Please fix this and resubmit as a new PR: 

```
The most notable issue is that the processor now resolves the linked issue before checking whether enrichment is enabled, which weakens the stated flag-off behavior but does not break review output because failures are swallowed into the number-only or undefined path.
```

PR #1886, opened 2026-06-30
test(utils): cover json helper edge cases

kiannidev:
## Summary
- Add focused unit tests for `parseJson`, `repoParts`, `errorMessage`, and related helpers in `src/utils/json.ts`

## Test plan
- [x] `npx vitest run test/unit/json-utils.test.ts`


JSONbored (commented):
Fix this: 

```
The only notable gap is that the rewrite drops one optional-argument coverage point from the previous test.
```

PR #1898, opened 2026-06-30
feat(enrichment): unsafe dynamic execution security analyzer

GildardoDev:
## What

A new REES security analyzer, `unsafeExec`, that flags code a PR adds which builds a shell command or an eval string by interpolation or concatenation and passes it straight to a dangerous execution sink. This is the classic command-injection / code-injection shape that the no-checkout reviewer cannot reliably catch by eye.

## No linked issue

no issue because: this is net-new analyzer work in the established REES analyzer pattern, and there is no open issue for an unsafe-execution signal. The existing security analyzers cover ReDoS, secrets-in-logs, and IaC misconfiguration, but not dynamic command or code execution, so this fills a real gap rather than duplicating an issue.

## Data source

Pure compute over the PR diff. No network, no external detector, no token. For each added line it finds a dangerous sink call and inspects only its first argument.

## Behavior

Conservative by design, mirroring the redos analyzer, so the false-positive rate stays near zero. A sink is flagged only when its first argument is dynamically built, that is a backtick template carrying a `${...}` interpolation or a `+` string concatenation outside any string literal. Sinks covered: `exec` and `execSync` (shell strings), `eval` and `new Function` (code), and the `execFile` / `spawn` family only when the command name itself is built dynamically. A pure string literal such as `exec("ls -la")`, a literal-command spawn with an args array such as `spawn("git", ["status"])`, a bare identifier such as `exec(cmd)`, and a commented-out line are all never flagged. Additive and fail-safe: it is bounded to 25 findings and a 2000-character line cap, a malformed or unterminated argument skips the line, and it reports only the file, line, sink name, and kind, never the argument text. It follows the established pattern entirely within `review-enrichment` (finding type in `types.ts`, a pure analyzer in `analyzers/unsafe-exec.ts`, a descriptor in `analyzers/registry.ts`, a public-safe block in `render.ts`), with the generated analyzer metadata regenerated to match.

## Tests

node:test cases with realistic diff patches cover the flagged cases (execSync, exec, eval, and new Function with interpolation or concatenation, a member-call sink, and spawn only when the command name is dynamic), the precision cases (pure string literals, a literal-command spawn, bare identifiers, commented-out sinks, removed lines, and sink-like substrings, none flagged), correct line numbers via hunk headers, multi-file scanning, MAX_FINDINGS bounding, and that neither the finding nor the rendered brief ever leaks the argument value. Full review-enrichment suite green.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,17 +65,17 @@ GITTENSORY_REVIEW_ENRICHMENT=false
 # Current analyzer names:
 # dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol,redos
 # provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature,iacMisconfig,nativeBuild
-# history,docCommentDrift,duplication,churnHotspot
+# history,docCommentDrift,duplication,churnHotspot,unsafeExec
 #
 # Profile defaults:
 # fast: dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol
-# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild
+# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild,unsafeExec
 # balanced (default): dependency,lockfileDrift,secret,license,installScript,heavyDependency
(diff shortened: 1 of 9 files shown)


JSONbored (commented):
See review for blockers. Closing.

Issue #998, opened 2026-06-22
Observability: Sentry error & performance instrumentation (optional per-deployment DSN)


JSONbored:
Part of #975. Related: #982 (self-host dashboard / observability), #983 (convergence & migration).

## Context
Add **Sentry** (errors + performance + distributed tracing) as an **optional, per-deployment** observability layer for the converged gittensory app. It plugs straight into the per-deployment-config model: an optional `SENTRY_DSN` binding — the **hosted** service uses gittensory's DSN; **self-hosters** supply their own; **unset = Sentry is fully off** (zero-config). Uses `@sentry/cloudflare` (Workers-native); Sentry's OSS / sponsored plan covers the hosted side.

It complements the existing Cloudflare observability + audit/usage telemetry by adding error **grouping & alerting**, **distributed tracing** of the review pipeline (webhook → queue → dual-AI → gate → post), **release health**, and — critically — **per-installation error/perf slicing**, which closes the "no per-tenant observability" gap.

## Deliverables (tracked as sub-issues)
- [ ] SDK integration + optional DSN seam (off by default)
- [ ] Privacy scrubber (`beforeSend`) — non-negotiable redaction
- [ ] Review-pipeline tracing + per-installation tagging
- [ ] Self-host + hosted wiring + docs

## Acceptance criteria
- With no `SENTRY_DSN`, Sentry is fully inert — zero behaviour change and zero added latency.
- With a DSN, errors + traces flow, grouped and sliceable per installation.
- **No** PR content, rubric, guardrail paths, tokens, BYOK keys, or wallet/reward/scoring context ever leaves in a Sentry event (verified by tests).
- Self-hosters enable their own Sentry with one env var; docs cover it.

## Notes
- Add the thin seam **during convergence** so it instruments the CONVERGED engine, but keep it dormant/optional until the converged stack is stable. Not a convergence blocker — a quality layer on top.
- Privacy scrubbing reuses the existing redaction discipline (the public-comment redactors / `sanitizePublicComment`) so it stays in sync with the comment surface.

## Current hardening child issues
- #1732 — REES degraded analyzer Sentry capture
- #1733 — Sentry cron monitors for scheduled self-host work
- #1734 — sampled Sentry spans for review latency
- #1735 — REES Sentry tags and fingerprints
- #1736 — post-deploy Sentry release/source-map validation
- #1737 — browser UI Sentry with private source maps
- #1738 — Cloudflare Worker Sentry release/source maps
[...]

PR #1904, opened 2026-06-30
feat(selfhost): sampled Sentry spans for review latency

### LINKED ISSUE #1734: observability(tracing): add sampled Sentry spans for review latency
Part of #998. Related: #1001 and #1207.

## Context
We need to answer operational questions like "why did this review take so long?" and "which stage is slow or failing?" without digging through scattered logs. The current Sentry setup is strong for errors; the next layer is sampled tracing across the review pipeline.

This issue is specifically for Sentry spans/traces. It should not replace the existing OpenTelemetry issue; it should either coexist cleanly or reuse the same span boundaries where practical.

## Requirements
- Keep tracing opt-in by sampling config. Default sample rate stays `0` unless explicitly configured.
- Add spans around the critical review path: webhook ingest, queue claim/process, GitHub API reads/writes, REES call, AI provider attempts, gate planning, comment/check-run publishing, and DB/cache work where it materially affects latency.
- Tag spans with safe dimensions: repo, PR number, job type, provider/model/effort, analyzer count, outcome, release, and environment.
- Never attach prompts, diffs, review bodies, tokens, headers, or private configuration blobs to spans.
- Keep trace cardinality controlled; avoid tags that contain full IDs or unbounded text unless hashed/truncated deliberately.

## Deliverables
- Shared span helper/wrapper so instrumentation is consistent and easy to audit.
- Instrumentation for self-host and REES call paths where they intersect review latency.
- Tests or focused smoke coverage proving tracing is no-op when sampling is off.
- Documentation for enabling a low sample rate and using traces to debug slow reviews.

## Acceptance criteria
- With sampling enabled, one review produces a connected trace with the major stages visible.
- With sampling disabled, behavior stays unchanged and no Sentry trace traffic is emitted.
- Slow or failed stages are filterable without exposing sensitive payloads.

nickmopen:
## Summary

Adds sampled **Sentry tracing** across the review pipeline so operators can answer "why was this review slow?" / "which stage failed?" without digging through scattered logs (JSONbored/gittensory#1734). A shared span helper keeps instrumentation consistent and easy to audit; it coexists with the existing OpenTelemetry spans by reusing the **same** boundaries.

## Opt-in & safe by construction

* **No-op until configured.** Spans are emitted only when `SENTRY_TRACES_SAMPLE_RATE > 0` (default `0`) — independent of error capture. `resolveTracesSampleRate` clamps to `[0,1]` and treats a malformed value as off, so a typo can't flood the tracer. With sampling off, no span is started and no trace traffic is emitted.
* **Never sensitive payloads.** `sentrySpanAttributes` keeps only the safe, low-cardinality subset (drops secret-keyed + null, keeps finite numbers/booleans, truncates strings) — never prompts, diffs, review bodies, tokens, headers, or config blobs.

## What it does

* `withSentrySpan(name, attributes, fn)` — runs `fn` inside a Sentry span (auto-closed, marked errored if `fn` throws) when sampling is on; a pure pass-through otherwise.
* `withReviewSpan(name, attributes, fn, options)` — the shared wrapper that opens **one** boundary feeding **both** tracers (OpenTelemetry + Sentry); each side independently no-ops when its backend is off.
* Wired at the existing trace boundaries for the latency-heavy stages: the **queue-job** span (whole-review latency — GitHub reads, REES call, AI, gate, publish all run inside it) and the nested **AI-provider** span. A sampled review therefore produces a connected trace with the major stages visible and slow/failed stages filterable.

## Acceptance criteria

* ✅ Sampling enabled → one review produces a connected trace (queue-job span with the AI-provider span nested).
* ✅ Sampling disabled → behavior unchanged, no Sentry trace traffic (the helpers reduce to `fn()`).
* ✅ Slow/failed stages filterable without exposing sensitive payloads (safe attributes + span error status).

## Validation

Node 24, from the repo root:

```sh
npm run typecheck        # clean
npx vitest run test/unit/selfhost-sentry.test.ts   # 47 pass
```

Unit tests cover: rate resolution (default/clamp/non-finite), attribute scrubbing (secrets/null/NaN/non-scalar dropped, truncation), no-op when off (DSN set, rate 0 → no span), span emission + safe attributes when on, `withReviewSpan` driving both tracers, and error propagation. The instrumented call-site lines (`selfhost.queue.job`, `selfhost.ai.provider`) are covered by the existing queue/AI suites. Documented in `.env.example`.

Part of JSONbored/gittensory#998. Closes JSONbored/gittensory#1734

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -230,7 +230,13 @@ REDIS_URL=redis://redis:6379                # REQUIRED for the self-host review 
 # --- Sentry error tracking (optional) ---
 # SENTRY_DSN=                                # enables self-host Sentry capture; unset = complete no-op
 # SENTRY_ENVIRONMENT=production
-# SENTRY_TRACES_SAMPLE_RATE=0                # traces are off by default; errors still report
+# SENTRY_TRACES_SAMPLE_RATE=0                # traces/spans are off by default; errors still report. Set a LOW rate
+#                                            #   (e.g. 0.05) with SENTRY_DSN to sample review tracing: each sampled
+#                                            #   review emits a connected trace — the queue-job span (whole-review
+#                                            #   latency) with the AI-provider span nested — so you can filter slow
+#                                            #   or failed STAGES in Sentry without reading scattered logs. Spans
+#                                            #   carry only safe dimensions (repo, job type, provider/model); never
+#                                            #   prompts, diffs, tokens, or bodies. Leave 0 to keep tracing a no-op.
 # SENTRY_RELEASE=                            # custom images only: set this ONLY when you uploaded source maps for
 #                                            #   the exact built bundle under this exact release id. Future official
 #                                            #   images bake GITTENSORY_VERSION=gittensory-selfhost@<version>, so do
--- a/src/selfhost/ai.ts
+++ b/src/selfhost/ai.ts
@@ -10,7 +10,7 @@ import type { CombineStrategy, OnMerge } from "../services/ai-review";
 import { isConfiguredSelfHostProvider, resolveConfiguredProviderNames } from "./ai-config";
 export { assertNoLegacySharedAiEnv } from "./ai-config";
 import { incr } from "./metrics";
-import { withOtelSpan } from "./otel";
+import { withReviewSpan } from "./sentry";
 import { delimiter } from "node:path";
 
 interface AiRunOptions {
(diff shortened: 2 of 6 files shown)


JSONbored (commented):
This is solid, however, we do need to fix this: 

```
The main maintainability gap is that the new shared wrapper lives in `sentry.ts`, which makes Sentry the owner of a cross-tracer abstraction and will encourage broader modules to import Sentry for non-Sentry concerns.
```

Closing, resubmit once that's fixed and it should be an easy merge. This is a high priority issue so it deserves a gittensor:priority label when merged.

PR #1915, opened 2026-06-30
feat(enrichment): flag unsafe DOM / code-execution sinks

jason020818:
## What

A new local REES analyzer, `unsafeDom`, that flags added JS/TS lines which pass data into a **DOM-HTML write sink** (`innerHTML`/`outerHTML` assignment, `insertAdjacentHTML`, `dangerouslySetInnerHTML`, `document.write`/`writeln`) or a **dynamic code-execution sink** (`eval`, `new Function`, a string-bodied `setTimeout`/`setInterval`) — the client-side XSS / arbitrary-code-execution surface a no-checkout reviewer cannot reliably catch by eye.

## Why it is a distinct dimension

Orthogonal to every existing analyzer: `redos` inspects regex backtracking *shapes*, `secretLog` inspects *logging* sinks, `secret`/`secretScan` inspect literal *values*, and `iacMisconfig` inspects *config-file paths* (it excludes `.ts`/`.js`). None model HTML-write or dynamic-eval sinks in source. The analyzer mirrors the proven `iac-misconfig`/`secret-log` skeleton (per-line patch scan, hunk-header line citation, `MAX_FINDINGS`/`MAX_LINE_CHARS` caps, `AbortSignal`).

## Precision — regex-literal aware

`codeOnly` strips string-literal, **regex-literal**, and `//` line-comment content in a single linear pass (no regex, so it can never backtrack) before matching. A sink named inside a string, a `/.../ ` regex, or a comment is therefore never flagged — e.g. `const re = /document\.write\(/;` is **not** a hit, and `setTimeout(fn, 1)` (function arg) and `el.innerHTML === x` (comparison, not assignment) are excluded. Regex-vs-division is disambiguated structurally, and `[...]` character classes are handled so an inner `/` does not end the literal. Every matcher is a flat, linear-time alternation, so the analyzer can never be the ReDoS class it sits beside.

## Output

Public-safe: location + sink + kind only, never code or values. Bounded at 25 findings/line-2000-chars, fail-safe on abort, `cost: "local"` (no network).

## No linked issue

Net-new analyzer, so there is no tracking issue to link.

## Tests

`review-enrichment/test/unsafe-dom.test.ts` (node:test) covers every branch: each sink kind, string/comment/regex-literal negatives, division-not-regex, character-class handling, multi-hunk line citation, removed/context-line skipping, long-line and budget caps, abort, the non-code-path skip, and both render arms. `npm test` is green and analyzer metadata is regenerated and committed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,17 +65,17 @@ GITTENSORY_REVIEW_ENRICHMENT=false
 # Current analyzer names:
 # dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol,redos
 # provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature,iacMisconfig,nativeBuild
-# history,docCommentDrift,duplication,churnHotspot
+# history,docCommentDrift,duplication,churnHotspot,unsafeDom
 #
 # Profile defaults:
 # fast: dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol
-# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild
+# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild,unsafeDom
 # balanced (default): dependency,lockfileDrift,secret,license,installScript,heavyDependency
 # actionPin,eol,redos,provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature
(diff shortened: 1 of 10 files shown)


JSONbored (commented):
See review for blockers, as well as failed CI.

PR #1928, opened 2026-06-30
fix(queue-trends): avoid Infinity review velocity when observedDays is zero

kiannidev:
## Summary
- Return `null` for `reviewVelocityPerDay` when snapshot spacing rounds to zero observed days.
- Render `n/a` in the trend summary instead of `Infinity/day` for malformed totals history.

## Test plan
- [x] `npx vitest run test/unit/queue-trends.test.ts`


kiannidev:
Addressed the latest Gittensory gate feedback:

- **Production fix restored** with a reachable path: when the penultimate and latest totals snapshots share `fetchedAt` (duplicate latest rows from malformed history), `reviewVelocityPerDay` is `null` and the summary renders `n/a/day` instead of dividing by `observedDays`.
- **Regression test** uses duplicate latest rows at the same timestamp and asserts `reviewVelocityPerDay === null` plus `review velocity n/a/day` in the summary.
- **Companion test** confirms the happy path still reports `1.43/day` when the latest snapshot is unique.

Please re-run the Gittensory review checkbox.

**No-issue rationale:** defensive queue-trends output hardening for duplicate totals snapshots; no linked issue.

kiannidev:
Updated per the latest Gittensory gate review:

- **Removed** the duplicate-latest `null` / `n/a` velocity suppression (it incorrectly dropped a valid metric when the 7d baseline was still usable).
- **Added** `sortTotals` deduplication: when multiple totals rows share the same `fetchedAt`, keep the newest row before trend math.
- **Regression test** now asserts duplicate latest rows collapse to `1.43/day`, `observedDays >= windowDays`, and never `Infinity`.

**No-issue rationale:** harden queue-trends against duplicate totals snapshot rows; no linked issue.

Please re-run the Gittensory review checkbox.

JSONbored (commented):
Fix review blockers: 

```
src/services/queue-trends.ts:91 constructs `baseline` with `Date.parse(snapshot.fetchedAt) <= targetMs`, so `observedDays` at src/services/queue-trends.ts:94 is always at least `windowDays` for ready windows and the duplicate-collapse change at src/services/queue-trends.ts:185 is a wrong-layer/no-op fix for the claimed Infinity case.
test/unit/queue-trends.test.ts:61 includes a real 7-day baseline snapshot, so the renamed regression test no longer exercises zero observed-day spacing or malformed totals history and cannot catch the PR's advertised failure mode.
```

PR #1929, opened 2026-06-30
feat(signals): treat __fixtures__ and __mocks__ paths as test evidence

kiannidev:
## Summary
- Count Jest/Vitest `__fixtures__` and `__mocks__` directories as test paths in `isTestPath`.
- Improves missing-test-evidence classification when only fixture/mock files accompany source changes.

## Test plan
- [x] `npx vitest run test/unit/test-evidence.test.ts`


JSONbored (commented):
Fix this, then resubmit: 

```
The only notable risk is classifier drift: `src/signals/engine.ts` still has a separate `isTestFile` regex set that does not include these directories, so you should consolidate it with `isTestPath`.
```

PR #1951, opened 2026-07-01
fix(github): attribute rate limits by token scope

### LINKED ISSUE #1940: fix(github): complete budget attribution for rate-limit admission
Parent: #1936

## Problem
Recent live behavior shows the queue can become intentionally parked by GitHub rate-limit admission. That is safer than retry storms, but beta-stable requires knowing exactly which actor, job type, and endpoint class is consuming budget.

## Requirements
- Every GitHub REST and GraphQL call path should have stable budget attribution where possible.
- Persisted observations should distinguish installation, brokered/global, repo fallback, and unknown/public buckets.
- Admission decisions should not be confused by legacy/null observations once keyed observations exist.
- Metrics and logs must make the root cause visible without SSHing into SQLite.

## Deliverables
- Audit all GitHub API call paths for admission key/rate observation coverage.
- Add missing labels/structured fields for job type, trigger kind, endpoint class, token scope, and installation/broker scope without high-cardinality labels.
- Add dashboard panels or alert text that explains parked queue states.
- Add tests for attribution invariants and legacy observation fallback behavior.

## Acceptance criteria
- A dashboard can answer: which queue job type is blocked, which budget bucket is low, and when it should resume.
- Unknown/global observations do not incorrectly mask or override newer exact-key observations.
- Repo-scoped jobs still honor relevant token-wide exhaustion.
- No new high-cardinality Prometheus labels are introduced.

## Expected outcome
Rate-limit incidents become diagnosable from metrics and logs, and admission behavior stays correct across migrations and multi-repo installations.

JSONbored:
## Summary

Closes #1940.

This fixes GitHub rate-limit attribution so self-host queue admission can distinguish installation-token work, public-token work, and unknown/no-token work instead of collapsing too much traffic into an ambiguous bucket.

## What changed

- Threaded installation ids through scheduled/API backfill, RAG indexing, and re-gate fan-out jobs.
- Passed installation/public admission keys into GitHub REST and GraphQL reads used by backfill, live PR freshness, live CI/merge facts, linked-issue fallback reads, and review file refreshes.
- Split rate-limit metric/admission scope labels into installation, public, global, unknown, and other.
- Added regression coverage for installed vs non-installed repo fan-out, public-token attribution, installation-token attribution, no-token fallback behavior, and queue/API payload propagation.

## Why

The self-host stack needs to keep unrelated GitHub work moving while respecting the actual depleted token bucket. Without stable attribution, observed rate-limit pressure can be too broad or too vague, which makes queue admission less useful during webhook and scheduled-work bursts.

## Validation

- `git diff --check`
- `npm run typecheck`
- `npx vitest run test/unit/github-client.test.ts test/unit/selfhost-queue-common.test.ts test/unit/selfhost-sqlite-queue.test.ts test/unit/selfhost-pg-queue.test.ts`
- `npx vitest run test/unit/backfill.test.ts test/unit/rag-index.test.ts test/unit/queue.test.ts test/unit/fetch-live-pr-state.test.ts test/unit/fetch-live-pull-request.test.ts test/integration/api.test.ts test/integration/routes-errors.test.ts`
- `npx vitest run test/integration/api.test.ts test/unit/backfill.test.ts test/unit/rag-index.test.ts test/unit/queue.test.ts`
- `npm run test:coverage`
- Local changed-line LCOV check: 369 changed source lines, 74 changed-line branches, 0 uncovered changed-line branches.


JSONbored:
Reworked and rebased onto `main` to resolve the flagged blocker and land cleanly on top of the recently-merged #1980 (GraphQL status rollup), #1985 (merged-PR file-hydration skip), and #1976 (conditional backfill).

**Blocker fixed — `linked-issue-hard-rules.ts` installation-token attribution.** The reviewer was right: the inline resolver only produced the public-token key, so an installation-token linked-issue fact read recorded as `unknown`. Rather than duplicate the token→key resolution in yet another spot (which is how it drifted), `resolveLinkedIssueHardRule` now takes the admission key as an **injected** arg — `maybeRunAgentMaintenance` already resolves the correct key (installation vs public) for exactly this `ciToken`, so it passes that in. Added a regression test asserting an installation key reaches `fetchLinkedIssueFacts` (a cross-module `vi.spyOn`).

**Scope-helper divergence reconciled.** `client.ts`'s `githubAdmissionKeyScope` now classifies identically to `queue-common`'s `githubRateLimitAdmissionKeyScope` (installation / public / global / unknown / other) and is exported + directly unit-tested (mirroring the queue-common test), so the two metric surfaces can't drift again.

**Threaded the admission key through the paths that didn't exist at this PR's base** (so attribution is complete, not partial):
- `#1985` `hydrateMergedPullRequestFiles` — now takes + forwards the key to the per-PR file reads.
- `#1980` `fetchLiveCiAggregateViaGraphQl` / `fetchLiveCiAggregatePreferGraphQl` — now accept + forward the key (the GraphQL rollup read is attributed too), and the check-suites backstop fetch inside `fetchLiveCiAggregate` gets it.

Conflicts resolved by keeping main's structure (GraphQL-preferring CI path, `hydrateMergedPullRequestFiles`) and threading this PR's admission key through it. Local gate green (`npm run test:ci` + `npm audit`).

JSONbored:
Addressed the API-drift note (duplicated token→key derivation + a helper that took a token without requiring the matching key):

- **One resolver.** Collapsed the three copies (`backfill.repoAdmissionKeyForToken`, `processors.githubAdmissionKeyForToken`, `pr-freshness`'s inline ternary) into a single exported `githubRateLimitAdmissionKeyForToken(env, token, installationId)` in `client.ts`. The two call-shaped wrappers now just delegate to it, so there's one place the public/installation/unattributed decision lives. Added a direct unit test covering every bucket (no-token, public, installation, non-public-without-id, non-finite-id).
- **Token can't travel without its key.** `resolveLinkedIssueHardRule` no longer takes a separable `admissionKey`; it takes `installationId` and **derives** the key from the same token it will read with, via that one resolver. So the key is always in sync with the token — there's no signature that lets a caller pass a token and omit/mismatch the key. Regression test updated to assert the derived installation key reaches the fact reads.

Gate green again locally (`npm run test:ci` + `npm audit`).

Issue #1936, opened 2026-07-01
chore(selfhost): beta-stable release readiness roadmap


JSONbored:
## Goal
Prepare Gittensory self-host for a beta-stable release candidate that other teams can safely run, without sharing our GitHub token budget or creating recurring rate-limit stalls.

## Current state
The current self-host stack is functional and recently redeployed from `main`, with Redis, Qdrant, queue admission, scheduler backpressure, webhook coalescing, cache metrics, and observability improvements in place. The system now avoids blindly retrying into exhausted GitHub buckets, but rate-limit pressure can still park jobs for minutes at a time.

That means the next milestone is not simply more retry logic. The beta-stable milestone must reduce unnecessary GitHub work at the source, make token ownership explicit for external self-hosters, publish an image-first release path, and document exactly how operators should run, observe, and troubleshoot the stack.

## Release target
Ship a beta/RC self-host image suitable for early external operators.

Recommended channel:
- `orb-v0.1.0-rc.1` or equivalent prerelease image tag
- GitHub Release marked prerelease
- no stable claim until rate-limit behavior, docs, and onboarding have survived real external usage

## Non-negotiables
- External self-hosters must use their own GitHub App installation tokens by default.
- Orb telemetry must remain available without routing all GitHub API calls through our central broker.
- Brokered Orb token mode must be explicitly managed/private-beta only.
- The default compose/env path must be image-first and safe for first-time operators.
- The queue must keep moving without wasteful polling or duplicate scheduled work.
- Rate-limit metrics must identify whether pressure comes from webhooks, background regates, status reads, brokered calls, or operator misconfiguration.
- Documentation must match the shipped image and defaults.

## Child issues
- [x] #1937
- [x] #1938
- [x] #1939
- [x] #1940
- [x] #1941
- [x] #1942
- [x] #1943
- [x] #1944
- [x] #1948 (net-new from rate-limit audit: temporal load-smoothing)
- [x] #1949 (net-new: coalesce scheduled RAG re-index over webhook incrementals)
- [x] #1950 (net-new: backoff/cap on retry-orb-relay fleet drains)
- [x] #2527 (net-new: scope PR-file hydration to the current review-critical set, cache by head SHA)
- [ ] #2533 (net-new from second-pass reliability audit: alerting reaches an operator by default + dead-letter blind spot)
- [ ] #2534 (net-new: automatic dead-letter job retry for self-host queues)
- [ ] #2535 (net-new: automate Docker disk hygiene)
[...]

PR #1981, opened 2026-07-01
fix(github): cache the App JWT so App-level reads cache-hit (#1940)

### LINKED ISSUE #1940: fix(github): complete budget attribution for rate-limit admission

JSONbored:
## Summary

`createAppJwt` minted a fresh App JWT (RS256 sign) on **every** call. Beyond the wasted CPU, the rotating JWT changes the auth-scoped **response-cache key** on every call (`responseCacheKey` hashes the `authorization` header), so the `metadata` cache class **never hits** for App-level reads (`GET /app/installations/{id}`) whenever two reads land more than a second apart — the exact heavy path used by the per-repo backfill / `refresh-installation-health`.

This caches the minted JWT for a margin of its ~9-minute validity (`APP_JWT_REUSE_MS = 8 min`), so repeated App-JWT reads share **one signature** and **one stable cache key** — the metadata cache class becomes effective, and the RS256 mint runs once per window instead of per call. Keyed by App id so a different App (or a test that swaps the private key) never reuses another's token, and cleared in `clearInstallationTokenCacheForTest`.

Advances #1940 (rec #8 from the #1936 audit).

## Scope

- [x] Focused, backend-only change (one function + a test).
- [x] Follows `CONTRIBUTING.md`; no `site/`/`CNAME`.
- [x] Linked issue (advances #1940).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` — added a regression test that advances fake time between two `getAppInstallation` reads and asserts the second is served from the response cache (`fetches === 1`) instead of re-fetching (it fails without the JWT cache, since a rotated JWT would miss). The test also covers the re-mint-after-window and different-App branches; the changed region is 100% covered (statements + branches).
- [x] `npm run test:ci`

- No `ui:openapi` / `cf-typegen` / migration regen: single-function backend change, no API/schema/binding/DB change.

## Safety

- [x] No secrets/wallets/hotkeys/trust-scores/etc. exposed (the JWT is held only in process memory, exactly like the existing installation-token cache; never logged or persisted).
- [x] No public GitHub text change.
- [x] The response-cache key still hashes the auth material (`responseCacheKey` is unchanged); this only makes the App JWT itself stable within its validity window, so no cross-tenant cache widening.
- [x] No API/OpenAPI/MCP change. No UI change.

## Notes

- The reuse window (8 min) sits safely inside the JWT's 9-min validity (iat backdated 60s, exp +540s), so a reused token is always still valid at GitHub.

### DIFF
--- a/src/github/app.ts
+++ b/src/github/app.ts
@@ -290,6 +290,7 @@ export function isForeignAppInstallation(
 export function clearInstallationTokenCacheForTest(): void {
   installationTokenCache.clear();
   externalTokenStore = null;
+  appJwtCache.clear();
   clearGitHubResponseCacheForTest();
 }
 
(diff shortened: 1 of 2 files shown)


JSONbored:
Fixed the blocker — the reviewers were right, thanks. `createAppJwt`'s cache identity now includes the **private key** (not just the App id), so a same-App credential rotation invalidates the cache immediately and never serves a JWT signed by the now-revoked old key. Added a regression test that rotates the key mid-window and asserts an immediate re-mint. Rebased on current `main` and re-ran the full gate (green, patch coverage 100%).

JSONbored:
Also folded in the perf nit: the JWT cache is now a **Map keyed by App id**, so a process that alternates between App identities keeps a per-App JWT instead of thrashing a single entry (added a test asserting alternating `appA`/`appB` reads don't evict each other). Gate green, patch coverage 100%.

PR #1992, opened 2026-07-01
fix(rees): fall back to direct OSV queries when lockfile batch fails

real-venus:
## Summary

The lockfile-drift analyzer's `queryOsvBatch` returned an empty result map on **any** `/v1/querybatch` failure:

    if (!response.ok) return results;

So a transient OSV batch-endpoint outage (503/500/timeout) silently reported **zero transitive CVEs for every changed package**. A vulnerable lockfile-only pin — precisely the transitive downgrade this analyzer exists to catch (it sees resolutions the manifest diff never names) — would escape the scan entirely. That's a fail-closed blind spot in a security check: an outage is indistinguishable from "clean."

The manifest `dependency-scan` analyzer already handles this correctly — on a batch failure it falls back to per-package direct queries. This aligns lockfile-drift with that proven pattern: when the batch request fails, fall back to per-package direct `/v1/query` calls and merge the partial results. The fallback honors the abort signal and the per-category call cap, and the happy path (batch succeeds) is unchanged.

Regression tests cover:
- batch fails → direct fallback succeeds → the vulnerable pin is still surfaced (and a clean package is queried, not silently dropped);
- batch and direct both fail → empty CVEs, no crash;
- an aborted signal stops the fallback loop.

## Scope

- [x] No linked issue — a small, self-evident robustness fix in one analyzer; the rationale is in the Summary.

## Validation

- [x] `git diff --check`
- [x] `npm --prefix review-enrichment test` (build → validate sourcemaps → analyzer-metadata check → full REES suite) — green; the three new `queryOsvBatch` fallback regression tests pass.
- [x] Rebased on `main` — no conflict.
- [ ] Root `npm run test:ci` jobs — not exercised by this change; it is confined to `review-enrichment/**`, which is outside the root `src/**` Codecov patch scope. Verified by CI.

- The change is entirely within `review-enrichment/src/analyzers/lockfile-drift.ts` and its test; the relevant local gate is the REES suite (`rees:test`), which is green. The root `src/**` coverage jobs do not cover this package.

## Safety

- [x] No auth/cookie/CORS/GitHub App/Cloudflare/session changes.
- [x] No API/OpenAPI/MCP behavior change.
- [x] No UI changes.
- [x] No docs/changelog changes.

### DIFF
--- a/review-enrichment/src/analyzers/lockfile-drift.ts
+++ b/review-enrichment/src/analyzers/lockfile-drift.ts
@@ -363,7 +363,49 @@ export function extractLockfileChanges(
   return changes;
 }
 
-/** Batch-query OSV.dev for lockfile resolutions. Best-effort: returns empty CVE arrays on any failure. */
+/**
+ * Direct single-package OSV query — the per-package fallback for a failed batch. Without it a transient
+ * /v1/querybatch outage silently yields zero transitive CVEs (#1502): a fail-closed blind spot where a
+ * vulnerable lockfile-only pin escapes the scan entirely. Mirrors the graceful batch→direct degradation the
+ * manifest dependency-scan analyzer already performs.
+ */
+async function queryOsvSingle(
+  change: LockfileChange,
+  fetchImpl: typeof fetch,
+  signal: AbortSignal | undefined,
+  options: Pick<ScanOptions, "analysis" | "diagnostics">,
+): Promise<Cve[]> {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Please fix + resubmit: 

```
The most notable maintenance gap is that the fallback helper hardcodes the default direct-query cap instead of reflecting any caller-supplied scan limit.
```

Issue #1942, opened 2026-07-01
fix(queue): eliminate duplicate scheduled work and gate jitter


JSONbored:
Parent: #1936

## Problem
Hourly or bursty queue spikes suggest scheduled work and webhook-triggered work can still overlap. Beta-stable needs strict no-waste scheduler behavior: no polling loops, no duplicate sweeps, and no review job starting before required CI state is meaningful.

## Requirements
- Scheduled regate/backfill work should not fan out while equivalent work is already pending or processing.
- Webhook bursts should coalesce where behavior is safe.
- Orb review/gate jobs should not repeatedly appear transiently running after already publishing a final state.
- Closed, stale, or head-changed PRs must not receive visible output from old work.

## Deliverables
- Audit scheduled jobs for hourly fanout patterns and overlapping triggers.
- Add explicit queue introspection/backpressure for remaining scheduled job types, not just regate sweeps.
- Add idempotency keys or terminal-state guards for public review/check publication where needed.
- Add tests for duplicate scheduled sweeps, closed PRs, head changes, and CI-not-ready transitions.

## Acceptance criteria
- Hourly scheduled events do not create avoidable queue spikes under normal load.
- A job already known to be stale exits before expensive GitHub reads.
- The Orb review check/comment does not flip back into running after reaching a terminal state unless there is a new head/event that legitimately requires it.
- The queue can continue draining useful work even when some jobs are deferred by GitHub budget.

## Expected outcome
The queue stays smooth under bursty GitHub activity and avoids creating work that cannot produce a valid review outcome.

JSONbored:
## Analysis input — temporal smoothing is a distinct gap from dedup

This issue targets *duplicate* scheduled work and terminal-state/gate jitter — both real. Two adjacent findings from the rate-limit audit that are worth scoping here or in a sibling:

**1. The hourly burst is a *temporal* problem, not (only) a duplication one.** Even with perfect dedup, `enqueueScheduledJobs` (`src/index.ts:91-190`) enqueues ~10 job types — several of them per-repo fan-outs — at the *same* `minute === 0` instant. The per-repo fan-outs stagger internally, but:

[...]

PR #1998, opened 2026-07-01
perf(github): dedup + budget-gate the upstream ref→SHA resolve

JSONbored:
## Summary

The upstream ref→SHA pin (`GET /repos/{repo}/commits/{ref}`) was fetched **twice per hour** — once in `refreshScoringModelSnapshot` (`refresh-scoring-model`) and once in the upstream-drift refresh (`refresh-upstream-drift`) — via two near-identical private resolvers, both firing in the same hourly cron window as the heavy maintenance fan-out.

This collapses them into one shared `resolveUpstreamCommitSha` and addresses both halves of the audit item (rec #7):

- **Dedup.** A bare `/commits/{ref}` read (no `/check-runs`, `/status`, `/pulls`, … suffix) is now a short-TTL `commit` GitHub-response-cache class. Both jobs issue the identical read through `timeoutFetch`, so the second one within the window is served from the cache (self-host Redis) — **one read per window instead of two**. Only these two upstream resolves use the bare-`/commits/{ref}` shape, so nothing else is affected (the mutable CI subresources under `/commits/{sha}` stay bypassed, as before).
- **Budget-gate.** The resolve yields (`null` → the caller falls back to the mutable ref, its existing degraded path) when the shared REST budget is at/below the low-water floor — so this best-effort audit pin never spends a scarce request during a rate-limit crunch.

TTL is configurable (`GITHUB_COMMIT_CACHE_TTL_SECONDS`, default 15m), documented in `.env.example` alongside the sibling cache TTLs.

Advances #1936 (self-host rate-limit RC; rec #7 of the #1942 audit). No `Closes` — #1942 is already closed and there's no dedicated open issue this fully resolves.

## Scope

- [x] Focused: one new shared module + a cache class; both call sites now delegate (removing two duplicated resolvers).
- [x] Behavior-preserving except the intended dedup/budget-gate; both callers already treat a null SHA as "use the mutable ref".

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` — new `resolveUpstreamCommitSha` tests (resolve / budget-gate-skips-the-request / **end-to-end dedup: 2 resolves → 1 fetch** / fail-open) + `commit` cache-class + TTL assertions; existing scoring-model + upstream-ruleset suites still green.
- [x] `npm run test:ci`

## Safety

- [x] No secrets/wallets/etc. Public-token, read-only.
- [x] The `commit` class caches only the bare ref-resolve at a short TTL; a `/commits/{sha}` SHA is immutable and a branch ref is at most TTL-stale — and the caller already tolerates the fully-mutable-ref fallback.
- [x] No public GitHub text / auth / UI change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -173,6 +173,7 @@ REDIS_URL=redis://redis:6379                # REQUIRED for the self-host review 
 # GITHUB_CACHE_TTL_SECONDS=20                # Short default Redis TTL for safe GitHub GET response caching. Set 0 to disable.
 # GITHUB_BRANCH_PROTECTION_CACHE_TTL_SECONDS=1200  # TTL for required-status branch protection reads.
 # GITHUB_METADATA_CACHE_TTL_SECONDS=600      # TTL for stable repo/user/installation metadata reads.
+# GITHUB_COMMIT_CACHE_TTL_SECONDS=900        # TTL for bare /commits/{ref} resolves; dedups the two hourly upstream ref→SHA reads.
 # QDRANT_URL=                                # set to http://qdrant:6333 to use Qdrant as the RAG vector store
 #                                            #   (--profile qdrant). Overrides the built-in sqlite-vec / pgvector.
 # DISCORD_WEBHOOK_URL=                        # one Discord channel for per-action notifications (merged/closed/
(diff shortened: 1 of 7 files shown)


JSONbored:
Fixed the review finding (budget-gate suppressing free cached resolves).

The gate is no longer an upfront check; it's passed as a `githubSkipNetworkWhen` callback that `timeoutFetch` consults **only when it is about to make a network read**. The order inside `timeoutFetch` is now: serve a cache **hit** → reuse an **in-flight** coalesced fetch → *then* (fresh network read only) consult the skip callback. So under REST pressure a cached ref→SHA is still returned for free; only a genuinely fresh read is skipped (→ synthetic `503` → the caller falls back to the mutable ref, as before). This holds on both the self-host cache path and the no-cache path.

Regression tests added: a cached resolve is served under budget pressure (two resolves, budget depleted on the second → still **one** fetch), and a cache-miss under pressure still skips the network (no request, null). Existing `github-client` + `scoring` + `upstream` suites green; full `test:ci` + `npm audit` clean.

Issue #1941, opened 2026-07-01
feat(github): collapse status reads with GraphQL rollups


JSONbored:
Parent: #1936

## Problem
The largest remaining avoidable GitHub cost is repeated status/check/branch-protection style reads. Caching and backpressure help, but beta-stable should reduce the number of calls created in the first place.

## Requirements
- Replace high-volume repeated REST reads with fewer equivalent reads where safe.
- Prefer GraphQL for aggregate PR/check/status snapshots when it can collapse many REST calls into one bounded query.
- Keep REST for writes, webhook-specific actions, and endpoints where GraphQL would add semantic risk.
- Preserve correctness for required checks, branch protection, mergeability, skipped checks, and gate decisions.

## Deliverables
- Identify the top GitHub endpoints by live metrics/logs.
- Build a targeted GraphQL/status-rollup path for PR status/check summary reads.
- Cache repo/base-branch policy snapshots for a bounded TTL where correctness allows.
- Add regression tests proving gate decisions match the REST path for pending, success, failure, skipped, missing, and required-check cases.

## Acceptance criteria
- The hot status/check path performs materially fewer GitHub requests per PR pass.
- GraphQL failures fail closed or fall back safely without creating request storms.
- Review/gate behavior remains semantically equivalent.
- Metrics show REST pressure decreases after deployment.

## Expected outcome
The system stops burning GitHub budget on repeated status reads and uses GitHub APIs in a way that scales better for active repos.

JSONbored:
## Implementation notes — GraphQL status rollup

### Exact REST call cluster this collapses (one gate/maintenance pass, one PR)

`fetchLiveCiAggregate` (`src/github/backfill.ts:2123-2285`) + siblings currently issue, per PR pass:

- `GET /commits/{sha}/check-runs?per_page=100&page=N` — paginated (`:2160`)
- `GET /commits/{sha}/status?per_page=100&page=N` — paginated combined status (`:2197`)
- `GET /commits/{sha}/check-suites?per_page=100` — the late-materialization backstop (`:2253`)
- `GET /branches/{base}/protection/required_status_checks` — required contexts (`:2100`; already `branch_protection`-cached, 20 min)
- `GET /pulls/{n}` — fetched up to **3×** for `mergeable_state` (`:2296`), `state` (`:2305`), `headSha` (`:2314`)

That's ~5–8 REST calls (plus pagination) against the **core** bucket, per PR, per sweep.

[...]

PR #2005, opened 2026-07-01
perf(queue): fan build-contributor-evidence out into per-batch jobs (config-driven)

JSONbored:
## Summary

The scheduled `build-contributor-evidence` job resolved evidence for **up to 500 contributors in ONE job** — each doing a `/users/{login}` read plus its (uncached, paginated) `/users/{login}/repos` reads — so the whole set bursts against GitHub in a single execution during the 6-hourly full-sync window.

This makes the scheduled trigger **fan out into per-batch jobs** so those per-login reads spread across the queue's paced execution + rate-limit admission instead of bursting:

- The cron trigger (no login) derives the contributor set from stored PRs + issues, then enqueues one `build-contributor-evidence` job per batch of `CONTRIBUTOR_EVIDENCE_BATCH_SIZE` logins (staggered), and returns.
- Each batch job loads the shared aggregate data once and processes its batch; the single-login re-index path is unchanged.
- Config-driven: `CONTRIBUTOR_EVIDENCE_BATCH_SIZE` (default **150**, `0` = disable → the pre-existing single job). A set that fits one batch stays a single job, so small instances see no change.

**Trade-off (called out honestly):** because each per-login computation needs the full aggregate (`buildContributorFit` / `buildContributorOutcomeHistory` cross-reference all PRs/issues), each batch re-loads that shared data. So fan-out trades a bounded `ceil(N/batchSize)×` DB re-read for spreading the GitHub burst — which is the right trade on a rate-limited instance, and the batch size lets operators tune it (or turn it off).

**Footgun avoided:** a fanned-out batch coalesces by its **first login** (batches are disjoint slices, so heads are unique) — distinct batches never collapse into one `:all` key (which would silently drop all-but-one batch). A duplicate re-enqueue of the same batch still coalesces.

Advances #1936 (rec #2 of the #1941 read-reduction audit).

## Validation

- [x] `git diff --check`, `npm run typecheck`
- [x] `npm run test:coverage` — new tests: the fan-out enqueues one batch job per login above the size; the disabled (`0`) path stays a single inline job; the batch-size reader defaults on unset/invalid/negative; and the batch coalesce key is unique per batch (never `:all`). Existing single-login + inline evidence tests still pass.
- [x] `npm run test:ci`, `npm audit --audit-level=moderate`
- No cf-typegen: `CONTRIBUTOR_EVIDENCE_BATCH_SIZE` is read via `process.env` (works on cloud + self-host), documented in `.env.example`.

## Safety

- [x] No secrets/etc.; output is byte-identical (same evidence for the same contributors), only the job topology changes.
- [x] Default 150 keeps small instances on the current single-job path; `0` fully disables the fan-out.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -199,6 +199,9 @@ REDIS_URL=redis://redis:6379                # REQUIRED for the self-host review 
 # --- Queue worker (#977/#1201) ---
 # QUEUE_CONCURRENCY=4                        # max concurrent job-processing loops per instance (default 4; set 1 for strict serial processing)
 # QUEUE_BACKGROUND_CONCURRENCY=1             # max low-priority/background jobs allowed to occupy QUEUE_CONCURRENCY slots
+# CONTRIBUTOR_EVIDENCE_BATCH_SIZE=150        # logins per build-contributor-evidence job; the scheduled run fans out into
+#                                            #   per-batch jobs above this so the per-login GitHub reads spread across the
+#                                            #   queue instead of bursting. Set 0 to disable the fan-out (single job).
 
 # --- Caddy HTTPS terminator (#1203; requires --profile caddy) ---
 # DOMAIN=gittensory.example.com             # fully-qualified domain; Caddy auto-obtains a Let's Encrypt cert
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -781,7 +781,7 @@ export async function processJob(env: Env, message: JobMessage): Promise<void> {
       await fileUpstreamDriftIssues(env);
       return;
     case "build-contributor-evidence":
-      await buildContributorEvidence(env, message.login);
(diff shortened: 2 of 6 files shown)


JSONbored:
Rebased onto main to resolve the \`queue-common.ts\` conflict from #2001 (the event-job coalescing that just merged): `jobCoalesceKey`'s destructure now carries both the event-job id fields and this PR's `logins`, and the two switch cases sit side by side.

Also tightened the coalescing contract the review flagged: a fanned-out batch can no longer fall through to the `:all` key (the *scheduled-trigger* slot) — a non-empty `logins` batch keys by its first login, and a batch with no usable head is left **uncoalesced (null)** rather than collapsed into `:all` (which would have dropped its work). Added a test for that edge. Gate + `npm audit` green.

PR #2010, opened 2026-07-01
feat(selfhost): add a backup restore/validation drill (verify-backup.sh)

### LINKED ISSUE #2011: Postgres reliability/scale hardening (post-cutover)
Tracking issue for Postgres reliability/scale hardening following the self-host Postgres cutover (Redis, Qdrant, backups, Grafana, Prometheus already healthy per that cutover). No prior upstream issue covered this; opened to track the sub-goals below since they span multiple PRs.

## Sub-goals

- [x] **Backup restore/validation drill** — `pg_restore --list` validation + a guarded opt-in scratch restore, documented. Shipped in #2010.
- [ ] **Postgres observability** — Prometheus/Grafana panels + alerts for connection count, lock waits, slow/problem queries, DB size/table growth, dead tuples/autovacuum pressure, and backup freshness. Shipped in #2015 
- [ ] **PgBouncer readiness guidance** — when to enable `--profile pgbouncer` (already shipped as a pooler, #1200/#977) + a safe connectivity smoke command. No default-on cutover.
- [ ] **Off-host backup planning** — docs/config placeholders first; if implemented, opt-in and off by default (S3/R2/B2 export never runs without explicit env vars).

Each sub-goal ships as its own narrow, tested PR.

JSONbored:
Part of #2011.

## Summary

The `backup` profile writes Postgres `pg_dump` / SQLite backups, but nothing verified they are actually restorable — a backup you can't restore is not a backup. This adds `verify-backup.sh`, an on-demand drill that checks the newest backup (or a specific file) **without touching the live database**, plus an opt-in scratch restore.

## What changed

- **`scripts/verify-backup.sh`** (new):
  - Postgres `.dump` → `pg_restore --list` must parse the archive and find a non-empty table of contents.
  - SQLite `.sqlite.gz` → `gzip -t` integrity + `PRAGMA integrity_check` on a temp copy.
  - Opt-in scratch restore (`VERIFY_RESTORE_SCRATCH=1` + `GITTENSORY_VERIFY_SCRATCH_DATABASE_URL`): restores the dump into a **throwaway** database and sanity-counts tables. It **refuses to run** when the scratch URL is empty or equals the live backup source, so a destructive `pg_restore --clean` cannot hit production by accident.
  - Accepts an explicit backup path, or auto-selects the newest for the active DB type.
- **`docker-compose.yml`** — mounts the script into the `backup` service (`sh /verify-backup.sh`) and passes the opt-in scratch env through.
- **Docs** — a "Verify a backup is restorable" section on the backup/scaling page, including what a healthy run looks like and the guarded scratch-restore recipe.

## Why

Goal #1 of the Postgres reliability hardening (#2011): give operators a safe, documented way to prove backups restore, prioritized as the smallest high-value slice.

## Validation

- `test/unit/selfhost-verify-backup-script.test.ts` (new, 10 cases; fake `pg_restore`/`psql`/`sqlite3`, real `gzip`): Postgres validation pass/fail, no-dump, the two scratch guards (missing URL / equals live), the scratch happy path, explicit-file mode, and the SQLite integrity/gzip failures.
- `npm run db:migrations:check`, `npx tsc --noEmit`, `npm run ui:lint`, `npm run ui:typecheck` — all clean.

## UI Evidence

Docs-only content addition to an existing page (`docs.self-hosting-backup-scaling.tsx`) — no new component, styling, or layout change; uses the existing `CodeBlock`/`Callout` primitives, and passes `ui:lint` + `ui:typecheck`. Rendered copy is the new "Verify a backup is restorable" section shown in the diff. Happy to attach a rendered screenshot if you'd like one.

## Notes

First slice of #2011, kept deliberately narrow. `pg_restore`/`psql` are already installed in the backup image (`postgresql16-client`); no image change needed.

### DIFF
--- a/apps/gittensory-ui/src/routes/docs.self-hosting-backup-scaling.tsx
+++ b/apps/gittensory-ui/src/routes/docs.self-hosting-backup-scaling.tsx
@@ -110,6 +110,41 @@ npm run selfhost:postgres:migrate -- --sqlite /data/gittensory.sqlite --postgres
         <li>Confirm recent review rows and job state are present.</li>
       </ul>
 
+      <h2>Verify a backup is restorable</h2>
+      <p>
+        The <code>backup</code> profile ships <code>verify-backup.sh</code>, which checks the newest
+        backup without touching the live database: Postgres <code>.dump</code> archives with{" "}
+        <code>pg_restore --list</code>, and SQLite <code>.sqlite.gz</code> backups with a gzip and{" "}
+        <code>integrity_check</code> pass. Run it against the newest backup, or a specific file:
+      </p>
+      <CodeBlock
+        lang="bash"
+        code={`docker compose --profile backup run --rm backup sh /verify-backup.sh
+docker compose --profile backup run --rm backup sh /verify-backup.sh /backups/postgres/gittensory-<timestamp>.dump`}
+      />
+      <p>
+        A healthy run ends with <code>[verify] postgres archive OK: … (N TOC entries)</code> (or{" "}
+        <code>[verify] sqlite backup OK</code>), then <code>[verify] complete</code>, and exits 0.
+        Corruption, a missing backup, or an empty archive exits non-zero with a{" "}
+        <code>[verify]</code> reason.
+      </p>
+      <p>
+        To prove a dump actually restores, opt into a scratch restore into a <em>throwaway</em>{" "}
+        database — never the live one:
+      </p>
+      <CodeBlock
+        lang="bash"
+        code={`docker compose --profile backup run --rm \\
+  -e VERIFY_RESTORE_SCRATCH=1 \\
+  -e GITTENSORY_VERIFY_SCRATCH_DATABASE_URL=postgres://user:pass@host:5432/gittensory_verify \\
+  backup sh /verify-backup.sh`}
+      />
+      <Callout variant="warn">
+        The scratch restore runs <code>pg_restore --clean</code> against{" "}
+        <code>GITTENSORY_VERIFY_SCRATCH_DATABASE_URL</code>, so point it at a dedicated database you
+        can afford to drop. The script refuses to run when that URL equals the live backup source.
+      </Callout>
+
       <p>
         After scaling, revisit <Link to="/docs/self-hosting-operations">Operations</Link> and{" "}
         <Link to="/docs/self-hosting-security">Security</Link> because network and credential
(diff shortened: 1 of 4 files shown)


JSONbored:
Fixed. Root cause was worse than a missing package: \`docker compose run --rm backup sh /verify-backup.sh\` replaces \`command:\`, not \`entrypoint:\`, and the old entrypoint (\`/bin/sh -c\`) folded the apk install into \`command:\` — so an override skipped the install AND, because \`-c\`'s extra args become the nested shell's \$0/\$1 rather than a script to run, it collapsed into a bare interactive \`sh\` that just hangs reading stdin. (Confirmed live with real \`docker compose run\` — it hung until killed. The pre-existing documented \`sh /backup.sh\` one-shot had the identical bug.)

Fix: the install now lives in the entrypoint, ending in \`exec "\$@"\`, with a placeholder \`sh\` entrypoint arg so the real payload (default loop or a \`run\` override) lands in \`"\$@"\` starting at \$1 — mirrors the inline entrypoint-script pattern this file already uses for grafana.

Verified against real Docker, not just reasoned about:
- \`docker compose --profile backup up -d backup\` → normal backup.sh cycle completes (\`[backup] complete …\`).
- \`docker compose run --rm backup sh /verify-backup.sh\` → now actually executes (previously hung).
- Seeded a dummy Postgres dump and confirmed \`pg_restore\` is present and invoked (\`pg_restore --list\` reports its own parse error on the fake dump, not "command not found").

New commit: 24d76af7.

JSONbored:
Confirmed valid, fixed. \`[ "\$scratch" = "\$PG_DB" ]\` compared connection strings byte-for-byte, so a differently-spelled equivalent (\`postgresql://\` vs \`postgres://\`, a host alias, explicit vs default port) would slip past the guard.

Replaced the string compare with an **identity check**: ask Postgres itself for \`current_database() || '@' || pg_control_system().system_identifier\` on both connections and compare that. \`system_identifier\` is a random 64-bit value fixed for the life of a cluster's data directory — independent of how the connection was dialed — and PUBLIC has EXECUTE on \`pg_control_system()\` by default (verified against a real non-superuser role, no privilege escalation needed).

I deliberately avoided a network-address-based fingerprint (e.g. \`inet_server_addr()\`): testing against a real Postgres 16 instance showed the SAME server can report different addresses across connections over different address families (IPv6 \`::1\` vs IPv4 \`127.0.0.1\` loopback) — a false "these differ" that would have reopened a gap in the wrong direction for a safety guard. Any failure to fingerprint either side now aborts (fails closed).

Validated two ways:
- Unit tests: a fake \`psql\` keyed by connection string simulates two differently-spelled URLs resolving to the SAME database (the exact bypass reported) plus fail-closed cases when either identity query fails.
- Real end-to-end run against Postgres 16 (matching the backup image): refuses a byte-identical URL, refuses a differently-spelled equivalent of the same database, and correctly allows + completes a restore into a genuinely different database on the same cluster — with the live database's table confirmed untouched throughout.

New commit: 7efaf83d.

PR #2122, opened 2026-07-01
feat(signals): add contributor intake score breakdown with weighted deductions and top lever

GildardoDev:
## What

A new pure service, `explainContributorIntake`, that decomposes the contributor-intake `score` carried on a computed `ContributorIntakeHealth` into the weighted deductions that lowered it from a perfect 100, and names the single highest-leverage lever a maintainer can pull to make the repo more attractive to contributors. It is a sibling of `queue-burden-breakdown.ts` and `score-breakdown.ts` in the same explanation-service family.

## No linked issue

no issue because: this is net-new analytical work that fills a real gap, and there is no open issue for it. The engine computes the intake `score` in `buildContributorIntakeHealth` but nothing explains how it breaks down, so a maintainer sees a single opaque number and its band with no view of which factor (queue burden, duplicate clusters, or config quality) is dragging intake down or which lever helps most. This adds that explanation without changing the score.

## Data source

Pure compute over an already-computed `ContributorIntakeHealth` object. No network, no GitHub fetch, no I/O, no database access. Every deduction is derived from fields already present on the object: `queueHealth.burdenScore`, `duplicateClusters`, and `configLevel`.

## Behavior

The intake score is a base of 100 minus three weighted deductions, mirroring `buildContributorIntakeHealth`: queue burden times 0.55, duplicate clusters times 8, and a fixed config-quality penalty by band (fragile 30, needs_attention 18, good 6, else 0). Each factor is reported with its observable driver, the points it removed, its share of the total deduction, a band (none, low, moderate, high), and a specific lever, plus the single highest-leverage lever. When nothing is deducted the lever is an explicit no-op rather than an arbitrary factor, and a `clamped` flag marks when the deductions exceeded 100 and the engine floored the score at 0. The weights mirror the engine, and a drift-guard test rebuilds a `ContributorIntakeHealth` through `buildContributorIntakeHealth` and asserts this module recomposes the exact same score, so an engine weight change fails the suite instead of silently producing a wrong breakdown. Public-safe by construction: it reports the same observable drivers the intake summary already prints (queue burden out of 100, duplicate-cluster count, config band) and routes every rendered string through `sanitizePublicComment`.

## Tests

Vitest cases cover a perfect-intake no-deduction case with an honest no-op lever, a dominant high-band drag with a small low-band factor, a moderate top drag with the largest-remaining lever, each config band mapping to its fixed penalty, the deductions-exceed-100 clamp, repo and timestamp passthrough, a public-safe check that no rendered string leaks reward or private terminology, and the engine-parity drift guard that recomposes the exact score from `buildContributorIntakeHealth`. Branch coverage on the new file is 100 percent.

## Validation

- [x] `npm run test:coverage` for the new file: 100 percent branch coverage (35/35).
- [x] `npm run typecheck`: no errors introduced by these files.
- [x] Purely additive: one new service module and one new test file, no changes to existing code, no new dependencies.

### DIFF
--- a/src/services/contributor-intake-breakdown.ts
+++ b/src/services/contributor-intake-breakdown.ts
@@ -1,0 +1,190 @@
+import { sanitizePublicComment } from "../github/commands";
+import type { ContributorIntakeHealth } from "../signals/engine";
+
+// ─── Contributor intake breakdown (explanation family) ───────────────────────────────────────────
+// A pure projection over a computed {@link ContributorIntakeHealth} that decomposes its otherwise-opaque
+// intake `score` into the weighted deductions that lowered it from a perfect 100, and names the single
+// highest-leverage lever a maintainer can pull to make the repo more attractive to contributors. Sibling of
+// `queue-burden-breakdown.ts` and `score-breakdown.ts`: deterministic, no I/O, no GitHub fetch. Public-safe by
+// construction — it reports the same observable drivers the intake summary already prints (queue burden out of
+// 100, duplicate-cluster count, config band) and routes every rendered string through `sanitizePublicComment`.
+
+export type IntakeDeductionBand = "none" | "low" | "moderate" | "high";
+
+export type IntakeDeduction = {
+  /** The intake factor this deduction comes from. */
+  component: "queueBurden" | "duplicateClusters" | "configQuality";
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Some issues: 

```
The main rough edges are polish and maintainability: the output exposes internal component ids in user-facing reason/summary strings, and the test file has a small unused type import.
```

PR #2368, opened 2026-07-01
fix(queue): serialize concurrent webhook/sweep passes for the same PR

### LINKED ISSUE #2129: fix(queue): concurrent webhook and sweep jobs for the same PR are not mutually excluded
Parent: #1936

## Problem

`jobCoalesceKey()` only dedups a *pending* job against another *pending* job with an identical key. A `check_run`/`check_suite completed` webhook keys on `github-webhook:ci-completed:{repo}@{headSha}#{prNumbers}`, a `pull_request synchronize` webhook keys on `github-webhook:pr-refresh:{repo}#{pr}@{headSha}`, and a sweep-driven re-gate keys on `agent-regate-pr:{repo}#{pr}` (`src/selfhost/queue-common.ts:727`, `src/github/webhook-coalesce.ts:19-34`). These are three different key shapes for the same PR, so they never coalesce with each other. Combined with the default `QUEUE_CONCURRENCY = 4` (explicitly sized to overlap I/O-bound jobs), two of these jobs for the same PR can be claimed and processed by two concurrent workers, each with its own freshly-fetched (and independently-timed) live CI/mergeable/reviewDecision snapshot.

The only per-PR-keyed coalescing window that exists (`ciReReviewCoalesced`, a 60s TTL) is wired into the `check_run`/`check_suite` webhook handler only — it is never consulted by the sweep's `agent-regate-pr` path, so sweep-vs-webhook races are entirely unguarded. This is a codebase-acknowledged gap: `src/env.d.ts` carries an explicit TODO for a per-PR `SubmissionLock` Durable Object, deliberately deferred as a separate, more-involved sub-task.

**Failure scenario:** a scheduled re-gate sweep enqueues `agent-regate-pr:{repo}#71` for a PR at the same moment GitHub redelivers a `check_run completed` webhook for that PR's current head (a different coalesce key). Both jobs are claimed concurrently. Both resync to the same live headSha and both pass their own freshness check. If the two live reads (fetched a few hundred ms apart) disagree on `reviewDecision`/`mergeable_state` — e.g. the bot's own prior `approve` review lands asynchronously between the two reads — one job could plan `close` while the other plans `merge`/`approve`, and both would pass their independent freshness checks and both execute.

## Requirements

- Two jobs targeting the same PR, regardless of coalesce-key shape, must not both reach live-read-and-actuate concurrently.
- The fix should not require every job type to share an identical coalesce key (their current key shapes serve other purposes).

## Deliverables

- Add a lightweight per-PR mutual-exclusion guard around the plan-and-execute critical section — e.g. a short-TTL advisory lock keyed `agent-maintenance-lock:{repo}#{pr}` (reusable via the existing `SELFHOST_TRANSIENT_CACHE`, the same mechanism `ciReReviewCoalesced` already uses), acquired at the start of `maybeRunAgentMaintenance` and released after `executeAgentMaintenanceActions` returns. A job that fails to acquire the lock skips this pass rather than proceeding (the next webhook/sweep tick is the backstop).
- Route **both** the sweep path (`regatePullRequest`) and the webhook paths through the same lock, closing the asymmetry where only the webhook path currently has any per-PR coalescing.
- Longer-term: implement the per-PR `SubmissionLock` Durable Object already noted as a TODO, for a durable primitive that works across the sqlite/pg self-host split.
- Add a regression test asserting two concurrently-claimed jobs for the same PR — one shaped as `agent-regate-pr`, one as a `github-webhook` (ci-completed or pr-refresh) — cannot both reach `executeAgentMaintenanceActions` for the same PR at the same time.

## Acceptance criteria

- Two jobs for the same PR arriving via different trigger shapes never run their plan-and-execute critical sections concurrently.
- A job that can't acquire the lock defers cleanly rather than erroring or silently dropping the PR from evaluation.

## Expected outcome

Actuation for a given PR is always serialized to one in-flight decision at a time, regardless of which trigger (webhook or sweep) initiated it — closing a real (if narrow) path to two independently-timed, possibly-disagreeing actuations racing each other.

JSONbored:
## What

`check_run`/`check_suite completed`, `pull_request synchronize`, and the scheduled `agent-regate-pr` sweep all key their job coalescing differently (`github-webhook:ci-completed:{repo}@{headSha}#{prNumbers}`, `github-webhook:pr-refresh:{repo}#{pr}@{headSha}`, and `agent-regate-pr:{repo}#{pr}` respectively), so they never dedup against each other. Combined with `QUEUE_CONCURRENCY` explicitly overlapping I/O-bound jobs, two of these can be claimed and processed concurrently for the same PR, each computing its own independently-timed live CI/mergeable/reviewDecision snapshot. If two such reads disagree — e.g. the bot's own prior `approve` review lands asynchronously between them — one pass could plan `close` while the other plans `merge`/`approve`, and both would pass their own freshness check and both execute.

The only existing per-PR-keyed coalescing (`ciReReviewCoalesced`, a 60s window) is wired into the `check_run`/`check_suite` handler only — it's never consulted by the sweep's `agent-regate-pr` path, so sweep-vs-webhook races are entirely unguarded.

## Fix

Add a short-TTL per-PR advisory lock (`agent-maintenance-lock:{repo}#{pr}`, via the existing `SELFHOST_TRANSIENT_CACHE` — the same mechanism `ciReReviewCoalesced` already uses) around the plan-and-execute critical section of `maybeRunAgentMaintenance` (`src/queue/processors.ts`). A pass that can't claim the lock defers cleanly — the next webhook/sweep tick is the backstop. Lightweight stand-in for the per-PR `SubmissionLock` Durable Object already noted as a longer-term TODO in `env.d.ts`.

The critical section is extracted into `runAgentMaintenancePlanAndExecute` so the `try/finally` wrapping it doesn't force-reindent the ~200 existing lines inside (which would have made this diff far larger and harder to review for no logic change). Both the direct webhook path and the sweep path (via `reReviewStoredPullRequest`) converge on the same `maybeRunAgentMaintenance` call, so wrapping it there covers both without touching either caller.

`claimAgentMaintenanceLock`/`releaseAgentMaintenanceLock` are exported for direct testing. The TTL (60s, matching the existing `CI_COALESCE_WINDOW_SECONDS`) is a crash-safety backstop only — the normal path releases explicitly in the `finally` block within seconds.

## Tests

- `claimAgentMaintenanceLock` claims when free, denies when held (per-PR, not repo-wide), and a released lock is claimable again.
- Fails open on a broken transient cache (never itself blocks actuation — it's defense-in-depth, not the primary gate).
- Integration: a maintenance pass defers its entire plan-and-execute critical section (no mutation, no `agent.action.*` audit) when another pass already holds the PR's lock — the scenario #2129 describes, using the same settings/PR shape a real webhook or sweep pass would see.

Full unsharded `test:coverage` green (5599 passed); `typecheck` green; `npm audit` clean.

Advances #1936. Closes #2129.

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -28,6 +28,12 @@ declare global {
       get(key: string): Promise<string | null>;
       set(key: string, value: string, ttlSeconds: number): Promise<void>;
       del?(key: string): Promise<void>;
+      /** Atomic "set only if absent": returns true when this call newly claimed the key, false when it was
+       *  already held by someone else. Unlike a get-then-set pair, there is no window where two concurrent
+       *  callers can both observe an absent key and both claim it — the store (e.g. Redis SET NX) performs the
+       *  check-and-set as one operation. Optional so a cache adapter that hasn't implemented it yet still
+       *  type-checks; callers fall back to the non-atomic get/set pair when absent (#2129). */
+      claim?(key: string, value: string, ttlSeconds: number): Promise<boolean>;
     };
     /** TODO (convergence follow-up): a per-PR LOCK Durable Object (`SubmissionLock` mutex) is a separate,
      *  more-involved sub-task — it needs the ported DO class + its own migration tag, not just a binding here.
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -2369,10 +2369,22 @@ export async function claimAgentMaintenanceLock(
   repoFullName: string,
   prNumber: number,
 ): Promise<boolean> {
+  const key = agentMaintenanceLockKey(repoFullName, prNumber);
+  // Atomic claim (#2129): a get-then-set pair has a window between the read and the write where two concurrent
+  // passes for the SAME PR can both observe an absent key and both claim it, defeating the serializer entirely.
+  // env.SELFHOST_TRANSIENT_CACHE.claim performs the check-and-set as one operation (Redis SET NX server-side),
+  // closing that window. Falls back to the non-atomic get/set pair only for a cache adapter that hasn't
+  // implemented claim yet — strictly no worse than this function's prior behavior.
+  if (env.SELFHOST_TRANSIENT_CACHE?.claim) {
+    try {
+      return await env.SELFHOST_TRANSIENT_CACHE.claim(key, "1", AGENT_MAINTENANCE_LOCK_TTL_SECONDS);
+    } catch {
+      return true; // fail open — see the doc comment above.
+    }
+  }
   // getTransientKey/putTransientKey already fail open internally (a missing cache or a thrown read/write error
   // both resolve rather than throw), so this never needs its own try/catch — a cache fault surfaces here as
   // "no lock held", which correctly falls through to claiming it.
-  const key = agentMaintenanceLockKey(repoFullName, prNumber);
   if (await getTransientKey(env, key)) return false; // another pass is already in-flight for this PR
   await putTransientKey(env, key, "1", AGENT_MAINTENANCE_LOCK_TTL_SECONDS);
   return true;
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -1571,6 +1571,48 @@ async function maybeRunAgentMaintenance(
   if (pr.isDraft) return;
   if (!gate) return;
 
+  // Per-PR mutual exclusion (#2129): a webhook re-review and a sweep-driven agent-regate-pr job use different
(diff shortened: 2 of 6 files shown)


JSONbored:
The #2368 verification confirms my own analysis — recommend merging as-is; the deeper issue found (AI-review-cache races producing divergent gate verdicts) is real but pre-existing and out of scope for that PR. I'll flag that as a follow-up and now focus on [#2369](https://github.com/JSONbored/gittensory/issues/2369)'s actual, confirmed bug via a workflow (implement + adversarially verify), since this is the second time this exact code area got a legitimate finding.

PR #2372, opened 2026-07-01
docs: align CONTRIBUTING patch coverage with codecov.yml (99%)

bohdansolovie:
## Summary
- Update CONTRIBUTING.md patch-coverage wording from 97% to 99% to match `codecov.yml` (#1975).

## Test plan
- [x] Doc-only change


JSONbored (commented):
Fix this, then resubmit: 

```
The main maintainability detail is that the new test still hard-codes the same target string it reads from config instead of deriving the expected documentation text from that value.
```

PR #2374, opened 2026-07-01
test(review): cover review-thread blocker title parsing

bohdansolovie:
## Summary
- Add unit tests for `buildReviewThreadBlocker` and `reviewThreadBlockerFinding` (scanner preference, markdown/XML priority markers, location formatting).

## Test plan
- [x] `npx vitest run test/unit/review-thread-findings.test.ts`


JSONbored (commented):
Fix this, then you can resubmit: 

```
The most notable gap is that one fixture includes a discussion URL without asserting whether that value is intentionally preserved or irrelevant.
```

PR #2376, opened 2026-07-01
test(github): cover ready_for_review webhook coalesce key

bohdansolovie:
## Summary
- Assert `ready_for_review` pull_request webhooks coalesce to the same `pr-refresh` key as opened/sync events.

## Test plan
- [x] `npx vitest run test/unit/github-webhook-coalesce.test.ts`


JSONbored (commented):
Problems: 

```
The most notable limitation is that the test uses a different PR/SHA than the opened case, so it proves the key format but not literal coalescing with an otherwise identical opened payload.
```

Closing.

PR #2379, opened 2026-07-01
fix(github): compare PR head SHAs case-insensitively

bohdansolovie:
## Summary
- Normalize PR head SHAs to lowercase before freshness comparison so mixed-case GitHub responses do not false-positive as `head_changed`.

## Test plan
- [x] `npx vitest run test/unit/pr-freshness.test.ts`


JSONbored:
Moving forward, please only open 2 PRs max - anything above that will automatically be closed.

JSONbored (approved):

PR #2415, opened 2026-07-01
fix(selfhost): target public-token admission for background jobs

JSONbored:
### Motivation
- Public GitHub REST observations for the shared `GITHUB_PUBLIC_TOKEN` are now persisted with the `public-token` admission key, but self-host queue admission still only matched installation keys or legacy NULL rows, causing public-token-depleted observations to be ignored for no-installation background jobs.
- The change ensures background jobs that run against the shared public token are pre-yielded when the public-token bucket is exhausted, preventing unnecessary queue/API churn.

### Description
- Route GitHub-budget background jobs without an installation id to the shared public-token admission bucket by updating `githubRateLimitAdmissionTargetForJob` in `src/selfhost/queue-common.ts` to use `admissionKey ?? githubRateLimitAdmissionKeyForPublicToken()` for background jobs.
- Add unit coverage for the new targeting behavior and for preserving webhook semantics in `test/unit/selfhost-queue-common.test.ts`.
- Update SQLite queue tests to assert `admission_key='public-token'` observations pre-yield public-token background work and update metric expectations in `test/unit/selfhost-sqlite-queue.test.ts`.
- Update Postgres queue tests to assert public-token admission behavior and adjust keyed-rate-limit retry expectations in `test/unit/selfhost-pg-queue.test.ts`.

### Testing
- Ran focused unit tests: `npx vitest run test/unit/selfhost-queue-common.test.ts test/unit/selfhost-sqlite-queue.test.ts test/unit/selfhost-pg-queue.test.ts` and the suite passed (all tests in those files passed).
- Ran quick repository checks: `git diff --check` and `npm run typecheck` which succeeded locally.
- Attempted full local gate `npm run test:ci` and `npm audit --audit-level=moderate` but these were blocked by environment/network limitations (actionlint setup could not reach github.com and `npm audit` returned 403 from the registry), so the full `test:ci` run could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a44de395530832cbc5039c9ed136ffe)

### DIFF
--- a/src/selfhost/queue-common.ts
+++ b/src/selfhost/queue-common.ts
@@ -368,9 +368,10 @@ export function githubRateLimitAdmissionTargetForJob(
     };
   }
   if (!isGitHubBudgetBackgroundJob(message)) return null;
+  const admissionKey = githubRateLimitAdmissionKeyForJob(message);
   return {
     kind: "background",
-    admissionKey: githubRateLimitAdmissionKeyForJob(message),
+    admissionKey: admissionKey ?? githubRateLimitAdmissionKeyForPublicToken(),
   };
 }
 
--- a/test/unit/selfhost-pg-queue.test.ts
+++ b/test/unit/selfhost-pg-queue.test.ts
@@ -615,14 +615,14 @@ describe("createPgQueue (durable #977)", () => {
     }
   });
 
-  it("pre-yields GitHub-budget background jobs without repo fields from global REST observations", async () => {
+  it("pre-yields public-token GitHub-budget background jobs without installation ids", async () => {
     vi.useFakeTimers({ toFake: ["Date"] });
     vi.setSystemTime(new Date("2026-06-24T12:00:00.000Z"));
     const oldJitter = process.env.QUEUE_RATE_LIMIT_JITTER_MS;
     process.env.QUEUE_RATE_LIMIT_JITTER_MS = "0";
     try {
       const m = makePool();
-      m.setRateLimitRows([{ admission_key: null, repo_full_name: "owner/repo", remaining: "120", reset_at: "2026-06-24T12:10:00.000Z", observed_at: "2026-06-24T12:00:00.000Z" }]);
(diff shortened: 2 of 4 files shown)


JSONbored:
@gittensory gate-override The flagged blocker is a false positive.

The reviewer conflated two separate rate-limit admission mechanisms in `src/selfhost/queue-common.ts`:

1. **Persisted-observation lookup** (`githubRateLimitAdmissionDelayMs`, called from `rateLimitAdmissionDelayMs` in both `sqlite-queue.ts` and `pg-queue.ts`) — this is what actually reads `github_rate_limit_observations` rows and decides whether to pre-yield a job. Its SQL query still unconditionally selects `admission_key IS NULL` rows via the `fallback_observation` CTE, and `fallbackObservationCanOverrideExact` still honors that legacy row (via timestamp comparison) exactly as before this PR. **This PR does not touch this path's backward compatibility at all** — a pre-existing self-host installation's legacy NULL-keyed observation rows are still correctly retrieved and still govern admission for a `"public-token"`-targeted job when they're the newer/only signal.

2. **Immediate in-memory batch-defer** (`matchesGitHubRateLimitAdmissionTarget` / `deferPendingJobsForRateLimit`) — this is what `test/unit/selfhost-sqlite-queue.test.ts:1345` and `test/unit/selfhost-pg-queue.test.ts:1066` actually exercise. It never reads `github_rate_limit_observations` at all; it purely compares freshly-computed in-memory admission targets against the ONE bucket that just threw a live rate-limit error. Its behavior *did* change: a no-installation background job (now targeting the distinct `"public-token"` bucket) is no longer blanket-deferred just because an *unrelated* installation's bucket got rate-limited. That's the intended precision improvement this PR describes ("preventing unnecessary queue/API churn"), not a regression — the two buckets are genuinely independent on GitHub's side.

Verified via three independent code traces (two blind investigations + an adversarial skeptic pass explicitly tasked with trying to refute the false-positive conclusion), all reaching the same verdict with exact file:line citations. No functional change needed.

PR #2448, opened 2026-07-02
fix(signals): align focus-manifest.ts's glob cap with group-based counting

JSONbored:
## Summary

- `focus-manifest.ts`'s `normalizeOptionalGlob` (parses `contentLane.entryFileGlob`/`providerFileGlob`/`artifactGlob` from `.gittensory.yml`) capped raw `*` **character** count at 3 — the same flawed model change-guardrail.ts's `globToRegExp`/`matchesAny` used before #2445's fix. A `**` pair compiles to a single `.*` group, not two independent wildcards, so raw character counting is wrong in both directions: it wrongly rejects a legitimate 2-group glob like `"public/**/*.json"` (3 raw star characters) in some cases, and would wrongly admit a genuinely dangerous 3-single-star-group glob (e.g. `"a-*-*-*-b"`) at the same raw count — exactly the class of bug #2445 fixed.
- **Not a live security hole.** `globToRegExp` (from #2445) already has its own internal group-based cap (`MAX_GLOB_WILDCARD_GROUPS = 2`) that safely short-circuits any over-complex glob to a never-matching sentinel regex regardless of what this parse-time check lets through — full defense-in-depth already exists. This closes a **consistency/UX gap**: today some over-complex `contentLane.*Glob` values get an explicit "too many wildcards" parse-time warning while others silently pass validation and then silently compile to a never-matching pattern at match-time with zero warning.
- Fix: export `change-guardrail.ts`'s `countWildcardGroups`/`MAX_GLOB_WILDCARD_GROUPS` (both already benchmarked and battle-tested in #2445) and reuse them directly in `focus-manifest.ts` instead of re-deriving a second, separately-maintained threshold — the two glob-safety checks in this codebase now share one source of truth instead of risking future drift. `change-guardrail.ts` has zero imports of its own (confirmed dependency-free), so this is a safe, non-circular sibling import within `src/signals/`.

## Scope

- [x] This PR is focused (glob-cap consistency only) and does not mix unrelated backend, UI, MCP, docs, dependency, and deploy changes.
- [x] I linked an issue, or this is small enough that the summary explains why an issue is not needed. (No issue — a small, self-contained follow-up flagged during #2445's review, same class of fix.)

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally — the one pre-existing sub-100% line in `focus-manifest.ts` (`MAX_FOCUS_MANIFEST_BYTES` over-length content path) is untouched by this diff and unrelated; every line/branch this PR changes is covered
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`
- [x] New or changed behavior has unit/integration tests for new branches, fallback paths, and sanitizer boundaries — added a REGRESSION test proving `"public/**/*.json"` (2 groups, 3 raw characters) is now accepted, and updated the existing "too many wildcards" test's "at cap" example to the new 2-group boundary

## Safety

- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. (N/A — no such changes.)
- [ ] API/OpenAPI/MCP behavior is updated and tested where needed. (N/A — internal config-parsing helper only.)
- [ ] UI changes use live API data or real empty/error/loading states. (N/A — no UI changes.)
- [ ] Visible UI changes include a `UI Evidence` section. (N/A — no UI changes.)
- [ ] Public docs/changelogs are updated where needed. (N/A — no public-facing behavior change; this is an internal consistency fix.)

## Notes

- This targets `claude/pr-c-guardrail-redos-hardening` (#2445) rather than `main` because it imports `countWildcardGroups`/`MAX_GLOB_WILDCARD_GROUPS` from that PR's `change-guardrail.ts` changes; rebasing onto `main` once #2445 merges is expected to be trivial (this PR only touches `focus-manifest.ts` and its test file beyond the shared import, neither touched by #2445 itself).
- Flagged as a follow-up during #2445's own adversarial review (the gate found the original ReDoS fix's threshold was inconsistent with its own benchmark data) — this closes the parallel gap in the sibling file that has the same class of counting bug but isn't a live exploit path today.

### DIFF
--- a/src/signals/change-guardrail.ts
+++ b/src/signals/change-guardrail.ts
@@ -23,14 +23,18 @@ function canonicalize(value: string): string {
 // matchesAny below) so every caller — present or future, direct or indirect, including
 // content-lane/spec-resolver.ts's config-driven globs — is protected automatically. A caller with a STRICTER
 // fail direction (e.g. matchesAny's fail-toward-guarding for a security guardrail) overrides at its own layer.
-const MAX_GLOB_WILDCARD_GROUPS = 2;
+// Exported so any OTHER glob-safety check in this codebase (e.g. focus-manifest.ts's parse-time
+// normalizeOptionalGlob, which rejects an over-complex contentLane.*Glob before it ever reaches globToRegExp)
+// shares this exact threshold instead of drifting with its own re-derived number.
+export const MAX_GLOB_WILDCARD_GROUPS = 2;
 
 /** Count `*` GROUPS in `glob` — a `**` pair is ONE group (it compiles to a single `.*`, see globToRegExp), not
  *  two. Mirrors globToRegExp's own tokenization exactly (including consuming a `**`'s trailing `/`) so the count
  *  reflects the actual number of backtracking-capable groups the compiled RegExp will contain, not raw `*`
  *  character count (which would double-count every globstar and reject legitimate globs like
- *  "public/**\/*.json" — 2 real groups — as if they were 3-groups-dangerous). */
-function countWildcardGroups(glob: string): number {
+ *  "public/**\/*.json" — 2 real groups — as if they were 3-groups-dangerous). Exported for reuse by any other
+ *  glob-safety check in this codebase — see MAX_GLOB_WILDCARD_GROUPS above. */
+export function countWildcardGroups(glob: string): number {
   let count = 0;
   for (let i = 0; i < glob.length; i += 1) {
     if (glob.charAt(i) !== "*") continue;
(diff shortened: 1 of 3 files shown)


JSONbored:
Superseded by 536e46a2 ("fix(content-lane): share the wildcard-safety predicate between glob parsing and compilation"), already merged into claude/pr-b-content-lane-genericity — same fix, done concurrently in another session. Closing to avoid duplicate/conflicting work.

PR #2468, opened 2026-07-02
fix(enrichment): join adjacent added lines before secret-scan matching

JSONbored:
## Summary

- `scanPatch` (the secret-scan analyzer's only live entry point, via `scanSecrets`/`secret/descriptor.ts`) matched each added diff line independently, so a real credential split across two added lines via string concatenation (`const a = "AKIA..."; const b = a + "REST";`) never appeared as a single matchable token on either line and evaded every `RULES` regex.
- Extracts quoted string-literal contents per added line and, when a line doesn't already match on its own, joins the immediately-preceding line's last literal with the current line's first literal and tests that too — downgraded to `medium` confidence since a joined pair is a heuristic, not a direct contiguous match.
- Bounded to consecutive added lines only: a hunk boundary, a context line, or a removed line resets the join window, so this only ever catches the realistic "two sequential variable assignments" shape, not an unbounded cross-file join.
- Existing per-line detection is untouched and unweakened.

## Scope

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck` (root — unaffected, no `src/**` changed)
- [x] `npm run test:coverage` — N/A for this PR's own diff: `review-enrichment/**` is a separate workspace excluded from the Codecov `src/**` patch gate (own `npm run rees:test` runner instead, see below)
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`

Also ran `npm run rees:test` (the analyzer package's own gate, part of `test:ci`): 374/374 passing, including a new `review-enrichment/test/secret-scan.test.ts` (9 tests) covering the pre-existing single-line rules plus 5 new tests for the cross-line join: the concatenation-split regression case itself, a context-line break, a hunk-boundary break, no-double-report when a line already matched directly, and a false-positive guard for two unrelated short literals. `npm run db:migrations:check` also unaffected/green.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed. Test fixtures build fake secret-shaped strings from split fragments at runtime rather than as contiguous literals, specifically so the fixture file itself doesn't trip GitHub push protection the way a real leaked credential would.
- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes — N/A, this is an enrichment analyzer, not an auth path.
- [x] API/OpenAPI/MCP behavior is updated and tested where needed. No OpenAPI/MCP surface changed.
- [ ] UI changes — N/A, no UI changed.
- [ ] Visible UI changes — N/A, no UI changed.
- [x] Public docs/changelogs are updated where needed; changelogs are only edited for release-prep PRs. No changelog edit.

## Notes

- Found via an adversarial audit of the review engine's security-relevant analyzers (which gate what counts as risky code in a PR). `scanAddedLinesForSecrets` (a second, unused export in the same file with the same per-line-only gap) was left untouched — it has zero callers anywhere in the codebase, so fixing it would add untestable dead-code coverage; out of scope for this fix.

### DIFF
--- a/review-enrichment/src/analyzers/secret-scan.ts
+++ b/review-enrichment/src/analyzers/secret-scan.ts
@@ -47,33 +47,65 @@ const RULES: Rule[] = [
   },
 ];
 
+/** Extract the inner text of every quoted string literal (single/double/backtick) on a line. Used to catch a
+ *  secret whose literal value is split across two adjacent added lines and joined at runtime (e.g.
+ *  `const a = "AKIA..."; const b = a + "REST";`) — pure per-line regex matching never sees the runtime-joined
+ *  value, only the two separate source literals either side of the `+`. */
+function extractStringLiteralContents(line: string): string[] {
+  const literals: string[] = [];
+  const re = /"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'|`(?:[^`\\]|\\.)*`/g;
+  let match: RegExpExecArray | null;
+  while ((match = re.exec(line)) !== null) literals.push(match[0].slice(1, -1));
+  return literals;
+}
+
 /** Scan one file's unified-diff patch, tracking new-file line numbers via hunk headers. Pure. Value never captured. */
 export function scanPatch(path: string, patch: string): SecretFinding[] {
   const findings: SecretFinding[] = [];
   let newLine = 0;
+  // Last added line's extracted string-literal contents, for the cross-line join check below. Reset whenever a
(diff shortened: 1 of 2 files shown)


JSONbored:
Secret key flag is a false positive, merging.

PR #2474, opened 2026-07-02
fix(signals): scan full open-issue sample for pre-start title matching

joaovictor712:
## Summary

Title-only check-before-start capped Jaccard matching at openIssues.slice(0, 300). Stale issues beyond index 300 were invisible to title resolution. Scan the full bounded open-issue set callers pass.

## Scope

Focused engine + regression test. No guarded paths.

## Validation

npx vitest run test/unit/pre-start-check.test.ts -t title scan cap


JSONbored (commented):
Closing: 

```
the main remaining gap is that the test locks the old boundary rather than the full caller-sample contract.
```

PR #2475, opened 2026-07-02
fix(review): redact duplicate surface summaries

JSONbored:
### Motivation

- Prevent leaking unvalidated, attacker-controlled duplicate entry fields (notably `url`) into public gate/finding text before the normal secret and URL validation steps run.  
- Address the vulnerability where duplicate-detection returned a summary containing raw `url` content which could include PAT-like tokens, wallet/hotkey text, newlines, or Markdown.

### Description

- Replace the duplicate-entry close summary to be fully generic and remove interpolation of the duplicate entry's `url` so duplicate detection no longer echoes unvalidated content.  
- Stop passing the duplicate entry into the summary generator and instead return the stable generic message from `duplicateEntryCloseSummary()` in `src/review/content-lane/orchestrator.ts`.  
- Update unit tests in `test/unit/content-lane-orchestrator.test.ts` to assert the generic close summary for same-PR and base-resubmission duplicates.  
- Add a regression test that constructs an unsafe duplicate `url` (PAT-like token + newline + Markdown) and asserts it is not reflected in the returned summary.

### Testing

- Ran `git diff --check` which reported no whitespace/conflict issues.  
- Ran the targeted unit file with `npx vitest run test/unit/content-lane-orchestrator.test.ts --reporter=dot` and the test file passed (all tests in that file succeeded).  
- Ran typecheck via `npm run typecheck` which succeeded.  
- Attempted a repository-wide coverage run with `npm run test:coverage`, which executed but failed the global coverage threshold (the project-level coverage requirement prevented the run from passing in this environment).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a45eed98140832ca5d0275684405aa0)


JSONbored:
False positive re: secret flag, merging.

PR #2483, opened 2026-07-02
fix(github): reject malformed repo slugs in label helpers

bohdansolovie:
## Summary

- Require exactly `owner/repo` in label mutation helpers, matching the merged `pr-actions` fail-closed contract.
- Reject multi-segment and padded slugs before any GitHub label create/apply/delete call.

## Test plan

- [x] `npx vitest run test/unit/github-labels.test.ts`

### DIFF
--- a/src/github/labels.ts
+++ b/src/github/labels.ts
@@ -6,6 +6,17 @@ type GitHubLabel = {
   name?: string | null;
 };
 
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Blockers, closing: 

```
The notable remaining gap is test precision around the empty-owner/empty-repo branch of the new parser, not a correctness defect in the implementation.
```

PR #2491, opened 2026-07-02
fix(github): evict a cached App JWT on rejection to stop fleet-wide token-minting poisoning

### LINKED ISSUE #2453: fix(github): evict a cached App JWT on rejection to stop fleet-wide token-minting poisoning
Parent: #1936

## Problem

`createAppJwt` (`src/github/app.ts` ~line 381) caches the App-level JWT for 8 minutes (`appJwtCache`, keyed only by `GITHUB_APP_ID`), with no eviction path on rejection. #2406 already hardened this for *installation* tokens (`expireCachedInstallationToken` + retry-once on a rejected installation token via `withInstallationTokenRetry`), but the App-level JWT itself has no equivalent.

If GitHub rejects the currently-cached App JWT for any transient reason (a validation hiccup, a clock-skew edge case, a brief App suspend/reinstate) while `env.GITHUB_APP_PRIVATE_KEY` is unchanged, `mintInstallationToken` (`app.ts` ~lines 244-251) keeps handing back the same poisoned JWT from cache (nothing about the cache-validity check — keyed on `privateKey === env.GITHUB_APP_PRIVATE_KEY` and `expiresAtMs > nowMs` — changes on a rejection). Every subsequent `POST /app/installations/{id}/access_tokens` across **every installation on the instance** gets a 401 and throws, until the 8-minute cache window naturally lapses. `appJwtCache` is otherwise only ever cleared by the test-only `clearInstallationTokenCacheForTest()`.

Production call sites of `createInstallationToken` (`src/queue/processors.ts`, `src/services/agent-approval-queue.ts`, `src/services/agent-action-executor.ts`, `src/github/backfill.ts`, `src/github/labels.ts`, `src/review/rag-index.ts`, `src/review/grounding-wire.ts`) either propagate the throw or silently swallow it — none re-signs a fresh JWT. Net effect: a single JWT rejection can stall merges/comments/check-runs/approvals fleet-wide for up to 8 minutes.

## Fix

On a 401 (or bad-credentials-shaped error) from the App-installations access-token endpoint, evict the current `appJwtCache` entry for `env.GITHUB_APP_ID`, re-sign a fresh JWT, and retry once — mirroring the `withInstallationTokenRetry` pattern already used for installation tokens.

## Regression tests

- A rejected App JWT triggers eviction + one fresh re-sign + retry, and a subsequent installation-token mint succeeds without waiting out the cache TTL.
- A JWT rejection for a genuinely bad private key (not transient) does not infinite-loop — still terminates after the single retry.

luciferlive112116:
## Summary

- Evict the cached App JWT and retry installation-token mint once when GitHub returns 401, mirroring the existing `withInstallationTokenRetry` pattern for installation tokens.
- Prevents a transient rejected App JWT from poisoning every installation-token mint fleet-wide for up to 8 minutes.

Fixes #2453

## Test plan

- [x] `REGRESSION (#2453): evicts a rejected App JWT and retries the mint once instead of failing outright`
- [x] `REGRESSION (#2453): does not infinite-loop when the retried App JWT is ALSO rejected`

### DIFF
--- a/src/github/app.ts
+++ b/src/github/app.ts
@@ -192,6 +192,21 @@ export async function withInstallationTokenRetry<T>(
   }
 }
 
+/** POST the App-installations access-token endpoint with a given JWT. Extracted so mintInstallationToken can
+ *  issue the same request twice — once with the cached JWT, once with a freshly-signed one on a 401 (#2453). */
+function requestInstallationTokenWithJwt(
+  jwt: string,
+  installationId: number,
+): Promise<Response> {
+  return timeoutFetch(
+    `https://api.github.com/app/installations/${installationId}/access_tokens`,
+    {
+      method: "POST",
+      headers: githubHeaders(`Bearer ${jwt}`),
+    },
+  );
+}
+
 /** Mint a fresh installation token (broker or local App-JWT) and cache it. `cached` is the expired/absent prior
  *  entry, consulted only for the brokered stale-token grace. Extracted from createInstallationToken so that
  *  function can single-flight concurrent cold-cache callers onto one mint (see inFlightMints). */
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Maintainer-only issue, and also, test gaps: 

```
the main weakness is that the regression test proves retry behavior but does not prove the retried request used a newly signed JWT rather than the same cached credential.
```

Closing.

PR #2396, opened 2026-07-01
test(enrichment): cover actions-pin analyzer

Florcitaq:
Closes #2101.

## Summary
- Add dedicated REES unit coverage for the GitHub Actions pin analyzer.
- Cover mutable third-party tag/branch refs, full-SHA and official-action exclusions, quoted `uses` keys, ignored unchanged/removed/non-uses lines, line citations, and workflow-file filtering through `scanActionPins`.

## Validation
- `npm ci --prefer-offline --no-audit --no-fund`
- `npm run build`
- `npx prettier --check test/actions-pin.test.ts`
- `node --test --experimental-strip-types "test/actions-pin.test.ts"`
- `npm run validate:sourcemaps`
- `git diff --check`

Known baseline: `node scripts/generate-analyzer-metadata.mjs --check` reports `review-enrichment/analyzer-metadata.json` stale on current `main`; this PR is test-only and does not update generated analyzer metadata.

PR #2492, opened 2026-07-02
fix(enrichment): treat uppercase action SHA pins as pinned

bohdansolovie:
## Summary

- Match GitHub commit refs case-insensitively in the actions-pin analyzer.
- Stop reporting uppercase 40-character SHA pins as mutable third-party workflow references.

## Why production-only

Regression coverage already lives on `main` in `review-enrichment/test/actions-pin.test.ts` (#2396). This PR changes only the analyzer so it does not collide with that test file.

## Test plan

- [x] `npm run rees:test -- test/actions-pin.test.ts`


JSONbored (commented):
Closing: 

```
the test diff deletes several existing regression cases that covered line attribution, quoted `uses` syntax, `.yaml` workflow files, unchanged lines, missing patches, and positive findings through `scanActionPins`. The regex change itself is safe enough, but the test rewrite weakens coverage around behavior this analyzer still owns.
```

Issue #2458, opened 2026-07-02
feat(mcp): thread size-gate and guardrail signals into the predicted-gate verdict


JSONbored:
Parent: #1936

## Problem

`buildPredictedGateVerdict` (`src/rules/predicted-gate.ts` ~line 230) never threads `sizeGateMode`/changed-file/line counts or `guardrailHit` into `evaluateGateCheck`, so the pre-submission predictor (the `predict-gate` MCP tool and its REST endpoint) can never predict the oversized-PR or guardrail-path manual-review HOLDs the live gate enforces (`evaluateGateCheck`'s `buildSizeHoldFinding`/guardrail-hold branches in `src/rules/advisory.ts`). `predictedGateNote` (~lines 56-74) discloses other known predictor gaps (slop, manifest-policy) but not this one.

`changedPaths` is already available to the predictor and used by the adjacent manifest-policy block (~lines 197-218) and pre-merge-check block (~lines 188-190) — the size/guardrail counts are computable from the same input, just never wired in.

**Concrete effect**: a miner calls `predict-gate` with `changedPaths` that hit a maintainer-configured hard-guardrail path, or that total over the configured size threshold. The predictor returns `conclusion: "success"` ("will pass"), while the live gate — running the identical `evaluateGateCheck`/`evaluateGateCheckCore` against the same real PR — returns a manual-review hold instead of merging. The miner is told the PR will merge and instead gets a surprise hold, with no disclaimer.

## Fix

Thread `sizeGateMode` + file/line counts (derivable from `changedPaths`) and `guardrailHit` (via `isGuardrailHit`/`loadHardGuardrailGlobs` from `src/signals/change-guardrail.ts`) into the `evaluateGateCheck` call at `predicted-gate.ts` ~line 230 whenever `hasChangedPaths` is true, mirroring how the manifest-policy block already conditions on `hasChangedPaths`. At minimum, add an explicit disclaimer to `predictedGateNote` for the size/guardrail gate, matching the existing slop-gap disclaimer.

## Regression tests

- A predicted-gate call with `changedPaths` exceeding the size threshold predicts a hold, not success.
- A predicted-gate call with `changedPaths` hitting a hard-guardrail path predicts a hold, not success.
- Predictor output stays unchanged for PRs that don't trip either gate.

PR #2494, opened 2026-07-02
feat(mcp): thread size-gate and guardrail signals into the predicted-gate verdict

JSONbored:
## Summary

- `buildPredictedGateVerdict` never threaded `sizeGateMode`/`changedFileCount` or `guardrailHit` into `evaluateGateCheck`, so the pre-submission predictor (the `predict-gate` MCP tool + REST endpoint) could never predict the oversized-PR or guardrail-path manual-review HOLDs the live gate enforces. A miner supplying `changedPaths` that hit a hard-guardrail path, or that exceed the configured size threshold, was told `conclusion: "success"` ("will merge") while the live gate — running the identical `evaluateGateCheck` against the same real PR — would actually hold it for manual review (`neutral`, never auto-merged).
- `sizeGateMode` now reads directly from the manifest's public `gate.size.mode`, mirroring how every other gate mode is already read in this file.
- `changedFileCount` and `guardrailHit` are threaded only when `changedPaths` are supplied — the only size/guardrail input this metadata-only predictor ever receives, so behavior stays byte-identical without paths (matches the existing pattern for the manifest-policy block just above it).
- The hard-guardrail globs are imported directly as the constants they actually are — `loadHardGuardrailGlobs` is `async` only for processor call-graph compatibility; its body is 100% hardcoded arrays with no live fetch — keeping this function synchronous rather than threading a needless `await` through every caller.
- `changedLineCount` is deliberately left unset: line-diff stats are never sent to this predictor, so the size hold can only be predicted from file count. The note now discloses this once paths are supplied, replacing the no-paths disclaimer it superseded.

## Scope

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` — new code is 100% line+branch covered (`src/rules/predicted-gate.ts`: 40/40 lines, 69/69 branches). Global 96.55%/95.53%.
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:openapi:check` — no drift; the `PredictedGateVerdict` type shape is unchanged (only internal logic + the `note` string content changed)
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`

Full suite: 6070/6070 passing. Also ran `npm run db:migrations:check` (unaffected, green).

## Safety

- [x] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. N/A (not an auth path), but added regression tests: size-hold fires with `gate.size.mode` on + files over threshold, does NOT fire when the mode is off or the PR is within threshold; guardrail-hold fires for a hard-guardrail path (always-on, no config needed) and not for an ordinary path; the note text correctly swaps its disclaimer once paths are supplied.
- [x] API/OpenAPI/MCP behavior is updated and tested where needed. No schema shape changed, so no regeneration needed (verified via `ui:openapi:check`).
- [ ] UI changes — N/A, no UI changed.
- [ ] Visible UI changes — N/A, no UI changed.
- [x] Public docs/changelogs are updated where needed; changelogs are only edited for release-prep PRs. No changelog edit.

## Notes

- Lowest-severity finding from an adversarial audit of the review engine (the predictor giving an over-optimistic answer, not a live-gate correctness or security issue) — last of a 7-PR series from the same audit.

### DIFF
--- a/src/rules/predicted-gate.ts
+++ b/src/rules/predicted-gate.ts
@@ -7,9 +7,18 @@ import {
   type IssueQualityReport,
 } from "../signals/engine";
 import { buildFocusManifestGuidance, type FocusManifest } from "../signals/focus-manifest";
+import { isGuardrailHit } from "../signals/change-guardrail";
+import { CONFIG_AS_CODE_GUARDRAIL_GLOBS, DEFAULT_CRUCIAL_GUARDRAIL_GLOBS, ENGINE_DECISION_GUARDRAIL_GLOBS } from "../review/guardrail-config";
 import { sanitizePublicComment } from "../github/commands";
 import { GITTENSOR_HOME_URL } from "../github/footer";
 import type { BountyRecord, GatePolicyPack, IssueRecord, PullRequestRecord, RepositoryRecord } from "../types";
+
+// The live gate's hard-guardrail globs (src/review/guardrail-config.ts) are 100% hardcoded engine constants —
+// loadHardGuardrailGlobs there is only `async` for processor call-graph compatibility, its body just
+// concatenates these three exported arrays with no live fetch. Predicted-gate is metadata-only/synchronous by
+// design (its own docstring: sourced ONLY from the PUBLIC .gittensory.yml + safe defaults), so import the
+// constants directly rather than making this whole function async for a call that was never actually live.
+const PREDICTED_GATE_HARD_GUARDRAIL_GLOBS = [...DEFAULT_CRUCIAL_GUARDRAIL_GLOBS, ...CONFIG_AS_CODE_GUARDRAIL_GLOBS, ...ENGINE_DECISION_GUARDRAIL_GLOBS];
 
 // Opt-in funnel (#694): a non-Gittensor adopter running the `oss-anti-slop` pack learns that Gittensor pays
 // contributors for OSS work like this. Public-safe "earn" wording only (never reward/payout/score).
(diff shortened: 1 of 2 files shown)


JSONbored:
**Overriding the AI reviewer's blocker as a false positive.**

The review claims `sizeGateMode: gate.sizeMode ?? undefined` should be `gate.size?.mode` because "the manifest contract and tests use `gate.size.mode`". This mixes up two different objects:

- The test fixtures (`gate: { size: { mode: "advisory" } }`) are the **raw public `.gittensory.yml` shape**, passed as input to `parseFocusManifest(...)`.
- `manifest.gate` in `buildPredictedGateVerdict` is the **post-parse internal `FocusManifestGateConfig`** (src/signals/focus-manifest.ts:22-45), which has a flat `sizeMode: GateRuleMode | null` field — there is no nested `.size` property on that type at all. `gate.size?.mode` would not even typecheck against it.

Every sibling field in the same destructure (`gate.duplicates`, `gate.aiReviewMode`, `gate.linkedIssue`, ...) follows the identical flat-internal-field convention, fed from nested public YAML through the parser's own normalization (`normalizeOptionalGateMode(sizeRecord?.mode, "gate.size.mode", warnings)` at focus-manifest.ts:492).

The 3 regression tests for this exact code path (`test/unit/predicted-gate.test.ts`, "#2458") already pass end-to-end through the real `parseFocusManifest`, confirming the size-hold prediction works correctly as written.


- [x] Re-run Gittensory review

PR #2518, opened 2026-07-02
fix(github): reject malformed repo slugs in comment helpers

bohdansolovie:
## Summary

- Require exactly `owner/repo` in sticky comment helpers, matching the merged `pr-actions` fail-closed contract.
- Reject multi-segment and padded slugs before any GitHub comment create/update call.

## Test plan

- [x] `npx vitest run test/unit/github-comments.test.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/github/comments.ts
+++ b/src/github/comments.ts
@@ -19,6 +19,17 @@ type IssueComment = {
   } | null;
 };
 
+function parseRepoFullName(repoFullName: string): { owner: string; repo: string } {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing: 

```
The notable remaining gap is that validation is still slash-count based rather than segment-syntax based, so some malformed per-segment values can still reach GitHub.
```

PR #2581, opened 2026-07-02
feat(selfhost): add automatic dead-letter job retry for self-host queues

### LINKED ISSUE #2534: feat(selfhost): add automatic dead-letter job retry for self-host queues
Parent: #1936

## Problem
Self-host's SQLite and Postgres queue backends have no mechanism to move a job out of `status='dead'` once it lands there. There is a dead-letter redrive path (`src/queue/dlq.ts`) but it is wired exclusively to the Cloudflare Workers Queues consumer entrypoint and is not reachable from the self-host queue implementations at all. In practice this means: a job that dies from a transient or since-fixed bug (a bad deploy, a backend-compatibility gap, a flaky third-party dependency) requires a manual database UPDATE or DELETE to ever run again — indefinitely, even after the underlying cause is resolved and redeployed.

## Requirements
- Add a bounded, safe automatic retry path for dead-lettered self-host jobs: e.g. a scheduled maintenance pass that re-queues dead jobs whose `job_key`/type hasn't already been retried within some cooldown window, capped to a small number of automatic attempts before requiring manual intervention again.
- Must not create a retry storm: a job that fails the SAME way immediately after requeue should not be retried indefinitely — a hard ceiling on auto-retries per job is required, after which it stays dead and relies on the alerting from the sibling alerting issue.
- Should reuse existing queue primitives (attempts counter, `job_key`) rather than inventing a parallel bookkeeping table.

## Deliverables
- A retry/redrive mechanism in `src/selfhost/sqlite-queue.ts` and `src/selfhost/pg-queue.ts` (or a shared helper both call into), gated by a bounded auto-retry ceiling and cooldown.
- Metrics distinguishing an auto-recovered job from a fresh enqueue.
- Tests covering: a dead job recovers automatically after its cooldown, a job that dies repeatedly stops auto-retrying at the ceiling, and the ceiling/cooldown are configurable.

## Acceptance criteria
- A job dead-lettered by a bug that gets fixed and redeployed recovers automatically within the next maintenance cycle, without manual database intervention.
- A job that is genuinely broken (fails identically every time) does not retry forever and does not starve other queue throughput.
- No regression to existing dead-letter counting/alerting.

## Expected outcome
Self-host operators don't need direct database access to recover from a transient bug that's already been fixed upstream.

JSONbored:
## Summary

Closes #2534.

Neither self-host queue backend (SQLite or Postgres) has a way to move a job out of `status='dead'` once it lands there. The Cloudflare Workers Queues DLQ redrive path (`src/queue/dlq.ts`) is wired exclusively to that platform's native queue consumer and is not reachable from the self-host tables. In practice this means: a job that dies from a transient bug, or a bug that's since been fixed and redeployed, requires a manual database `UPDATE`/`DELETE` to ever run again — indefinitely.

## What changed

- Added `reviveDeadLetterJobs()` to both `createSqliteQueue` and `createPgQueue`: requeues dead jobs whose lifetime `attempts` is still under a ceiling (`maxRetries` + a small configurable extra-attempts budget, default 3), clearing `last_error`. `attempts` itself is left untouched, so a revived job gets exactly **one** more attempt before either succeeding or dying again — not a fresh full retry budget — bounding how many times a permanently-broken job can cycle before it stops being revived and requires manual intervention.
- Runs on a separate, much slower timer (default 30 minutes, `QUEUE_DEAD_LETTER_REVIVE_INTERVAL_MS`) than the normal poll tick — that interval *is* the cooldown between auto-retry rounds, so no separate timestamp bookkeeping was needed. Reuses the existing `attempts`/`job_key` columns rather than a parallel table, per the issue's requirement.
- New `gittensory_jobs_dead_letter_revived_total` metric, distinct from the existing crash-recovery `gittensory_jobs_recovered_total` counter (a different scenario — process-restart recovery vs. dead-letter revival).
- `reviveDeadLetterJobs()` is exposed on the public queue interface (not just wired to the internal timer), so it's directly testable without waiting on real intervals, and available for a future operator-triggered repair path.
- Documented the new behavior (and both env-var knobs) in the self-hosting Operations doc.

## Correctness/safety notes

- No retry-storm risk: the revive interval bounds how often any one job can be retried, and the `attempts < ceiling` filter bounds how many total revivals a job can get before it's left dead permanently.
- A job that dies again after revival returns to normal dead-letter handling (including the sibling alerting work in #2549) — this feature only gives a bounded number of extra chances, it never suppresses the existing dead-letter signal.

## Validation

- Full local gate (`npm run test:ci`) green, `npm audit --audit-level=moderate` clean, `git diff --check` clean.
- Every new/changed line and branch across `sqlite-queue.ts`, `pg-queue.ts`, and `queue-common.ts` is covered — cross-checked the diff hunks directly against `coverage/lcov.info`; the handful of uncovered lines/branches in those files are all pre-existing and unrelated to this change.
- New tests cover: revival under the ceiling with `last_error` cleared, the ceiling actually stopping further revival, no-op when nothing is dead, the periodic timer firing a real end-to-end revival (SQLite, fake timers), the equivalent mocked-pool behavior for Postgres, and both new env-var config knobs (defaults + overrides).

## Safety

- No secrets or credentials introduced or exposed.
- No behavior change to existing retry/dead-letter thresholds — purely additive.

### DIFF
--- a/apps/gittensory-ui/src/routes/docs.self-hosting-operations.tsx
+++ b/apps/gittensory-ui/src/routes/docs.self-hosting-operations.tsx
@@ -120,6 +120,16 @@ docker compose --profile postgres --profile observability --profile backup up -d
         check if you haven&apos;t wired up push notifications yet — it&apos;s exactly what the{" "}
         <code>Dead jobs stay at zero</code> routine check below is watching for.
       </p>
+      <p>
+        Dead-lettered jobs also get one automatic revival attempt every 30 minutes (
+        <code>QUEUE_DEAD_LETTER_REVIVE_INTERVAL_MS</code>), as long as the job hasn't already been
+        revived more than a small, bounded number of extra times (
+        <code>QUEUE_DEAD_LETTER_AUTO_RETRY_MAX_EXTRA_ATTEMPTS</code>, default 3) — so a job that
+        died from a bug that's since been fixed and redeployed recovers on its own within the next
+        cycle, without needing direct database access. A job that keeps failing the same way
+        eventually exhausts this budget and stays dead, which is exactly what the alert above is
+        watching for.
+      </p>
 
       <h2>Docker resource hygiene</h2>
       <p>
--- a/src/selfhost/pg-queue.ts
+++ b/src/selfhost/pg-queue.ts
@@ -229,11 +229,16 @@ export function createPgQueue(
     const maxJitter = queueRecoveryJitterMs();
     for (const row of res.rows as Array<{ id: string; payload: string; job_key?: string | null }>) {
       const runAfter = now + deterministicJitterMs(`revive:${row.job_key ?? ""}:${row.id}:${row.payload}`, maxJitter);
-      await pool.query(`UPDATE ${TABLE} SET status='pending', run_after=$1, last_error=NULL WHERE id=$2`, [
-        runAfter,
-        row.id,
-      ]);
-      revived += 1;
+      // AND status='dead' re-checks the row is STILL dead at UPDATE time (mirrors reclaimExpiredProcessingJobs /
+      // deferPendingJobsForRateLimit above) — the SELECT above is a stale snapshot, and without this predicate an
+      // overlapping reviver (another self-host instance, or a slow prior revive tick still running when the next
+      // one fires) could flip a row that's already been claimed into 'processing' back to 'pending', letting it
+      // run a second time concurrently. rowCount is 0 (not counted as revived) when another reviver won the race.
+      const update = await pool.query(
+        `UPDATE ${TABLE} SET status='pending', run_after=$1, last_error=NULL WHERE id=$2 AND status='dead'`,
+        [runAfter, row.id],
+      );
+      revived += update.rowCount ?? 0;
     }
     return revived;
   }
--- a/src/selfhost/pg-queue.ts
+++ b/src/selfhost/pg-queue.ts
@@ -24,6 +24,8 @@ import {
   jobPriority,
   parsePositiveIntEnv,
   queueBackgroundConcurrency,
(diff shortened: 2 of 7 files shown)


JSONbored:
Resolved the flagged defect: \`reviveEligibleDeadJobs\`'s SELECT-then-UPDATE is a stale snapshot — an overlapping reviver (a second self-host instance sharing the same Postgres DB, or a slow prior revive tick still running when the next fires) could move a row out of \`'dead'\` between the SELECT and this row's own UPDATE, and the UPDATE (keyed on `id` alone) would then blindly flip whatever the row's *current* status is back to \`'pending'\` — including a row already claimed into \`'processing'\`, letting it run a second time concurrently.

Both backends now add \`AND status='dead'\` to the revive UPDATE, mirroring the identical defensive predicate already used elsewhere in both files for the same class of stale-snapshot sweep (\`reclaimExpiredProcessingJobs\` / \`deferPendingJobsForRateLimit\`). The revived count now reflects rows the UPDATE actually matched, not the raw SELECT count.

Added a regression test per backend:
- Postgres: a mock pool with per-call configurable UPDATE \`rowCount\`s, proving a 0-rowCount UPDATE isn't counted as revived.
- SQLite: a \`driver.query\` spy that injects the competing status change at the exact point the real UPDATE would otherwise race against it, proving the row is left untouched (not reverted to pending).

Also rebased onto current \`main\` (two other PRs merged in the interim). Full local gate (\`test:ci\`, unsharded \`test:coverage\`, \`npm audit\`) green; every changed line/branch in \`pg-queue.ts\`/\`sqlite-queue.ts\` is covered (cross-checked the remaining uncovered branches directly against the diff — all pre-existing and unrelated).

PR #2595, opened 2026-07-02
feat(github): cache bare PR reads and reviews, mirroring the head-SHA sync-state pattern

### LINKED ISSUE #2537: feat(github): extend the head-SHA snapshot cache to live PR reads and reviews
Parent: #1936

## Problem
A recently-merged fix (#2527) added a durable repo+PR+head-SHA cache for a PR's changed files, cutting redundant `GET /pulls/{n}/files` calls. Live rate-limit observation data during a near-exhaustion window showed two sibling routes as even larger contributors that the same fix deliberately left uncached:

- A bare `GET /pulls/{n}` (PR state/head/mergeable-state) fetch, implemented as four separate near-identical helper functions with no caching or cross-call coalescing at all, called from many independent sites (readiness checks, freshness guards, the maintenance planner, and once per planned auto-maintain action) — several of which fire multiple times for the same PR within one review pass.
- `GET /pulls/{n}/reviews`, which was explicitly left out of the earlier fix ("reviews are more volatile than files" — true at a fixed head, but reviews only actually change on a `pull_request_review` webhook, not on every sweep tick).

Together these two routes were the largest and third-largest contributors observed in a live rate-limit trough, larger than the files route the earlier fix addressed.

## Requirements
- Add a short-TTL or event-invalidated cache for the bare PR read, keyed by repo+PR, refreshed by the relevant webhook types (synchronize/closed/reopened) and read by the freshness-guard/readiness/maintenance-planner call sites — while explicitly preserving any call site that intentionally bypasses caching because it sits at a mutation-act boundary (there is at least one such site today, documented in its own code comment; do not weaken it).
- Extend the existing head-SHA file-cache pattern to reviews, but invalidate on `pull_request_review` webhook events (submitted/dismissed/edited) rather than on head-SHA change, since reviews are independent of the head.
- Preserve exact current behavior for any call site that must observe a live, uncached read (act-boundary merge/close decisions).

## Deliverables
- A cache for the bare PR-state read, applied to the read-only call sites (freshness guards, readiness checks) while leaving the documented act-boundary call site live.
- A head-SHA/review-webhook-scoped cache for `fetchPullRequestReviews`, following the same storage pattern as the existing files cache.
- Tests mirroring the existing files-cache test suite: cache miss on first read, cache hit on unchanged state, cache invalidation on the relevant webhook, and a regression test proving act-boundary reads are never served from cache.

## Acceptance criteria
- Both routes' call volume drops materially for repeat reads of an unchanged PR, verified via the existing bounded-caller metrics pattern.
- No regression in any decision that depends on a live PR state (merge eligibility, close eligibility, mergeable-state gating).
- Full local gate green with near-100% patch coverage on the new caching logic, matching the standard for this repo's Codecov gate.

## Expected outcome
The GitHub REST budget consumed by routine PR-state and review polling drops by a similar order of magnitude to what the files cache already achieved, without weakening any correctness-critical live read.

JSONbored:
## Summary

Closes #2537.

`GET /pulls/{n}` (PR state/mergeable_state) and `GET /pulls/{n}/reviews` were the largest and third-largest GitHub REST contributors observed in a live rate-limit trough — larger than the PR-files route the earlier fix (#2527) addressed, and deliberately left uncached by that fix. The bare-PR-state read was implemented as four near-identical helper functions (`fetchLivePullRequestMergeState`, `fetchLivePullRequestState`, `fetchLivePullRequestHeadSha`, `fetchLivePullRequest`) with no caching or cross-call coalescing, called from several independent read sites (readiness checks, the maintenance planner, duplicate-sibling reconciliation, a gate-override command). Reviews were explicitly excluded from #2527 ("reviews are more volatile than files" — true at a fixed head, but reviews only actually change on a `pull_request_review` webhook, not on every sweep tick).

## What changed

- **Schema**: extends `pull_request_detail_sync_state` (migration `0093`) with `prMergeableState`/`prState`/`prStateFetchedAt`, and reuses the already-present (previously-unused) `reviewsSyncedAt` column as the review cache's freshness marker — no new table.
- **PR-state cache** (`src/github/backfill.ts`): new `cachedFetchLivePullRequestMergeState` / `cachedFetchLivePullRequestState` / `cachedFetchLivePullRequestHeadSha`, each a read-through wrapper over the durable row with a 5-minute safety-net TTL (`PR_STATE_CACHE_MAX_AGE_MS`) so a missed webhook self-heals within one sweep tick. `primeDurablePrStateCache` lets an already-paid-for full-payload fetch (the sweep resync) prime the cache for other readers instead of them re-fetching moments later. Wired into every **read-only** call site (readiness, duplicate-sibling reconciliation, the gate-override command).
- **Reviews cache** (`fetchAndStorePullRequestDetails`): gated purely on `reviewsSyncedAt` being set — no head-SHA comparison, since reviews are independent of the head (unlike files). Only stamps `reviewsSyncedAt` when the review fetch actually succeeded that pass (a failed sync must not be cached as if it succeeded).
- **Webhook invalidation** (`src/queue/processors.ts`): `invalidatePrStateCache` on `pull_request` `synchronize`/`closed`/`reopened`; `invalidatePrReviewsCache` on `pull_request_review` `submitted`/`edited`/`dismissed` (already the exact 3 actions `shouldProcessPullRequestPublicSurface` restricts that event to). Both explicitly null the relevant columns (PARTIAL-UPDATE CONTRACT: omitted = unchanged, explicit `null` = cleared).
- **Metrics**: `gittensory_pr_state_cache_total{field,result}` and `gittensory_pr_reviews_cache_total{result}`, mirroring the existing `gittensory_github_pull_request_files_fetch_total` bounded-label style.
- **Act-boundary preserved, unchanged**: the pre-merge verdict thread (`runAgentMaintenancePlanAndExecute`'s disposition input) and its unified-comment mirror — both already documented against the #4220 contradiction — continue to call the raw, uncached `fetchLivePullRequestMergeState` via `refreshLiveMergeState`, which is untouched by this diff. They must never be routed through the new cache: a base-conflicting PR reading a stale `clean` here while the disposition reads live `dirty` is exactly the bug #4220 fixed.

## Correctness verification

I adversarially verified the act-boundary claim with actual mutation testing, not just a read of the code: temporarily rerouted `refreshLiveMergeState` through the new cache, confirmed a **pre-existing, unrelated** test (`#audit-rate-headroom: the per-PR re-review refreshes merge state and CI...`) failed as expected, then confirmed the two purpose-built `#4220 regression` tests in this diff did **not** originally catch it — they were confounded by other same-request cache-priming/invalidation side effects that happened to clear the stale seed before the disposition ran, for unrelated reasons. Fixed both tests to no-op those confounding calls and assert directly (via pass-through spies) that the act-boundary's own read goes through the raw fetcher and never the cached one — then re-ran the same mutation to confirm both now fail correctly. Reverted the mutation; both pass cleanly on the real code.

Also added dedicated tests for `primeDurablePrStateCache` (previously exercised only indirectly, with no assertions on its own output), covering its early-exit, nullish-fallback, and PARTIAL-UPDATE CONTRACT branches.

## Known limitation

`invalidatePrStateCache` fires on `synchronize`/`closed`/`reopened` only (matching the issue's own scope), not on an `edited` action that retargets a PR's base branch — a rare case where GitHub could recompute `mergeable_state` for the new base without one of those three actions firing. Bounded by the 5-minute safety-net TTL; not a correctness issue for the act-boundary path (which never reads this cache). Left as a documented gap rather than expanding scope.

## Tests

- `test/unit/pr-detail-durable-cache.test.ts` (new): cache miss/hit/expiry for all three PR-state fields, fail-open on a write hiccup, `invalidatePrStateCache`, the write-through status-preservation regression, and `primeDurablePrStateCache`'s three branches.
- `test/unit/backfill-file-hydration-scoping.test.ts`: cold-cache fetch, nullish-state fallback, head-independent reuse (contrasted with the files cache's head-scoped behavior), and invalidated-cache refetch for reviews.
- `test/unit/queue.test.ts`: webhook invalidation for all 3 PR-state actions and all 3 review actions (`it.each`), non-invalidating actions left alone, sibling-event non-invalidation, and the two mutation-tested `#4220` regression tests described above.

## Validation

- `npm run typecheck`
- `npx vitest run test/unit/pr-detail-durable-cache.test.ts test/unit/backfill-file-hydration-scoping.test.ts test/unit/queue.test.ts`
- `npm run test:changed`
- `npm run test:coverage` (unsharded, full suite — 100% line and branch coverage on every changed line in `src/**` per `coverage/lcov.info`, cross-checked programmatically against the diff)
- `npm run test:workers`
- `npm run db:migrations:check`
- `npm run test:ci` (the full local gate, exit 0)
- `npm audit --audit-level=moderate`
- `git diff --check`
- Manual mutation testing of the act-boundary regression tests (see Correctness verification above)

## Scope

- [x] Change is narrow and limited to the stated problem
- [x] No secrets, wallets, hotkeys, trust scores, or reward values added anywhere
- [x] No edits to `site/`, `CNAME`, `**/lovable/**`, or `CHANGELOG.md`

## Safety

- [x] No auth/session/CORS surface touched
- [x] The act-boundary merge/close decision is unchanged and verified via mutation testing to still force-refetch live, uncached


JSONbored:
Fixed the CI failure: `migrations/0093_pull_request_detail_sync_pr_state.sql` collided with `migrations/0093_gate_require_fresh_rebase_window.sql` (added by #2616, merged after this branch was cut). Rebased onto current `main` and renumbered this migration to `0094` — `npm run db:migrations:check` now reports a clean, contiguous 0001..0094 with no new duplicates. Full local gate (`test:ci`, `npm audit`) green; no other changes.

JSONbored:
Fixed both flagged blockers:

**1. Cross-field cache-freshness bug (real defect).** `cachedFetchLivePullRequestMergeState` / `cachedFetchLivePullRequestState` / `cachedFetchLivePullRequestHeadSha` each wrote through only the one field they cared about, but all three share a single `prStateFetchedAt` stamp — so a live fetch from any one of them made the OTHER two fields look "fresh" to a later reader despite never having been fetched, silently returning `undefined` for a field that was simply never populated. Since all three narrow fetchers already hit the exact same `GET /pulls/{n}` endpoint (no extra API cost), a new `fetchAndCachePrStateFields` helper now fetches the full payload once per cache miss and writes `mergeable_state`/`state`/`headSha` through together — `headSha` still omitted (not nulled) when absent, preserving the existing PARTIAL-UPDATE CONTRACT so a prior value the files cache depends on is never cleared. It also only writes on an actual fetch success, so a transient failure no longer poisons the cache with a false "confirmed fresh" stamp. The three public, uncached narrow fetchers used by the act-boundary/gate-override callers are untouched. Added a direct regression test proving a fetch from one reader now warms the other two, plus updated/added tests for the corrected headSha behavior.

**2. Possible leaked secret (false positive, confirmed via the actual scanner).** `{ token: "installation-token" }` in six of this PR's own new test fixtures matched the deterministic `generic_secret_assignment` heuristic (a keyword-shaped pattern-match, not a real credential format — the value itself is an established, harmless test placeholder used hundreds of times elsewhere in this same file, just never previously part of a scanned diff). Ran the actual `scanForSecrets` function from `src/review/secrets-scan.ts` directly against the diff to confirm both that this was the *only* trigger and that renaming to `"fake-installation-token"` (recognized by the scanner's own placeholder-value allowlist) clears it — verified `{"found":false,"kinds":[]}` before pushing. No change to the shared scanner itself.

Also rebased twice more to keep pace with `main` (renumbered the migration 0093→0094, unrelated to this PR's own logic). Full local gate (`test:ci`, `npm audit`) green; 389 tests passing across the three affected test files.

JSONbored:
Fixed a third, separate defect from the latest review pass — a real backward-compatibility issue:

**Legacy reviews_synced_at markers from failed syncs.** Confirmed by diffing against \`origin/main\`: every pre-existing writer of \`reviews_synced_at\` (\`backfillOpenPullRequestDetails\` / \`refreshPullRequestDetails\` / \`backfillRepository\`) stamped it *unconditionally* on every sync pass, success or failure — there was no gate on it at all before this PR. This PR's new durable review cache treats *any* non-null \`reviews_synced_at\` as "fully synced, never refetch" until a \`pull_request_review\` webhook invalidates it. So an existing row whose review sync previously failed (rate limit, transient error) already carries a timestamp from that failed attempt, and the new cache would trust that incomplete/stale review data indefinitely for any PR that doesn't happen to get a new review action after this deploys.

Added migration \`0095\`: a one-time \`UPDATE\` clearing \`reviews_synced_at\` on every existing row. There's no reliable way to tell a trustworthy stamp from an untrustworthy one from the row alone (\`status\` reflects the aggregate files/reviews/checks outcome for a pass, not reviews specifically), so it clears unconditionally rather than guessing — the cost is one redundant review refetch per already-correctly-synced PR, which is cheap and bounded. Pinned the exact migration behavior with a dedicated test that applies the real migration file against a raw in-memory SQLite DB (mirroring the existing \`digest-subscription-normalization.test.ts\` pattern), seeding legacy-shaped rows and asserting only \`reviews_synced_at\` clears while every other column is untouched.

Rebased again to keep pace with \`main\` (two more unrelated merges landed). Full local gate (\`test:ci\`, \`npm audit\`, \`db:migrations:check\`) green.

JSONbored:
Superseded by #2633, which merged independently while this PR was still being iterated on.

#2633 delivers a hardened, complete implementation of the review-caching half of #2537 — including two things this PR's own fix history was still working through: making `fetchAndStorePullRequestDetails` report the real per-pass sync outcome instead of inferring it from warning strings (closing a TOCTOU race between the cache snapshot read and the final write, not just the "cache hit re-stamps the TTL" bug this PR caught), and a targeted legacy-data migration keyed on sync status rather than a blanket reset. It went through its own multiple gate-review rounds and is the better implementation for that half.

This PR's bare PR-state caching (mergeable_state/state/head-SHA, scoped to the non-mutating freshness-guard/readiness/dup-winner reads while keeping every act-boundary read — merge/close, draft-dodge, reopen-reclose, gate-override — live) is the half #2633 explicitly deferred as its own follow-up, and isn't touched by it. That work is tested and gate-verified in this branch's history but conflicts too deeply with #2633's now-merged review-cache implementation to rebase cleanly; closing here rather than force a risky reconciliation. Worth a fresh, narrowly-scoped PR against current main if that caching is still wanted — the design and tests already exist in this branch for reference.

PR #2668, opened 2026-07-03
feat(enrichment): flag risky workflow permissions and triggers

jason020818:
## What

A new local REES analyzer, `workflowPermissions`, that flags changes to `.github/workflows/*` which grant broad or sensitive permissions, or add high-risk event triggers — the `pull_request_target` + write-token supply-chain class, where untrusted fork code runs with the base repo's secrets and a write token.

## Detections (the fixed GitHub Actions schema — a bounded enum)

- `permissions: write-all`
- `id-token: write` (OIDC cloud-credential exchange)
- `secrets: inherit` (forwards all of the caller's secrets to a reusable workflow)
- sensitive per-scope writes — `contents` / `packages` / `actions` / `deployments` / `security-events` `: write` (reports the scope)
- `pull_request_target` and `workflow_run` triggers, matched only as a trigger *declaration* — an `on:` line naming it, a mapping key, or a `- ` list item — so the token merely appearing in a name/value is not flagged

## Why it is bounded / merge-safe

It matches the fixed GitHub Actions workflow schema (a documented enum of permission scopes and event names), never arbitrary code, so it has no unbounded parsing edge cases. YAML `#` comments are stripped before matching. It mirrors the existing `actions-pin` and `iac-misconfig` line-cited patch analyzers (per-line scan, hunk tracking, `MAX_FINDINGS`/`MAX_LINE_CHARS` caps, dedup by kind+line, fail-safe on abort).

## Value

No existing analyzer covers workflow permission escalation or dangerous event triggers. A PR that adds `pull_request_target` (running untrusted fork code with the base repo's secrets) or a broad `write-all` grant is a well-known supply-chain vector worth surfacing before merge.

## Tests

`review-enrichment/test/workflow-permissions.test.ts` covers every kind, each sensitive scope, YAML-comment stripping, read-grant and `actions/*` uses-ref non-matches, new-file line tracking across removed/context lines, per-kind dedup, the budget/abort paths, and the workflow-path filter. Analyzer metadata is regenerated and committed.

## No linked issue

No linked issue because this is a net-new analyzer; there is no tracking issue to link.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,17 +65,18 @@ GITTENSORY_REVIEW_ENRICHMENT=false
 # Current analyzer names:
 # dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol,redos
 # provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature,iacMisconfig,nativeBuild
-# history,docCommentDrift,duplication,churnHotspot,blameLink
+# history,docCommentDrift,duplication,churnHotspot,blameLink,workflowPermissions
 #
 # Profile defaults:
 # fast: dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol
-# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild
+# redos,provenance,secretLog,typosquat,iacMisconfig,nativeBuild,workflowPermissions
 # balanced (default): dependency,lockfileDrift,secret,license,installScript,heavyDependency
 # actionPin,eol,redos,provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature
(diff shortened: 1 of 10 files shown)


JSONbored (commented):
Closing, see gittensory review for blockers.

```
AI reviewers agree on a likely critical defect: review-enrichment/src/analyzers/workflow-permissions.ts:129 resets `OnBlockTracker` at every hunk and only enters the block after seeing `on:`, so adding `pull_request_target:` or `- workflow_run` under an existing `on:` block whose `on:` line is outside the diff context produces no finding even though the PR claims those trigger declarations are detected. — Resolve the flagged defect, or override if the AI reviewers are mistaken, then re-run the gate.
```

PR #2614, opened 2026-07-02
feat(miner): add laptop-mode init and doctor commands

Crystora:
## Summary

- Adds laptop-mode `gittensory-miner init` and `gittensory-miner doctor` commands for #2329.
- Initializes local SQLite state idempotently, including existing empty state files.
- Documents the laptop-mode install/init/doctor quickstart.

## Scope

- [x] The PR title follows `type(scope): short summary` Conventional Commit format, for example `fix(api): restore profile access checks`.
- [x] This PR is focused and does not mix unrelated backend, UI, MCP, docs, dependency, and deploy changes.
- [x] This follows `CONTRIBUTING.md` and does not reintroduce GitHub Pages, VitePress, `site/`, or `CNAME`.
- [x] I linked an issue, or this is small enough that the summary explains why an issue is not needed.

Fixes #2329

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally; **`codecov/patch` requires ≥99% coverage of the lines AND branches you changed** (aim for **100%** on your diff so CI variance does not fail near the threshold). Global coverage is a non-blocking trend with a loose 90% backstop, not the gate.
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:openapi:check`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`
- [x] `npm audit --audit-level=moderate`
- [x] New or changed behavior has unit/integration tests for new branches, fallback paths, and sanitizer boundaries

Also run:
- [x] `npx vitest run test/unit/miner-cli.test.ts`
- [x] `npm --workspace @jsonbored/gittensory-miner run build`
- [x] `npm run build:miner`
- [x] `npm run test:ci`

If any required check was skipped, explain why:

- None.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed.
- [x] Public GitHub text stays sanitized, low-noise, and does not imply compensation guarantees or optimization tactics.
- [x] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests.
- [x] API/OpenAPI/MCP behavior is updated and tested where needed.
- [x] UI changes use live API data or real empty/error/loading states, not production mock/demo fallbacks.
[...]

Issue #2288, opened 2026-07-01
feat(miner-foundation): slim laptop-mode CLI entry point (status/doctor commands)


JSONbored:
Build out the miner CLI's first real commands: `gittensory-miner status` (prints package version, resolved `@jsonbored/gittensory-engine` version via the earlier `ENGINE_VERSION` export, config-file discovery state, and local run-state DB path) and `gittensory-miner doctor` (checks Node version against `engines.node`, checks the engine package resolves, checks a writable local-state directory exists/can be created — mirrors `gittensory-mcp doctor`'s spirit of "is this laptop set up correctly" without any network calls). This is explicitly the "slim laptop-mode CLI entry point" the phase brief calls for — no repo-scanning, no coding-agent invocation, no GitHub writes; those are later phases.

## Deliverables
- [ ] `gittensory-miner status` — prints package + engine version, config path, local-state directory, as human-readable text by default and `--json` for machine consumption (mirror `gittensory-mcp version --json` / `config --json` flag convention).
- [ ] `gittensory-miner doctor` — checks: Node major version >= `engines.node` floor, `@jsonbored/gittensory-engine` resolves and its `ENGINE_VERSION` is readable, local-state directory is writable (create if absent). Exits non-zero with a clear message on any failure (mirror `gittensory-mcp doctor --exit-code`).
- [ ] `gittensory-miner --help` lists both commands with one-line descriptions.
- [ ] Unit tests for both commands' happy-path and at least one failure path each (e.g. `doctor` reporting an unwritable directory).
- [ ] No network calls anywhere in this issue's code — assert this with a `vi.stubGlobal("fetch", ...)` throw-on-call test the way the phase's "never phone home for core operation" requirement demands.

## References
- `packages/gittensory-mcp/bin/gittensory-mcp.js` — `doctor`/`status`/`version --json` command precedents (see the `CLI_COMMAND_SPEC` table, line 27-46, listing `status`, `doctor`, `version` as existing top-level commands with no subcommands).
- `packages/gittensory-miner/` — created by the package-skeleton issue (must land first).
- Phase brief: "build the slim laptop-mode CLI entry point."

PR #2688, opened 2026-07-03
feat(miner): add PreToolUse-style deny-hook primitives

### LINKED ISSUE #2295: feat(miner-foundation): PreToolUse-style deny-hook primitives for gittensory-miner
Pull forward (from the roadmap's later Phase 5) the hook-enforced house-rule primitives: a small, pure `evaluateDenyHooks(toolCall: { name: string; input: Record<string, unknown> }, rules: DenyRule[]): { allowed: boolean; reason?: string }` mechanism modeled on Claude Code's PreToolUse deny-hook shape, so that even in this foundation phase, the miner package has a checked, testable place to enforce house rules ("never touch `.github/workflows/**`", "never run `git push --force`", "never read/write `.env`") BEFORE any later phase wires an actual coding-agent tool-call loop through it. This issue implements the deterministic rule-evaluator only — there is no live tool-call interception yet (that needs a real coding-agent driver, out of scope for Phase 0), but the primitive and its rule format exist and are fully tested so later phases just plug a real event source into it.

## Deliverables
- [ ] A `deny-hooks` module exporting a `DenyRule` type (e.g. `{ matcher: string /* tool-name glob or exact */; pathPattern?: string /* glob against a path-shaped input field */; reason: string }`) and a pure `evaluateDenyHooks(toolCall, rules): { allowed: boolean; blockedBy?: DenyRule }`.
- [ ] A small built-in default rule set covering the repo's own house rules as a starting example: block any tool call whose input touches `.github/workflows/**`, `.env*`, or any path matching `**/secret*/**` or `**/*private*key*` (mirroring the forbidden-path patterns already enforced in `scripts/check-mcp-package.mjs`), and block any shell-command-shaped tool call containing `--force` combined with `push` (a conservative git-force-push guard).
- [ ] Unit tests: each built-in rule fires on a crafted matching tool call and does NOT fire on a clearly benign one (both sides of every rule, per the repo's own branch-coverage discipline); an empty rule set always allows; a tool call matching zero rules is allowed.
- [ ] Doc comment stating explicitly that this module NEVER executes or intercepts anything live in this phase — it is a pure decision function a later phase's real hook wiring will call.
- [ ] `gittensory-miner doctor` (from the CLI-entry-point issue) gains a line confirming the deny-hook module loads and its default rule set is non-empty, so "the safety primitive exists and is wired at least this far" is checkable from day one.

## References
- `scripts/check-mcp-package.mjs` (`forbiddenPath`/`forbiddenContent` regexes, lines 15-16) — existing house-rule-as-regex precedent in this exact repo to mirror for the built-in default rules.
- `src/settings/autonomy.ts` (`resolveAutonomy`, `isActingAutonomyLevel`) — THE reusable autonomy dial the phase brief points to; this deny-hook module is a complementary, lower-level primitive (blocks specific dangerous actions outright) rather than a replacement for the autonomy level system (which gates whether acting is allowed AT ALL) — cite both so a reviewer understands they compose rather than duplicate.
- Phase brief: "pull forward hook-enforced house-rule primitives early (PreToolUse-style deny hooks) since the roadmap explicitly says to pull these forward from Phase 5."
- New path: `packages/gittensory-miner/lib/deny-hooks.js`.

dhgoal:
## Summary

Adds the PreToolUse-style **deny-hook primitives** to `@jsonbored/gittensory-miner`: a pure, deterministic `evaluateDenyHooks(toolCall, rules)` decision function plus a non-empty built-in `DEFAULT_DENY_RULES` set, so the miner package has a checked, testable place to enforce house rules **before** any later phase wires a real coding-agent tool-call loop through it.

This is the deterministic rule-evaluator only — there is **no live tool-call interception** in this phase (that needs a real coding-agent driver, out of scope here). The module never executes, intercepts, or mutates anything; a later phase's real hook wiring calls the verdict.

A `DenyRule` is `{ matcher, pathPattern?, inputIncludesAll?, reason }`. A rule fires when its tool-name `matcher` matches **and** every constraint it declares matches: `pathPattern` (a glob) against any path-shaped string in the tool-call input, and/or `inputIncludesAll` (substrings that must all appear in one string field, e.g. a shell command). The built-in rules mirror the forbidden-path regex in `scripts/check-mcp-package.mjs` — block `.github/workflows/**`, `.env*`, `**/secret*/**`, `**/*private*key*` — plus a conservative git force-push guard (a command carrying both `push` and `--force`). An empty rule set, or a call matching no rule, always allows.

`inputIncludesAll` is a minimal, data-driven extension of the example `DenyRule` shape: the required force-push guard is a command-**content** check that a path glob cannot express, so it is modeled generally (order-independent substring set) rather than as a one-off special case.

Closes #2295.

### Follow-up (not in scope here)

The issue also lists wiring a status line into `gittensory-miner doctor`. That command does **not exist yet** — it comes from the CLI-entry-point issue (#2288), which is still open (and #2614 is an open PR for the laptop-mode doctor). Adding a `doctor` command here would duplicate/overlap that work, so the one-line "deny-hook module loads, default rule set non-empty" check is deferred to land alongside the CLI entry point. The primitive and its default set are complete and ready to wire; the non-empty invariant is already pinned by a test here.

## Scope

- [x] I linked an issue (#2295).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally — the new `test/unit/miner-deny-hooks.test.ts` passes; the whole suite is green. (This change lives entirely in `packages/**`, which Codecov does not measure, so it carries no `codecov/patch` obligation; the logic is nonetheless covered on both sides of every built-in rule.)
- [x] `node --check lib/deny-hooks.js` via `npm run --workspace @jsonbored/gittensory-miner build`
- [ ] `npm audit --audit-level=moderate` — this PR adds **no dependencies**, so dependency-review has nothing new to evaluate.
- [x] New behavior has unit tests for new branches (each built-in rule fires and does not fire; empty-rule-set and no-match allow paths; matcher/pathPattern/array-field/force-push-order composition; malformed-input degrade-to-allow).

- UI/OpenAPI/migration/workers checks are not applicable: this change is a single pure module in `packages/gittensory-miner/lib` plus its test — no `src/**`, UI, API schema, DB, or Cloudflare-binding surface is touched.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed. (The built-in rules reference secret-bearing **path patterns** as deny targets; no secret values appear.)
- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. — n/a: pure decision function, no auth/session/network surface.
- [ ] API/OpenAPI/MCP behavior is updated and tested where needed. — n/a: no API/OpenAPI/MCP surface changed.
- [ ] UI changes use live API data or real states. — n/a: no UI change.
- [ ] Visible UI changes include a `UI Evidence` section. — n/a: no visible UI, frontend, docs, or extension change.

## Notes

Additive: two new files (`lib/deny-hooks.js` + its `lib/deny-hooks.d.ts`) mirroring the package's existing hand-authored JS+`.d.ts` pattern (`lib/opportunity-fanout.*`), one line added to the package `build` (`node --check`) gate, and one new test file. No existing code is modified.

### DIFF
--- a/packages/gittensory-miner/lib/deny-hooks.d.ts
+++ b/packages/gittensory-miner/lib/deny-hooks.d.ts
@@ -1,0 +1,24 @@
+export type DenyRule = {
+  /** Tool-name glob (`*` = any within a segment, `**` across segments) or an exact tool name. */
+  matcher: string;
+  /** Optional glob tested against every path-shaped string in the tool-call input. */
+  pathPattern?: string;
+  /** Optional substrings that must ALL appear in one string-shaped input field (e.g. a shell command). */
+  inputIncludesAll?: string[];
+  /** Human-readable reason surfaced when this rule blocks a call. */
+  reason: string;
+};
+
+export type DenyVerdict = {
+  allowed: boolean;
+  blockedBy?: DenyRule;
+};
+
+export type ProposedToolCall = {
+  name: string;
+  input: Record<string, unknown>;
+};
+
+export const DEFAULT_DENY_RULES: DenyRule[];
+
+export function evaluateDenyHooks(toolCall: ProposedToolCall, rules?: DenyRule[]): DenyVerdict;
--- a/packages/gittensory-miner/lib/deny-hooks.js
+++ b/packages/gittensory-miner/lib/deny-hooks.js
@@ -1,0 +1,100 @@
+// PreToolUse-style deny-hook primitives (#2295). A pure, deterministic rule evaluator modeled on Claude Code's
+// PreToolUse deny-hook shape: given a proposed tool call and a set of deny rules, it decides allow/block WITHOUT
+// executing, intercepting, or mutating anything. There is NO live tool-call interception in this phase — a later
+// phase's real coding-agent driver plugs an event source into `evaluateDenyHooks`; this module is only the
+// decision function. No IO, no globals, no Date/random: identical inputs always yield the identical verdict.
+//
+// A rule fires when its tool-name `matcher` matches AND every constraint it declares also matches:
+//   - `pathPattern` (a glob) must match some path-shaped string in the tool-call input, and/or
+//   - `inputIncludesAll` (substrings) must ALL appear in a single string-shaped input field (e.g. a command).
+// A rule with neither constraint fires on the matcher alone. The built-in DEFAULT_DENY_RULES mirror the
+// forbidden-path patterns enforced in `scripts/check-mcp-package.mjs` plus a conservative git force-push guard.
+
+/**
+ * Compile a glob to an anchored RegExp. `**` matches across path segments (any char incl. `/`); a leading `**​/`
+ * also matches zero directories; `*` matches within a single segment (no `/`); every other char is literal. Kept
+ * intentionally small — it only needs the shapes the built-in rules use (`.github/workflows/**`, `**​/.env*`,
+ * `**​/secret*​/**`, `**​/*private*key*`).
+ */
+function globToRegExp(glob) {
+  let source = "";
+  for (let i = 0; i < glob.length; i += 1) {
+    const char = glob[i];
+    if (char === "*") {
+      if (glob[i + 1] === "*") {
+        i += 1;
+        if (glob[i + 1] === "/") {
+          i += 1;
+          source += "(?:.*/)?"; // '**/' — any (or zero) leading directories
+        } else {
+          source += ".*"; // '**' — any char, including '/'
+        }
+      } else {
+        source += "[^/]*"; // '*' — any char except '/'
+      }
+    } else {
+      source += char.replace(/[.+?^${}()|[\]\\]/g, "\\$&");
+    }
+  }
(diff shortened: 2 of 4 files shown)


JSONbored (commented):
Please fix this: 

```
The most notable gap is that path globs only evaluate whole top-level string values, so command strings containing protected paths are not covered unless a later caller tokenizes them first.
```

dhgoal:
Fixed. The evaluator now tokenizes each input string (whitespace-split, quotes stripped) and tests every path glob against the whole value **and** each token, so a protected path embedded as a command argument — e.g. `git add .github/workflows/ci.yml` — is caught without relying on a later caller to pre-split the command. Added a regression test covering an embedded workflow path, an embedded secret path, and a benign command that stays allowed.

Issue #2562, opened 2026-07-02
feat(agent-actions): add an install-wide contributor open-item cap across repos


JSONbored:
Parent: #1936

## Problem
`contributorOpenPrCap`/`contributorOpenIssueCap` live on the per-repo `repository_settings` row, and the enforcement query only lists open items scoped to the SAME repo. A self-hosted install that gates multiple repos (a common topology — e.g. one operator running 3 repos on one self-host instance) has no aggregate view: an actor spreading low-volume spam/farming PRs across several gated repos in the same install never trips any single repo's cap, even though their fleet-wide open-item count is high.

This is a simpler, more immediately buildable version of the already-tracked cross-*instance* federation work (#1970 — opt-in federated intelligence across separate self-hosted instances, which needs real cross-instance privacy/trust design). This is cross-REPO-*within-one-install*: no federation, no cross-instance privacy design needed — just an aggregate query against the same database every repo on that install already shares.

## Requirements
- Add an optional, install-level `globalContributorOpenItemCap` (env var or a maintainer-only settings row — not per-repo, since it aggregates ACROSS repos), off by default.
- When set, check the actor's total open PR/issue count across ALL repos this install gates, in addition to (not instead of) the existing per-repo cap, in the same short-circuit block as the existing `contributorCapMatch`.
- Must not require any new cross-instance networking/API — purely a same-database aggregate query.
- The existing `autoCloseExemptLogins` shared exemption list (already reusable across features per its own design) should apply here too.

## Deliverables
- The `globalContributorOpenItemCap` setting (install-scoped, not per-repo — needs its own config surface, likely an env var or a singleton settings row mirroring `global_contributor_blacklist`'s existing tenant-free-singleton pattern).
- An aggregate open-item-count query spanning all repos this install tracks.
- Wiring into the existing `contributorCapMatch` short-circuit in `src/settings/agent-actions.ts`, reusing `itemKind`/`closeKind: "contributor_cap"` rather than inventing a new kind.
- Tests: an actor over the install-wide cap but under every individual repo's cap is still caught; the existing per-repo cap continues to work unchanged; off-by-default confirmed.

## Acceptance criteria
- With `globalContributorOpenItemCap` configured, an actor whose SUM of open items across every repo on the install exceeds the cap is caught, even if no single repo's own cap is tripped.
- Off by default — zero behavior change for a single-repo install or one that hasn't opted in.
[...]

PR #2689, opened 2026-07-03
fix(agent-actions): live-verify + installation-scope the global contributor cap

JSONbored:
## Summary

Adversarial review of the already-merged #2678 (issue #2562) found two confirmed, empirically-verified bugs, and a third surfaced while porting the fix:

- **No live verification before an irreversible close**: `countOpenItemsForAuthorAcrossRepos` trusted the raw D1 cache directly. The sibling per-repo issue-cap path (`maybeCloseIssueOverContributorCap`) already live-verifies each sibling via `fetchLiveIssueState` before trusting it toward a close, specifically because the stored cache can lag GitHub. The install-wide path skipped this entirely.
- **itemKind mislabeling**: `contributorCapMatch.itemKind` was hardcoded to `"pull requests"` (PR call site) or `"issues"` (issue call site) even though the count sums PRs + issues together — a mixed-kind contributor got a close message with a wrong noun/count.
- **No installation scoping**: the aggregate query had no `installationId` filter — it summed across the ENTIRE D1 database regardless of which GitHub App installation each repo belongs to. A D1 instance serving more than one installation could wrongly count a contributor's unrelated-installation activity toward a close here.

Replaced `countOpenItemsForAuthorAcrossRepos` (a count) with `listOpenItemsForAuthorAcrossInstall` (rows, installation-scoped), added `verifiedGlobalOpenItemCount` which live-confirms every OTHER counted item (bounded worker-pool fan-out, concurrency 10, to avoid an unbounded burst against the installation's rate limit) before trusting it toward the cap, and widened `itemKind` to include `"pull requests and issues"` for the install-wide case.

Addresses gate-review findings on #2678.

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run test/unit/global-contributor-cap.test.ts test/unit/queue.test.ts` — 395/395 pass, including all 9 existing install-wide cap webhook-flow tests (updated to register repos under the calling installation, matching the new installation-scoping requirement) and 3 new/updated `listOpenItemsForAuthorAcrossInstall` tests covering cross-installation isolation and list-truncation observability

### DIFF
--- a/src/db/repositories.ts
+++ b/src/db/repositories.ts
@@ -3104,26 +3104,91 @@ export async function countOpenPullRequests(env: Env, fullName: string): Promise
   return Number(row?.count ?? 0);
 }
 
-/**
- * Install-wide open-item count for one author (#2562, anti-abuse): SUM of this author's open PRs + open
- * issues across EVERY repo tracked in this install's database -- deliberately NOT scoped by repoFullName,
- * unlike countOpenPullRequests/countOpenIssues above. This is what makes the globalContributorOpenItemCap
- * catch an actor spreading low-volume spam across several gated repos in the same self-hosted install: no
- * single repo's own cap trips, but the aggregate does. Same-database aggregate only -- no cross-instance
- * networking, mirroring the install-scoped singleton shape of global_contributor_blacklist. Case-insensitive
- * login match (mirrors loginMatches/findBlacklistEntry elsewhere in this file).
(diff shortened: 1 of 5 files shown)


JSONbored:
Superseded by a fresh PR from current main — this branch predates an installation-scoping fix that landed concurrently on main and conflicts here.

Issue #2563, opened 2026-07-02
feat(gate): add a lockfile-tamper-risk check


JSONbored:
Parent: #1936

## Problem
No check exists that a changed `package-lock.json` (or `review-enrichment/package-lock.json`, `apps/gittensory-ui/package-lock.json`) stays internally consistent with its `package.json`, or that lockfile entries weren't hand-edited to swap a resolved tarball URL/integrity hash for something outside the registry's normal resolution — a classic supply-chain vector. `npm ci` only catches STRUCTURAL mismatch locally for the person running it; nothing surfaces a maliciously-crafted-but-internally-consistent lockfile edit as a gate finding the way `secret_leak` does. This is real but speculative — no evidence of exploitation, lower priority than the other findings in this batch.

## Requirements
- Add an advisory-by-default (block opt-in, matching house style) `lockfile_tamper_risk` finding: flag any diff hunk inside a `package-lock.json`/`*.lock` file where a `resolved:`/`integrity:` value changed WITHOUT the corresponding `package.json` dependency version changing, or where a `resolved` URL points outside `registry.npmjs.org`.
- Config-driven gate dimension (`gate.lockfileIntegrity: off|advisory|block`), off by default.
- Distinct from #1474 (dependency-diff + OSV.dev CVE analyzer) — that covers KNOWN-CVE detection in changed deps, this covers lockfile-tamper/resolved-URL integrity, a different threat model.

## Deliverables
- A new gate rule producing a `lockfile_tamper_risk` finding when the conditions above are met.
- The `gate.lockfileIntegrity` setting wired through the config-as-code chain.
- Tests: a legitimate dependency bump (version + resolved + integrity all change together) does not trigger; a hand-edited `resolved`/`integrity` with no corresponding version bump does trigger; a `resolved` URL outside the npm registry triggers.

## Acceptance criteria
- A tampered lockfile entry (resolved/integrity changed without version bump, or non-registry URL) produces the advisory finding.
- A normal `npm install`-driven lockfile update does not false-positive.
- Off by default.

## Expected outcome
A previously entirely-silent supply-chain vector (hand-edited lockfile entries) gets a real, if currently low-priority, deterministic gate signal — available to enable for repos with elevated supply-chain concerns.

PR #2692, opened 2026-07-03
fix(review): stop false-positiving on transitive bumps and workspace links

JSONbored:
## Summary

Adversarial review of the already-merged #2676 (issue #2563) found two confirmed, empirically-reproduced false-positive bugs that directly violate the issue's own acceptance criterion ("a normal npm install-driven lockfile update does not false-positive"):

- **Transitive dependency bumps misclassified as tamper**: `packagesWithManifestVersionChange()` only scanned `package.json`'s own dependency blocks (direct dependencies). The vast majority of any real `package-lock.json`'s entries are transitive dependencies, which never appear in any `package.json` at all — so the check misfired on nearly every ordinary `npm install`/`npm update`.
- **npm workspace-local packages misclassified as off-registry**: `NPM_REGISTRY_HOST_RE` required an exact `https://registry.npmjs.org/` prefix. A workspace's own local packages (e.g. this repo's `packages/gittensory-mcp`, `apps/gittensory-ui`) have a `resolved` field that's a relative filesystem path, not a URL at all — it failed the registry-host check trivially and got flagged the moment that entry's resolved line changed (entirely routine).

**Fix**: replaced the package.json cross-reference with a self-contained signal — every package-lock.json entry (direct AND transitive) carries its own `version`/`resolved`/`integrity` trio, and a genuine install/update always bumps all three together. The tamper signal now compares a `resolved`/`integrity` change against that SAME entry's own `"version"` line within the same diff, which covers transitive and direct dependencies uniformly and is a more specific, self-contained tamper tell (a hand-edited resolved/integrity with the version left unchanged) than the old cross-file check. Also exempted non-`http(s)://` resolved values from the off-registry check.

Addresses gate-review findings on #2676.

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run test/unit/lockfile-tamper.test.ts test/unit/lockfile-tamper-wiring.test.ts test/unit/queue.test.ts` — 417/417 pass, including 3 new regression tests (transitive-only bump does not trigger, workspace-local resolved path does not trigger, a genuine off-registry URL still triggers) and one existing test updated to reflect the new, more accurate design (a changed integrity with no version bump is now correctly flagged even when the package is also being dropped from package.json — previously silently ignored)

### DIFF
--- a/src/review/lockfile-tamper.ts
+++ b/src/review/lockfile-tamper.ts
@@ -1,15 +1,31 @@
 // Lockfile-tamper-risk gate check (#2563). Deterministic scan of a changed `package-lock.json` (or another
-// `*.lock` file) diff for the classic supply-chain tell: a `resolved`/`integrity` value changed WITHOUT the
-// corresponding `package.json` dependency version changing, or a `resolved` URL that points outside the public
-// npm registry. Distinct from the OSV.dev CVE analyzer (review-enrichment/src/analyzers/lockfile-drift.ts) —
-// that flags KNOWN-CVE versions; this flags tamper/integrity-substitution regardless of whether the substituted
+// `*.lock` file) diff for the classic supply-chain tell: a `resolved`/`integrity` value changed WITHOUT that
+// SAME package-lock entry's own `"version"` field genuinely changing, or a `resolved` URL that points outside
+// the public npm registry. Distinct from the OSV.dev CVE analyzer (review-enrichment/src/analyzers/lockfile-drift.ts)
+// — that flags KNOWN-CVE versions; this flags tamper/integrity-substitution regardless of whether the substituted
 // version has a published CVE. Config-driven, off by default (see rules/advisory.ts isConfiguredGateBlocker +
 // signals/focus-manifest.ts gate.lockfileIntegrity) — this module only PRODUCES the finding; it never decides
 // whether the finding blocks.
+//
+// Why compare against the lockfile entry's OWN version rather than package.json (see #2563 gate-review
+// follow-up on #2676): every package-lock.json entry — direct AND transitive — carries its own version/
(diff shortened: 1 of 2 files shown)


JSONbored:
Both findings investigated:

1. **Nested-entry collision (blocker)** — confirmed real. `scanPackageLockPatch` keyed candidates by the bare package name, so two distinct package-lock.json entries for the same package at different `node_modules` nesting depths (e.g. `node_modules/foo` and `node_modules/bar/node_modules/foo`, which npm creates when dependents need incompatible versions) were merged into one shared tamper-tracking record. A legitimate bump on one entry could mask an unbumped, tampered `resolved`/`integrity` edit on the other. Fixed by keying the map on the full entry path instead of the bare name; the bare name is still used for display only. Added regression tests covering both processing orders.

2. **CI: codecov/patch 95% vs 99%** — the gap was a single unreachable branch in `versionChanged()`'s dead both-undefined guard (mathematically redundant: `removed !== added` already evaluates to `false` when both are `undefined`). Simplified the function; coverage is now 100% lines/branches/statements/functions on the touched file.

3. **Separate security-scanner P1 finding** (context lines treated as removed versions) — investigated and confirmed a **false positive**: `scanPackageLockPatch` already has an earlier unconditional guard in the same loop (`if (!currentEntryKey || !currentPackageName || line.sign === " ") continue;`) that filters out every context-sign line before the version/resolved/integrity match logic runs, so a context `"version"` line can never reach the `removedVersion` assignment. This predates the current PR (present since the original #2676). No code change was needed for this one, since the guard already makes it structurally unreachable — confirmed independently by tracing every `continue` in the loop and by checking `patchLines()` cannot yield any sign outside `"+"`/`"-"`/`" "`.

All 30 tests in `lockfile-tamper.test.ts` + `lockfile-tamper-wiring.test.ts` pass, typecheck is clean, coverage is 100%.

Issue #2567, opened 2026-07-02
feat(config): expose the dual-AI combine strategy as a per-repo .gittensory.yml setting


JSONbored:
Parent: #1936

## Problem
`ai-review.ts`'s dual-AI `CombineStrategy`/`onMerge`/`reviewers` config is a real, shipped capability that changes how the two independent AI opinions become one gate decision — but it's only settable via the self-host operator-level `env.AI_REVIEW_PLAN` boot config, not per-repo. There is no `review.combine`, `review.on_merge`, or `review.reviewers` field in the manifest schema, so an individual repo maintainer on a shared self-host instance cannot opt into synthesis (block on either reviewer) or pick a different confidence trade-off without an operator redeploying the whole instance's env config. This breaks the house pattern (`.gittensory.yml` > DB > safe default, config-driven per repo) every other AI-review knob (profile, inline_comments, path_instructions, instructions, exclude_paths, pre_merge_checks) already follows. Lower priority — the operator-level default is a reasonable v1, this is a refinement.

## Requirements
- Add `review.combine`/`review.onMerge`/`review.reviewers`-equivalent fields to `FocusManifestSettings`'s review config, allowing a per-repo override of the operator's `AI_REVIEW_PLAN` default.
- Per-repo override should be a REFINEMENT only, not a way to bypass an operator-imposed floor (e.g. an operator running a strict instance shouldn't have a repo able to loosen `onMerge` below the operator's own minimum, if the operator has set one) — worth a short design note on precedence before implementing.
- Config-driven, matching the existing `.gittensory.yml` > DB > safe-default precedence pattern.

## Deliverables
- The new manifest fields wired through `src/signals/focus-manifest.ts` and consumed by `src/services/ai-review.ts` in place of the unconditional `env.AI_REVIEW_PLAN` read.
- A short precedence design note (operator floor vs. per-repo override) resolved before/during implementation.
- Tests covering: a repo without an override inherits the operator default unchanged; a repo with a valid override uses it; an invalid/out-of-bounds override is rejected or clamped per the precedence design.

## Acceptance criteria
- A repo can set its own combine strategy via `.gittensory.yml`, overriding the operator's instance-wide default, within whatever precedence bounds are decided.
- A repo without an override sees zero behavior change.

## Expected outcome
The dual-AI combine strategy follows the same config-as-code precedence every other AI-review knob already follows, instead of being the one operator-only exception.

PR #2695, opened 2026-07-03
fix(review): close two gaps in the per-repo dual-AI combine override

JSONbored:
## Summary

Two confirmed bugs in the already-merged #2677 (issue #2567):

**1. The operator's onMerge floor could be bypassed via unclamped combine/reviewers.** Flagged by gittensory's own live AI review on the original PR: `resolveEffectiveAiReviewOnMerge` clamped `onMerge` alone, so an operator plan of `{ combine: "synthesis", onMerge: "either", reviewers: [A, B] }` (two reviewers, either one can flag a blocker) could be effectively neutered by a repo override of `{ combine: "single", reviewers: [A] }` — reviewer B's vote never happens, without ever touching `onMerge` directly.

Added `resolveEffectiveAiReviewPlan`, which clamps `combine`/`reviewers` together with `onMerge`: when the operator has set an `either` floor, a repo override that would reduce the effective reviewer count (a shorter `reviewers` list or a switch to `combine: "single"`) is clamped back to the operator's own values, while `onMerge` still resolves through the existing `resolveEffectiveAiReviewOnMerge`.

**2. The AI-review cache fingerprint didn't include the new per-repo overrides.** Found by an independent adversarial review: `aiReviewCacheInputFingerprint`'s `reviewerPlan` field only reflects `env.AI_REVIEW_PLAN` (the operator's boot config) — never `aiReviewCombine`/`aiReviewOnMerge`/`aiReviewReviewers`. Since the actual gate-blocking verdict is directly shaped by these, a same-head-SHA cache hit (a re-delivered webhook or the block-mode re-gate sweep — both normal operational triggers) could replay a stale verdict computed under the old plan after a maintainer changes the override on an already-open PR.

Added the three fields to `AiReviewCacheInput`/`aiReviewCacheInputFingerprint`, mirroring exactly how the precedent field `aiReviewCloseConfidence` is already handled there.

Addresses gate-review findings on #2677.

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run test/unit/ai-review.test.ts test/unit/ai-review-cache-input.test.ts test/unit/ai-review-cache.test.ts test/unit/ai-review-advisory.test.ts test/unit/focus-manifest.test.ts test/unit/queue.test.ts` — 732/732 pass, including 7 new tests covering `resolveEffectiveAiReviewPlan`'s clamp/no-clamp cases (no floor, reduced reviewer count, collapse to single, implicit default reviewer count, unaffected tightening, independent onMerge-only clamp) and 1 new test covering the cache fingerprint change for all three new fields


JSONbored:
Fixed the blocker: \`collapsesToSingleReviewer\` computed its combine value via \`repoOverride.combine ?? operatorPlan?.combine\`, so an operator plan that itself sets \`combine: "single"\` (with no repo override at all) fell through to that same value and reported \`clamped: true\` on every single call, incrementing the clamp metric for no reason — there was nothing for a repo to have bypassed. Now requires \`repoOverride.combine === "single"\` directly, i.e. an actual repo-authored override.

Added a regression test for the operator's-own-single-combine case (asserts \`clamped: false\`) and a companion test confirming the reviewer-count clamp still correctly fires when a repo *does* try to shrink the reviewer list under that same operator plan. Verified by temporarily reverting the fix and confirming the new test fails exactly as expected, then restoring it.

Issue #2133, opened 2026-07-01
fix(agent-actions): owner-exemption logic does not recognize ADMIN_GITHUB_LOGINS


JSONbored:
Parent: #1936

## Problem

The main planner's close-eligibility gate (`src/settings/agent-actions.ts:315-320`) computes `isContributor = !authorIsOwner && !authorIsAutomationBot`, where `authorIsOwner` (`src/queue/processors.ts:1635-1637`, duplicated at `3359-3364`) is a literal case-insensitive match against the repo-owner segment of `repoFullName`. This exemption set does **not** include the fleet-operator `ADMIN_GITHUB_LOGINS` allowlist — which **is** honored as a maintainer-trust signal in the separate reopen-reclose path (`hasMaintainerPermission`, `processors.ts:7006-7008`: `login === repoOwner || admins.has(login)`), and was explicitly unified across the codebase in a prior PR specifically so it would be "honored everywhere." It wasn't threaded into the primary close-eligibility computation.

**Failure scenario:** an operator/admin login (present in `ADMIN_GITHUB_LOGINS` but not the literal repo-owner account) opens a PR on a gate-managed repo. If that PR trips a heuristic close condition (red CI, gate blocker, base conflict) or an erroneous blacklist match, the planner will close it — even though the same login is treated as a trusted, reopen-immune maintainer elsewhere in the same codebase. A genuine, low-likelihood but real cross-path authorization inconsistency.

## Requirements

- The same "trusted operator" identity must be honored consistently across every close-eligibility check in the codebase, not just the reopen-reclose path.

## Deliverables

- Extend the planner's input with an `authorIsAdmin` (or fold into a broader `authorIsMaintainer`) field, computed using the same `parseGitHubLoginList(env.ADMIN_GITHUB_LOGINS)` helper already used by `hasMaintainerPermission`.
- Thread it into `isContributor`/`closeEligible`/`blacklistContributor` in `agent-actions.ts`, and apply the identical fix to the duplicate `authorIsOwner` computation in the draft-dodge path (`processors.ts:3359-3369`).
- Add a regression test in `test/unit/agent-actions.test.ts` asserting an admin-login (non-owner) author is not close-eligible under failing CI / a conflict / a blacklist match, mirroring the existing `authorIsOwner` tests.

## Acceptance criteria

[...]

PR #2723, opened 2026-07-03
fix(agent-queue): re-gate a staged close/merge on the admin (#2133) close exemption

glorydavid03023:
## What
The accept-time re-checks in `agent-approval-queue.ts` re-derive `closeEligible` against the **current** settings so a staged close/merge doesn't act on a stale `closeOwnerAuthors` toggle (the head-SHA pin can't catch a settings flip). But **both** copies (the close re-check and the merge re-check) computed:

```ts
const closeEligible = (!authorIsOwner && !authorIsAutomationBot) || (authorIsOwner && settings.closeOwnerAuthors === true);
```

— omitting `authorIsAdmin` entirely.

## Why it matters
`#2133` folds the fleet-operator allowlist (`ADMIN_GITHUB_LOGINS`) into the **same never-auto-closed exemption** as the repo owner — see `settings/agent-actions.ts` (`isContributor = !owner && !admin && !bot`; `closeEligible = isContributor || ((owner || admin) && closeOwnerAuthors)`) and `queue/processors.ts`. The approval-queue re-checks never mirrored it, so a **non-owner admin** fell into the first clause and was treated as an ordinary contributor — *always* close-eligible.

**Observable consequence:** an admin PR staged for a `linked-issue-hard-rule` close while `closeOwnerAuthors` was **on** would still auto-close after the operator turned it **off** — the owner is correctly superseded by the re-check, but the admin was not (the re-check's whole purpose is to catch exactly that toggle flip). Same gap on the merge-re-check's `closeEligible`.

## Fix
Mirror the primary derivation in both re-checks:
```ts
const authorIsAdmin = authorLogin.length > 0 && parseGitHubLoginList(env.ADMIN_GITHUB_LOGINS).has(authorLogin.toLowerCase());
const closeEligible = (!authorIsOwner && !authorIsAdmin && !authorIsAutomationBot) || ((authorIsOwner || authorIsAdmin) && settings.closeOwnerAuthors === true);
```

## Tests
Adds admin regressions mirroring the existing owner ones: (1) accept **supersedes** a staged close for a fleet-admin when `closeOwnerAuthors` is turned off before accept (fails on the old code); (2) accept **still executes** the close when it's still on; (3) the merge-re-check **supersedes** an admin's staged merge when the linked-issue rule is now violated and admin-close is enabled. All branches on the changed lines covered; full approval-queue suite green (59 tests).

### DIFF
--- a/src/services/agent-approval-queue.ts
+++ b/src/services/agent-approval-queue.ts
@@ -5,6 +5,7 @@ import { loadLinkedIssueHardRules, resolveLinkedIssueHardRule } from "../review/
 import { executeAgentMaintenanceActions, pendingActionToPlanned } from "./agent-action-executor";
 import { downgradeCloseToHold, downgradeMergeToHold, isProtectedAutomationAuthor, type PlannedAgentAction } from "../settings/agent-actions";
 import { findBlacklistEntry } from "../settings/contributor-blacklist";
+import { parseGitHubLoginList } from "../auth/security";
 import { isCloseHoldOnly, isHoldOnly } from "../review/outcomes-wire";
 import { fetchLiveCiAggregate, fetchLivePullRequestMergeState, fetchLivePullRequestReviewDecision } from "../github/backfill";
 import { githubRateLimitAdmissionKeyForToken } from "../github/client";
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing: 

```
The merge-side production change is consistent, but its new test does not actually distinguish old from new behavior because closeOwnerAuthors is true there and the old contributor branch was already close-eligible.
```

PR #2725, opened 2026-07-03
feat(enrichment): flag PRs significantly behind the default branch

jason020818:
## What

A new local REES analyzer, `staleBranch`, that flags a PR whose head is significantly behind the repo's current
default branch — a staleness risk GitHub's own PR page doesn't summarize as a number (it only shows an
"out-of-date" badge, not how far).

## Detection (structured GitHub repo + compare API fields only — a bounded, documented schema)

- Fetches the repo's current `default_branch`, then compares it against the PR's head via the GitHub compare API.
- Flags when `behind_by >= 100` commits.

## Why it is bounded / merge-safe

It reads only documented fields from the GitHub repo API (`default_branch`) and the compare API (`behind_by`) and
compares a single number against a fixed threshold — never diff, commit, or file content. There is no
text/YAML/code parsing, so there are no ambiguous-syntax edge cases. Fail-safe throughout: a missing token,
missing head SHA, malformed repo slug, a missing/non-numeric `behind_by`, or either fetch failing all yield no
finding rather than an error (the repo-info fetch failing short-circuits before ever calling compare). Mirrors
the existing `blame-link` / `approval-integrity` / `ci-check-signals` analyzers' structure (same header/fetch
helper shape, same `ScanOptions`/fail-safe conventions, a pure reducer separated from the network entrypoint).

## Value

A branch far behind the current default branch is more likely to hide a subtle semantic conflict that a clean
`mergeable` check alone would miss — the merge itself can be mechanically clean while the combined result is
still wrong. Surfacing how far behind (not just "out of date") gives a reviewer a concrete signal for whether a
rebase is worth requesting before merge.

## Tests

`review-enrichment/test/stale-branch.test.ts` covers: the pure `evaluateStaleBranch` reduction (the threshold's
boundary in both directions, the identical/ahead cases, and fail-closed on a missing/non-numeric/negative
`behind_by`), plus the network entrypoint's fail-safe paths (no token, no head SHA, malformed repo slug, the
repo-info fetch failing without ever calling compare, a repo response missing `default_branch`, the compare fetch
failing) and the exact two-request URL shape. Analyzer metadata is regenerated and committed.

## No linked issue

No linked issue because this is a net-new analyzer; there is no tracking issue to link.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,6 +66,7 @@ GITTENSORY_REVIEW_ENRICHMENT=false
 # dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol,redos
 # provenance,codeowners,secretLog,assetWeight,typosquat,commitSignature,iacMisconfig,nativeBuild
 # history,docCommentDrift,duplication,churnHotspot,blameLink,approvalIntegrity,ciCheckSignals
+# staleBranch
 #
 # Profile defaults:
 # fast: dependency,lockfileDrift,secret,license,installScript,heavyDependency,actionPin,eol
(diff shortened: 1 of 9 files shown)


JSONbored (commented):
Not something we need, closing.

PR #2754, opened 2026-07-03
feat(miner): add append-only event ledger

### LINKED ISSUE #2290: feat(miner-foundation): immutable event ledger for gittensory-miner
Add an append-only, immutable local event ledger (a second small SQLite table, or a separate DB file) recording every significant miner-loop event (`discovered_issue`, `plan_built`, `plan_step_completed`, `pr_prepared`, etc. — a small fixed event-type enum for this foundation phase; the full discover/analyze/plan/prepare loop's actual event vocabulary grows in later phases) with a monotonic sequence number and timestamp. This is the audit trail the phase brief calls for — append-only means no `UPDATE`/`DELETE` statements are ever issued against this table, only `INSERT`, so a contributor auditing the miner's history later can trust it was not retroactively edited.

## Deliverables
- [ ] An `event-ledger` module exposing `appendEvent(event: { type: string; repoFullName?: string; payload: Record<string, unknown> }): LedgerEntry` and `readEvents(filter?: { repoFullName?: string; since?: number })`.
- [ ] Schema: `id INTEGER PRIMARY KEY AUTOINCREMENT, seq INTEGER NOT NULL, event_type TEXT NOT NULL, repo_full_name TEXT, payload_json TEXT NOT NULL, created_at TEXT NOT NULL` — `seq` is a monotonically increasing counter maintained by the module (not just relying on `AUTOINCREMENT`'s reuse-after-vacuum behavior) so consumers get a stable ordering guarantee.
- [ ] A test asserting the module NEVER issues an `UPDATE` or `DELETE` against the ledger table — e.g. grep the module source in a unit test for those SQL keywords, or (stronger) wrap the DB handle in a test double that throws if either is called.
- [ ] Unit tests: append-then-read round-trip, sequence monotonicity across multiple appends, filter-by-`repoFullName`, filter-by-`since`.
- [ ] Doc comment stating the immutability invariant explicitly (mirroring `src/signals/duplicate-winner.ts`'s style of stating its own PURE invariant in a header comment) so a future contributor editing this file sees the constraint before touching it.

## References
- `migrations/0010_agent_orchestrator.sql` (55 lines) — the existing `agent_actions`/`agent_context_snapshots` append-oriented schema shape (created_at defaults, indexed by run/time) to mirror the SPIRIT of, though this is a local SQLite table, not a D1 migration.
- `src/signals/duplicate-winner.ts` (header comment, lines 1-13) — precedent for stating a hard invariant (there: PURE/no-IO; here: append-only) directly in the module's doc comment.
- New path: `packages/gittensory-miner/lib/event-ledger.js` (or co-located with `run-state`, per the run-state issue's file layout).

dhgoal:
## Summary

Adds the local **append-only event ledger** to `@jsonbored/gittensory-miner` — an immutable audit trail of every significant miner-loop event (`discovered_issue`, `plan_built`, `plan_step_completed`, `pr_prepared`, … — a small fixed vocabulary for this foundation phase that grows later), backed by a small local SQLite table. It mirrors the package's existing local-store pattern (`lib/run-state.js`, `lib/portfolio-queue.js`) and never uploads, syncs, or phones home — the DB is owner-only and stays on the miner's machine.

**Immutability invariant:** the module only ever issues `INSERT` and `SELECT` — it never rewrites or removes a row — so a contributor auditing the miner's history later can trust it was not retroactively edited. The header comment states this constraint explicitly, and a test asserts the module source contains no mutating/removing SQL keyword.

API (`lib/event-ledger.js`):
- `appendEvent({ type, repoFullName?, payload })` → `LedgerEntry` — stamps a module-maintained monotonic `seq` and an ISO timestamp; `payload` is stored as JSON and round-trips verbatim.
- `readEvents(filter?)` — all events in `seq ASC` order, optionally filtered by `repoFullName` and/or `since` (events with a seq **strictly greater** than `since` — the "give me everything after the last seq I saw" polling shape).

Schema is per the issue spec: `id INTEGER PRIMARY KEY AUTOINCREMENT, seq INTEGER NOT NULL, event_type TEXT NOT NULL, repo_full_name TEXT, payload_json TEXT NOT NULL, created_at TEXT NOT NULL`. `seq` is maintained by the module (`MAX(seq) + 1`) rather than relying on `AUTOINCREMENT`'s reuse-after-vacuum behavior, so consumers get a stable ordering guarantee.

Closes #2290.

## Scope

- [x] I linked an issue (#2290).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally — the new `test/unit/miner-event-ledger.test.ts` passes (whole miner suite green). This change lives entirely in `packages/**`, which Codecov does not measure, so it carries no `codecov/patch` obligation; the logic is nonetheless exercised across append/read round-trip, seq monotonicity, both filters (and their combination), null vs present repo scope, input rejection, and the append-only-source assertion.
- [x] `node --check lib/event-ledger.js` via `npm run --workspace @jsonbored/gittensory-miner build`
- [ ] `npm audit --audit-level=moderate` — this PR adds **no dependencies**, so dependency-review has nothing new to evaluate.
- [x] New behavior has unit tests for new branches, fallback paths, and the immutability invariant.

- UI/OpenAPI/migration/workers checks are not applicable: this change is one local-persistence module in `packages/gittensory-miner/lib` plus its test — no `src/**`, UI, API schema, Drizzle, or Cloudflare-binding surface is touched. The SQLite table is a miner-local file, not a repo migration.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed. The DB is created `0o600` in a `0o700` dir, owner-only, and never leaves the machine.
- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. — n/a: local-only SQLite persistence, no auth/session/network surface.
- [ ] API/OpenAPI/MCP behavior is updated and tested where needed. — n/a: no API/OpenAPI/MCP surface changed.
- [ ] UI changes use live API data or real states. — n/a: no UI change.
- [ ] Visible UI changes include a `UI Evidence` section. — n/a: no visible UI, frontend, docs, or extension change.

## Notes

Additive and consistent with the package's existing local-store pattern: two new files (`lib/event-ledger.js` + its `lib/event-ledger.d.ts`) mirroring `lib/run-state.js` / `lib/portfolio-queue.js` (path resolution, `0o600`/`0o700` perms, prepared statements, default-store singleton), one line added to the package `build` (`node --check`) gate, and one new test file. No existing code is modified.

### DIFF
--- a/packages/gittensory-miner/lib/event-ledger.d.ts
+++ b/packages/gittensory-miner/lib/event-ledger.d.ts
@@ -1,0 +1,36 @@
+export type LedgerEntry = {
+  id: number;
+  seq: number;
+  type: string;
+  repoFullName: string | null;
+  payload: Record<string, unknown>;
+  createdAt: string;
+};
+
+export type AppendEventInput = {
+  type: string;
+  repoFullName?: string;
+  payload: Record<string, unknown>;
+};
+
+export type ReadEventsFilter = {
+  repoFullName?: string;
+  since?: number;
+};
+
+export type EventLedger = {
+  dbPath: string;
+  appendEvent(event: AppendEventInput): LedgerEntry;
+  readEvents(filter?: ReadEventsFilter): LedgerEntry[];
+  close(): void;
+};
+
+export function resolveEventLedgerDbPath(env?: Record<string, string | undefined>): string;
+
+export function initEventLedger(dbPath?: string): EventLedger;
+
+export function appendEvent(event: AppendEventInput): LedgerEntry;
+
+export function readEvents(filter?: ReadEventsFilter): LedgerEntry[];
+
+export function closeDefaultEventLedger(): void;
--- a/packages/gittensory-miner/lib/event-ledger.js
+++ b/packages/gittensory-miner/lib/event-ledger.js
@@ -1,0 +1,195 @@
+import { chmodSync, mkdirSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+import { DatabaseSync } from "node:sqlite";
+import { isDeepStrictEqual } from "node:util";
+
+// The miner's local, append-only event ledger (#2290): an immutable audit trail of every significant miner-loop
+// event (discovered_issue, plan_built, plan_step_completed, pr_prepared, … — a small fixed vocabulary for this
+// foundation phase that grows in later phases), each stamped with a module-maintained monotonic `seq` and a
+// timestamp. IMMUTABILITY INVARIANT: this module only ever issues INSERT and SELECT — it NEVER rewrites or removes
+// a row, so a contributor auditing the miner's history later can trust it was not retroactively edited. Keep it
+// that way: do not add any statement that mutates or removes an existing row. The database is 100% local; this
+// module never uploads, syncs, or phones home with its contents. Mirrors the local-store pattern of run-state.js.
+
+const defaultDbFileName = "event-ledger.sqlite3";
+let defaultEventLedger = null;
+
+export function resolveEventLedgerDbPath(env = process.env) {
+  const explicitPath = typeof env.GITTENSORY_MINER_EVENT_LEDGER_DB === "string"
+    ? env.GITTENSORY_MINER_EVENT_LEDGER_DB.trim()
+    : "";
+  if (explicitPath) return explicitPath;
+
+  const explicitConfigDir = typeof env.GITTENSORY_MINER_CONFIG_DIR === "string"
+    ? env.GITTENSORY_MINER_CONFIG_DIR.trim()
+    : "";
+  if (explicitConfigDir) return join(explicitConfigDir, defaultDbFileName);
+
+  const configHome = typeof env.XDG_CONFIG_HOME === "string" && env.XDG_CONFIG_HOME.trim()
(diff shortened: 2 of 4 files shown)


dhgoal:
Fixed the flagged concurrency defect. `seq` allocation is now atomic: the schema enforces `UNIQUE(seq)` and `appendEvent` runs the `MAX(seq)+1` read and the INSERT inside a single `BEGIN IMMEDIATE` transaction (with `PRAGMA busy_timeout`), so two ledger instances opened on the same file serialize on the write lock and can never compute or persist a duplicate `seq`. Added a test appending 50 events and asserting the seqs are exactly 1..50 (gapless + unique).

dhgoal:
Fixed the payload round-trip defect. `appendEvent` now enforces that the payload survives JSON verbatim: it rejects (`invalid_payload`) any object JSON would alter — `undefined`/function/symbol values, `NaN`/`Infinity`, BigInt, or a circular reference — by verifying `isDeepStrictEqual(JSON.parse(JSON.stringify(payload)), payload)`. So a read-back is always identical to the appended event. Added tests for the rejected cases plus a nested JSON-safe payload that reads back identically.

Issue #2550, opened 2026-07-02
feat(gate): re-check migration-file collisions against live main before merge


JSONbored:
Parent: #1936

## Problem
`scripts/check-migrations.mjs` (wired into CI via `npm run db:migrations:check`) validates migration-number contiguity/duplicates by reading `migrations/` off disk in the checked-out branch. CI checks out `github.event.pull_request.head.sha` — the contributor's own branch tip — not a fresh merge against the current tip of `main`. Two PRs can each independently add a new, non-overlapping migration file (e.g. `0090_a.sql` and `0090_b.sql`); each individually passes `check-migrations.mjs` (no duplicate visible in its own branch), each gets `mergeable_state: clean` from GitHub (two distinct file additions, no textual conflict), and the gate's own live pre-merge recheck (`refreshLiveMergeState`/`refreshLiveCiAggregate` in `runAgentMaintenancePlanAndExecute`, `src/queue/processors.ts`) only re-fetches GitHub's mergeable_state and CI-check status — neither is content-aware, so neither catches this.

This is not hypothetical: it has now happened 4 times in this repo's history (0015/0017, 0074, 0075/0076, and most recently 0090 — PR #2479 and PR #2527 each independently claimed migration number 0090, merged cleanly, and required a reactive one-off patch, PR #2532, to grandfather the duplicate). No systemic fix exists — only the reactive patch.

## Requirements
- Add a live, content-aware recheck step to the gate's pre-merge pass (not just the CI/review-gate pass) for any PR touching `migrations/**`: re-run migration-collision-detection logic against a FRESH merge of the PR's head with the current live tip of `main`, immediately before the actual merge call.
- If a live collision is detected (that wasn't visible at CI time because it originated from another PR merged since), downgrade the disposition to a HELD state with a clear "rebase needed — a sibling migration landed on main since your last CI run" comment, never merge blind.
- Config-driven, off-by-default per the house pattern (e.g. `gate.premergeContentRecheck`), gated by the same autonomy/dry-run/kill-switch stack every other agent-driven mutation already uses.
- Must not add meaningful latency to the common case (no `migrations/**` changes in the PR ⇒ skip entirely).

## Deliverables
- A new pre-merge check function reusing (not duplicating) `scripts/check-migrations.mjs`'s duplicate-detection logic, invoked from `runAgentMaintenancePlanAndExecute` immediately before the merge action executes.
[...]

PR #2763, opened 2026-07-03
fix(gate): recheck migration collisions at merge

JSONbored:
### Motivation
- The live pre-merge migration collision check ran only during planning, leaving a TOCTOU window where a sibling PR could land a colliding migration before the staged/approved merge executed.
- The intent is to ensure the same fresh-tree collision logic used in CI/planning is enforced at the final actuation boundary (including approval-queue replay) to prevent landing duplicate migration numbers.

### Description
- Extracted the filename-extraction and live-collision logic into a shared helper `src/services/migration-collision-recheck.ts` with `migrationFilenamesForLiveRecheck` and `resolveLiveMigrationCollisionHold` so planning and actuation use identical logic.
- Updated the planner (`src/queue/processors.ts`) to call the shared helper instead of its private copy of the logic so the two paths stay consistent.
- Added an actuation-time recheck in the executor (`src/services/agent-action-executor.ts`) that, for `merge` actions, refreshes the PR's changed files and the base tree and re-runs the live collision check; when a collision is found the executor applies the existing `gittensory:migration-collision` label + comment and denies the merge instead of executing a stale plan.
- Added unit tests: `test/unit/migration-collision-recheck.test.ts` for helper behavior and a targeted regression in `test/unit/agent-approval-queue.test.ts` proving a staged merge is denied when a sibling migration lands before approval, plus a fast-path test that skips the recheck when the PR has no migration files.

### Testing
- Ran the focused unit suites covering the change: `npx vitest run test/unit/agent-approval-queue.test.ts test/unit/migration-collision-recheck.test.ts` and `npx vitest run test/unit/agent-approval-queue.test.ts test/unit/agent-action-executor.test.ts test/unit/migration-collisions.test.ts test/unit/migration-tree.test.ts test/unit/migration-collision-recheck.test.ts` and they passed locally (all related tests green). 
- Verified `git diff --check` (no whitespace/conflict markers) and the modified tests exercise collision/no-collision and fail-open behavior.
- `npm run typecheck` could not be completed cleanly due to pre-existing unrelated test-fixture type errors in `test/unit/queue.test.ts` (missing `securityFocus` fixture field), not caused by this change.
- `npm audit --audit-level=moderate` and the full `npm run test:ci` were blocked by network/registry constraints in the environment (`npm audit` returned a `403 Forbidden` and `actionlint` setup hit network resolution issues), so those checks were not completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4749ebfb54832cbcfba81206c892e8)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -247,8 +247,7 @@ import {
 } from "../settings/agent-actions";
 import { isAutoCloseExempt } from "../settings/auto-close-exempt";
 import { resolveGlobalContributorOpenItemCap } from "../settings/global-contributor-cap";
-import { detectMigrationCollisions, extractMigrationNumber, KNOWN_MIGRATION_DUPLICATES } from "../db/migration-collisions";
-import { listMigrationFilenamesAtRef } from "../github/migration-tree";
+import { migrationFilenamesForLiveRecheck, resolveLiveMigrationCollisionHold } from "../services/migration-collision-recheck";
 import {
   executeAgentMaintenanceActions,
   executeIssueMaintenanceActions,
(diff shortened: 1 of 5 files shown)


JSONbored:
Closing as superseded: this exact feature (live migrations/** collision recheck at merge, #2550) already shipped on `main` via #2585, and the specific bug this PR targets (a stale cached live-tree snapshot defeating the recheck) was already fixed there too, in `dce0fad5` ("fix(gate): remove the live migration-tree cache — it could serve a stale pre-merge snapshot"). Rebasing this branch and merging it as-is duplicates the live-tree fetch at a second call site (actuation time, on top of main's existing planning-time check), which is a real regression, not a fix — confirmed locally by `queue.test.ts`'s `treeCalls` assertion going from 2 to 3 after reconciling the rebase conflict. No further action needed here.

PR #3281, opened 2026-07-05
fix(observability): report actual orb ai token usage on the dashboards

JSONbored:
## Summary

- The "Gittensory - Codex usage (self-host)" dashboard's durable SQL panels (as flagged) filtered on `model LIKE '%codex%'` and surfaced only `estimated_neurons`, even though migration `0109_ai_usage_actual_tokens.sql` added real `provider`/`effort`/`input_tokens`/`output_tokens`/`total_tokens`/`cost_usd` columns to `ai_usage_events` — and #3270 wired all four live AI features to actually populate them.
- Rewrote every durable panel in `codex-usage.json` to filter on `provider = 'codex'` instead of pattern-matching the model string.
- Replaced the "Estimated neuron usage by day" panel with a real "Durable token usage by day" panel (input/output tokens) and added a new "Durable cost by day" panel; the recent-events table now includes `effort`/`input_tokens`/`output_tokens`/`total_tokens`/`cost_usd` alongside the existing `estimated_neurons` (kept for reference, not as the only metric).
- Added `grafana/dashboards/orb-ai-usage.json`: a provider-neutral dashboard over `ai_usage_events` covering every configured provider (Codex, Claude Code, OpenAI-compatible, Ollama, Anthropic BYOK), broken down by provider/model/effort/repo/PR, plus an "events missing real usage" counter to track which features still haven't threaded real usage through `recordAiUsageEvent`.
- Clarified `claude-usage.json`'s description: it's Claude Code's own OTEL/session telemetry (a session-scoped cumulative counter via `last_over_time`), not the durable `ai_usage_events` DB attribution — cross-references the new `orb-ai-usage` dashboard for that.
- All three dashboard JSON files are auto-provisioned from `grafana/dashboards/` (directory-mounted in `docker-compose.yml:424`, no config wiring needed for the new file) and validated by `scripts/validate-observability-configs.mjs`. `scripts/export-grafana-reporting-db.sh` already exports every column these panels reference (verified — no changes needed there).

## Scope

- [x] No issue linked — direct follow-up to #3270 with clear scope, per the observability gap identified there.

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint` (via `npm run test:ci`)
- [x] `node scripts/validate-observability-configs.mjs` — dashboards and alert rules valid
- [x] `npm run typecheck`
- [x] `npm run test:coverage` — no `src/**` changes in this PR (JSON config only); full suite green (8807 passed), including `test/unit/selfhost-grafana-dashboard.test.ts` and `test/unit/selfhost-grafana-reporting.test.ts`
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`
- [x] Ran the full `npm run test:ci` gate locally — green

## Safety

- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. — N/A
- [ ] API/OpenAPI/MCP behavior is updated and tested where needed. — N/A, no API/OpenAPI/MCP changes; Grafana dashboard JSON only.
- [ ] UI changes use live API data or real empty/error/loading states. — N/A, not a `gittensory-ui` change (Grafana operator dashboards, not the app UI).
- [ ] Visible UI changes include a `UI Evidence` section. — N/A per the template's own scope (operator-only Grafana dashboards, not `apps/gittensory-ui`); did not stand up a Grafana instance to screenshot since no live datasource is available in this environment, but every panel query was validated against the actual `ai_usage_events`/reporting-DB schema and the repo's own dashboard/alert-rule validator.
- [ ] Public docs/changelogs are updated where needed. — N/A, no changelog edit.

### DIFF
--- a/grafana/dashboards/claude-usage.json
+++ b/grafana/dashboards/claude-usage.json
@@ -4,10 +4,10 @@
   "tags": ["gittensory", "claude", "ai"],
   "timezone": "browser",
   "schemaVersion": 39,
-  "version": 3,
+  "version": 4,
   "refresh": "30s",
   "time": { "from": "now-7d", "to": "now" },
-  "description": "Claude Code review-CLI usage via OpenTelemetry. Each headless review is a session-scoped cumulative counter, so totals use last_over_time (sum of each session's final value) — increase()/rate() would read flat because a session exports its total in one shot.",
+  "description": "Claude Code CLI/session telemetry via OpenTelemetry — NOT the durable ORB ai_usage_events DB attribution (see the 'Gittensory - ORB AI usage' dashboard for that, filterable by provider/model/repo/PR). Each headless review is a session-scoped cumulative counter, so totals use last_over_time (sum of each session's final value) — increase()/rate() would read flat because a session exports its total in one shot.",
   "templating": {
     "list": [
       {
--- a/grafana/dashboards/codex-usage.json
+++ b/grafana/dashboards/codex-usage.json
@@ -4,10 +4,10 @@
   "tags": ["gittensory", "codex", "ai"],
   "timezone": "browser",
   "schemaVersion": 39,
-  "version": 3,
+  "version": 4,
   "refresh": "30s",
   "time": { "from": "now-7d", "to": "now" },
-  "description": "Codex review observability for the self-hosted stack. Prometheus panels are subscription-CLI attempt counters for the selected range; durable review records come from ai_usage_events, filtered by provider='codex' (migration 0109), and must be read with their status. Token/cost columns are populated whenever the Codex CLI reports real usage; estimated_neurons is a heuristic budget proxy, not billing-grade accounting.",
+  "description": "Codex review observability for the self-hosted stack. Prometheus panels are subscription-CLI attempt counters for the selected range. Durable review records come from ai_usage_events (migration 0109) and must be read with their status. Two different filters are used deliberately: the token/cost panels (real usage sums) filter strictly on provider='codex', since an unattributed row always contributes 0 to a sum either way. The record-count/status/recent-events panels instead match 'codex' as a reviewer-name token in either provider OR the model label (\"+\"-joined reviewer names, e.g. 'codex', 'codex:o4-mini', 'codex+claude-code') — a quota-blocked call never gets provider attribution (no call ever ran) but IS attributable via its configured model label, so filtering by provider alone silently drops every quota_exceeded row for a codex-configured repo. estimated_neurons is a heuristic budget proxy, not billing-grade accounting.",
   "panels": [
     {
       "id": 1,
@@ -52,7 +52,7 @@
       "id": 5,
       "type": "stat",
       "title": "Successful review records",
-      "description": "Durable Codex-attributed ai_review_pr records with status=ok.",
+      "description": "Durable ai_review_pr records with status=ok where codex was the reviewer — matched by provider attribution OR the configured model label (see dashboard description), so this stays consistent with the status-breakdown panel below rather than under-counting successes where the CLI's response reported no usage envelope at all.",
       "gridPos": { "h": 5, "w": 6, "x": 18, "y": 1 },
       "datasource": { "type": "frser-sqlite-datasource", "uid": "gittensory-db" },
       "fieldConfig": { "defaults": { "unit": "short", "color": { "mode": "fixed", "fixedColor": "green" } } },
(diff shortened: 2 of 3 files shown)


JSONbored:
Addressed the flagged blocker and the test-expectations note in 842e7284:

**Blocker (orb-ai-usage.json:62):** added the `feature <> 'ai_key_change'` filter directly to the "Events missing real usage" query, rather than relying on `recordAiKeyChange`'s `status: "set"|"replace"|"delete"` type constraint in `src/db/repositories.ts` to keep those rows out implicitly. Verified behavior-neutral against a seeded `sqlite3` table matching the `ai_usage_events` schema: identical result for realistic data with/without the filter, and the new filter correctly excludes a hypothetical `ai_key_change` row with `status='ok'` that the old query would have wrongly counted (the current type system rules this out today, but the panel no longer depends on that holding).

**Test expectations:** these are Grafana dashboard JSON files with no corresponding unit test in the repo (confirmed no existing test references `codex-usage.json`/`orb-ai-usage.json`/`claude-usage.json`). In lieu of a unit test, validation evidence for every query change in this PR:
- `node scripts/validate-observability-configs.mjs` — pass (dashboards and alert rules valid)
- All 4 rewritten/added SQL queries verified against a real `sqlite3` instance seeded with rows for every status/provider combination (`quota_exceeded`, `ai_key_change` set/replace/delete, a genuine `status='ok'`-with-null-provider gap row), confirming each panel counts exactly the rows its description claims
- `npm run typecheck` — pass
- Full `npm run test:ci` — green (ran on the earlier commit in this PR; this follow-up is JSON-only with no `src/**` changes)

JSONbored:
Fixed in c743eed2, and this one needed a real code change, not just a defensive no-op — the underlying concern was genuine.

**What was actually wrong:** the panel only checked `provider IS NULL`, missing a second reachable gap: `coerceAiUsage` (ai-review.ts) extracts `provider` independently of `input_tokens`/`output_tokens`/`total_tokens`/`cost_usd` from a provider's raw usage object. A response shaped like `{"usage": {"provider": "codex"}}` with no parseable token counts produces a stored row with `provider='codex'` but all four numeric columns at their zero default — invisible to the old query.

**Why I didn't copy the reviewer's literal suggested SQL:** `(input_tokens IS NULL AND output_tokens IS NULL AND total_tokens IS NULL AND cost_usd IS NULL)` is dead code against this schema — migration `0109` declares all four columns `NOT NULL DEFAULT 0`, and `recordAiUsageEvent`'s `finiteNumber()` coercion means they can never be SQL NULL, only `0`. Ran two independent agents to trace the schema/coercion/extraction code separately, then two adversarial agents that built a scratch `sqlite3` table from the actual migration SQL and ran the corrected predicate against 16 seeded rows total (across both verification passes) covering: the original provider-null case (still counted), the new provider-set-all-zero case (now counted), a provider-set-with-real-nonzero-usage case (correctly NOT counted — no false positive), and `ai_key_change`/`quota_exceeded` rows (still excluded). Also empirically confirmed a literal `NULL` insert into those columns is rejected by the schema itself, ruling out the reviewer's literal predicate for good.

Final query:
```sql
SELECT count(*) AS unattributed FROM ai_usage_events
WHERE status = 'ok' AND feature <> 'ai_key_change'
  AND (provider IS NULL OR (input_tokens = 0 AND output_tokens = 0 AND total_tokens = 0 AND cost_usd = 0))
```

Validation: `node scripts/validate-observability-configs.mjs` (pass), `npm run typecheck` (pass), plus the empirical sqlite3 verification above in lieu of a unit test (no test file covers these dashboard JSONs).

JSONbored:
Fixed in 910267b8. This one was correct in substance but the reviewer's framing conflated two different statuses:

- **quota_exceeded** rows ARE real and DO get dropped by a plain `provider = 'codex'` filter — traced why: `record()` in `ai-review.ts` computes `model` unconditionally via `reviewerModelLabel(env, input)` on every call regardless of status, but `provider` comes from the `actualUsage` parameter, which is never passed at either quota_exceeded call site (the call short-circuits before any provider ever runs). So a quota-blocked review has a real, meaningful `model` (e.g. `codex:o4-mini`) but `provider = NULL` — genuinely invisible to the old filter.
- **disabled/unavailable** rows, also named in the review comment, are not actually affected — traced every return path in `runGittensoryAiReview` and confirmed those two statuses never reach a `record()`/DB-write call at all. There was nothing there for any filter to drop.

Fix: match `'codex'` as a reviewer-name token in *either* `provider` OR the `+`-joined `model` label (reusing the exact pattern this PR replaced, which is already covered by `test/unit/selfhost-grafana-reporting.test.ts`), applied to the "Review record status" panel (flagged), "Recent Codex-attributed review events" (same root cause — confirmed empirically), and "Successful review records" (so it doesn't quietly disagree with the now-fixed status panel on how many successful reviews there were). Left the two token/cost sum panels on the strict `provider` filter and documented why in both dashboard and panel descriptions, since an unattributed row contributes 0 to a sum regardless of filter — no need to complicate those.

Verified with two independent agents tracing the code paths separately, then two adversarial agents running the corrected predicate against seeded sqlite3 data (9-10 rows each pass) covering the flagged case, cross-provider exclusion, dual-reviewer/model-override label formats, substring false-positive safety, cross-feature scoping, and no double-counting — plus a final end-to-end check against the exact shipped query text. `node scripts/validate-observability-configs.mjs` and `npm run typecheck` both pass.

PR #3698, opened 2026-07-06
fix(review): keep gate evaluation for ignored authors

JSONbored:
### Motivation
- The `review.auto_review.ignore_authors` path previously created a completed `skipped` gate check and returned early, which could suppress deterministic gate evaluation and allow required `gate` blockers (like linked-issue) to be bypassed.
- The change scopes ignored-author behavior to suppress public review output and auditing only, while preserving the deterministic gate evaluation when a gate/autonomy conclusion is required.

### Description
- In `maybePublishPrPublicSurface` (`src/queue/processors.ts`) stop publishing a skipped gate check for ignored authors and avoid the early `return`; instead set `publicSurfaceSkipped = true` and only early-return when `shouldEvaluateGate` is false. This preserves downstream deterministic gate computation such as `evaluateGateCheck`.
- Prevent type-label publication when the public surface has been suppressed by adding a `!publicSurfaceSkipped` guard to the type-label branch in `maybePublishPrPublicSurface` (`src/queue/processors.ts`).
- Update the regression unit test (`test/unit/queue.test.ts`) to assert that an ignored author still triggers the gate evaluation (POST+PATCH check-run sequence with a `failure` conclusion for a linked-issue block) and to adjust mocked call counts accordingly.

### Testing
- Ran `git diff --check` which produced no whitespace/conflict errors and passed.
- Ran `npx vitest run test/unit/queue.test.ts -t "ignored authors" --reporter=verbose --no-color` which passed.
- Ran `npx vitest run test/unit/queue.test.ts -t "evaluates the gate while suppressing public review output for ignored authors|audits ignored authors without a skipped check" --reporter=verbose --no-color` which passed.
- Ran `npm run typecheck` (`tsc --noEmit`) which completed without type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4b0b8ce078832c94b80a0730cf6c18)

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -7465,28 +7465,8 @@ async function maybePublishPrPublicSurface(
       reviewEligibility.skipReason,
       webhook.deliveryId,
     );
-    if (gateEnabled) {
-      const gateCheckResult = await createOrUpdateSkippedGateCheckRun(
-        env,
-        installationId,
-        repoFullName,
-        advisory,
-        "Review skipped: ignored author.",
-        mode,
-      );
-      /* v8 ignore next -- permission-missing audit behavior mirrors the existing skipped-check path above. */
-      if (gateCheckResult?.kind === "permission_missing") {
-        await auditGateCheckPermissionMissing(
-          env,
-          author,
-          repoFullName,
-          pr.number,
-          webhook.deliveryId,
-          gateCheckResult.warning,
-        );
-      }
-    }
-    return undefined;
+    publicSurfaceSkipped = true;
+    if (!shouldEvaluateGate) return undefined;
   }
   // A missing author already forces publicSurfaceSkipped=true above (decidePublicSurface's own
   // "missing_author" skip), so the guard just above already returns undefined whenever `!author` combines with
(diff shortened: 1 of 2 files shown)


JSONbored:
Rebased onto current \`main\`. Found and fixed an additional issue surfaced by the rebase: #3694 (merged independently, unaware of this PR) made the gate check-run publish a quiet "skipped (reason)" status whenever auto-review is skipped — including for an ignored author, since it checks the same \`review.auto_review.ignore_authors\` config this PR's eligibility check does. That silently reintroduced the exact bypass this PR closes: an ignored author's gate check-run would show "skipped" instead of the real evaluated conclusion (e.g. a linked-issue hard-rule failure).

Fix: the "skipped status" shortcut now only applies when \`publicSurfaceSkipped\` is false — reusing the existing flag this PR's own ignored-author path already sets, rather than introducing new reason-string matching. A benign content-based skip (draft, WIP title, too-large, docs-only, base-branch, auto-pause) never sets \`publicSurfaceSkipped\`, so #3694's quiet status is unaffected for those cases.

Full \`test/unit/queue.test.ts\` (588 tests) + typecheck + actionlint + audit all green locally.

Issue #3612, opened 2026-07-05
(stretch) Short-clip/GIF evidence capture for non-static-visible changes


JSONbored:
Part of #3607. Sequenced after #3608–#3611 (the static-screenshot path) are live and validated — does not block them.

## Motivation
The original ask also covers changes that aren't visible in a static before/after screenshot (an animation, a hover/focus state, a multi-step interaction, a scroll-triggered effect) — for these, a short clip or GIF is more honest evidence than a single frame. Nothing in the existing pipeline captures this; it's PNG-only today.

## Requirements
- Scope narrowly for a first cut rather than open-ended "demonstrate any interaction" — e.g. a fixed, short, scripted pass per route (auto-scroll through the page, or hover/focus the first few interactive elements) rather than inferring WHAT interaction matters from the diff, which is a much harder, more speculative problem.
- Reuse the same rendering backend (Cloudflare Browser Rendering / self-host browserless) rather than introducing a second one — check what the Puppeteer API / Browser Rendering binding supports for multi-frame capture (may require assembling frames into a GIF client-side rather than a native "record video" API, depending on binding support).
- Storage: a short GIF/clip is meaningfully larger than a PNG — confirm the R2 / self-host filesystem blob store's size assumptions still hold, and consider a stricter retention/cleanup policy for this asset type specifically.
- Must not block or slow down the static-screenshot sub-issues; this is deliberately sequenced last.

## Deliverables
- [ ] Written scoping decision: what exact interaction sequence gets captured, and why that's sufficient evidence without being a general "demonstrate the change" system
- [ ] Capture + storage implementation for the chosen sequence
- [ ] Rendered in the PR comment alongside (not replacing) the static before/after table

## Expected outcome
PRs whose change genuinely isn't visible in a static screenshot still get real automated evidence, without requiring the contributor to record and upload their own video.

## Effort
L — the most open-ended piece; deliberately scoped as a stretch goal, not a hard commitment.

PR #3712, opened 2026-07-06
fix(review): bound visual screenshot rendering

JSONbored:
### Motivation
- Full-page screenshots (`fullPage: true`) let attacker-controlled document height drive unbounded Chromium raster work and unbounded R2 storage from the public `/gittensory/shot?url=` surface.  
- The change restores safety by bounding document height, pixel area, output size, and raster timeout while preserving full-page captures for normal review pages.

### Description
- Add explicit caps and a bounded helper: `MAX_SCREENSHOT_HEIGHT`, `MAX_SCREENSHOT_PIXELS`, `MAX_SCREENSHOT_BYTES`, and `SCREENSHOT_TIMEOUT_MS` plus `captureBoundedFullPageShot(page, viewport)` which measures document height, enforces the caps, times out the raster, and checks returned PNG bytes.  
- Replace the direct `page.screenshot({ type: "png", fullPage: true })` call with the bounded helper so oversized or slow renders degrade to `png: null`.  
- Classify the public screenshot route as `expensive` in `routeClassForPath` so the endpoint is rate-limited more strictly.  
- Add regression/unit tests in `test/unit/visual-shot.test.ts` and update `test/unit/auth.test.ts` to cover bounded full-page captures, over-height pages, excessive pixel-area, oversized PNG output, a screenshot timeout, and the route classification.

### Testing
- Ran `npx vitest run test/unit/visual-shot.test.ts test/unit/auth.test.ts` and all targeted unit tests passed (47 tests passed).  
- Ran `npm run typecheck` and `git diff --check`, both succeeded locally.  
- Attempted to run the full gate with `npm run test:ci`, but the environment hit unrelated long-running/unit timeouts during unrelated queue/backfill tests so the full suite could not be completed here.  
- Attempted `npm audit --audit-level=moderate`, but the registry audit endpoint returned `403` in this environment so that check could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4b34f1c7d8832ca7eadc4c450361a7)

### DIFF
--- a/src/auth/rate-limit.ts
+++ b/src/auth/rate-limit.ts
@@ -108,6 +108,7 @@ export function routeClassForPath(path: string): RateLimitClass {
   if (path === "/v1/orb/ingest") return "strict";
   if (path === "/v1/auth/session" || path === "/v1/auth/logout") return "normal";
   if (path.startsWith("/v1/auth/")) return "strict";
+  if (path === "/gittensory/shot") return "expensive";
   if (
     path.includes("/branch-analysis") ||
     path.includes("/v1/agent/") ||
--- a/src/review/visual/shot.ts
+++ b/src/review/visual/shot.ts
@@ -34,9 +34,17 @@ type ScreenshotRequest = {
   abort(): Promise<unknown>;
   continue(): Promise<unknown>;
 };
+type ScreenshotPage = {
+  evaluate<T>(fn: () => T): Promise<T>;
+  screenshot(options: { type: "png"; fullPage: true }): Promise<Uint8Array>;
+};
 export const DESKTOP_VIEWPORT: Viewport = { width: 1440, height: 900 };
 export const MOBILE_VIEWPORT: Viewport = { width: 390, height: 844 }; // iPhone-class portrait
 const VIEWPORT = DESKTOP_VIEWPORT;
+export const MAX_SCREENSHOT_HEIGHT = 10000;
+export const MAX_SCREENSHOT_PIXELS = 14_400_000; // 1440 × 10000, matching the full-page cap.
+export const MAX_SCREENSHOT_BYTES = 5 * 1024 * 1024;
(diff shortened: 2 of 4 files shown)


JSONbored:
Rebased onto current \`main\` to clear the \`codecov/patch\` failure (patch coverage was 83.33%, 4 lines missing in \`src/review/visual/shot.ts\` — the \`page.evaluate()\` height-computation callback inside \`captureBoundedFullPageShot\` was never actually invoked by the test's mock, just short-circuited).

The rebase conflicted with \`test/unit/visual-shot.test.ts\` against the since-merged scroll-through GIF capture feature (#3612/#3688), which had already solved the identical problem (a mocked \`evaluate()\` that needs to really invoke its callback for coverage) with a cleaner technique than my first attempt — it just calls the callback and swallows the expected "no \`document\` in this Node test env" throw, rather than stubbing a fake DOM object. Verified that resolution already covers the previously-missing lines (confirmed via local coverage), so I dropped my own redundant fix and kept the merged result. 45/45 tests pass, typecheck/actionlint/audit clean.

JSONbored:
Addressed Superagent's P2 finding (\`src/review/visual/shot.ts:126\`): the pre-capture height check runs \`page.evaluate()\` inside the target page's own JS context, so a malicious page can override \`document.body.scrollHeight\`/\`offsetHeight\` (e.g. via \`Object.defineProperty\`) to report a small height, pass the bound, and still get rasterized to its real, oversized size.

Added a tamper-proof recheck: decode the actual width/height Chromium wrote into the output PNG's own IHDR chunk (a cheap ~24-byte header read, not a re-render) and reject if those real dimensions exceed the same bounds — independent of anything the page's own JS could influence. 2 new tests (spoofed-height rejection + normal-dimensions acceptance), 47/47 pass, typecheck/actionlint clean.

PR #4027, opened 2026-07-07
fix(queue): repair main-red syntax error from #3998

JSONbored:
## Summary

main is currently broken (confirmed via direct GitHub API fetch of the file at HEAD, not a local artifact): #3998 (a sentry[bot] auto-fix PR, merged 2026-07-07T10:09 UTC) applied its diff hunks in the wrong positions, corrupting `src/queue/processors.ts` in 4 places:

- `sweepRepoRegate`'s job-builder: an inserted comment+spread split a ternary expression mid-statement.
- A duplicate `REGATE_REPAIR_MAX_ATTEMPTS_PER_SHA` declaration (block-scoped redeclaration), with the new value (5) landing inside a split comment instead of replacing the old value (2).
- `regatePullRequest`'s signature: a required parameter placed after an optional one (invalid TS).
- `regatePullRequest`'s body: a stray `}` closed the JOBS.send payload early, orphaning the trailing spread lines.

Reconstructed the logic exactly as #3998's own PR description states the intent: move the `REGATE_REPAIR_ATTEMPT_EVENT_TYPE` audit recording from dispatch-time to execution-time (so a rate-limit-deferred dispatch no longer burns an attempt), thread `repairHeadSha` through the job payload as a trailing optional param, and keep the attempt cap bumped 2→5.

Also fixed, discovered while getting this file green:
- Two `queue.test.ts` regression tests still asserted the old dispatch-time-records-immediately behavior and the old cap of 2 — updated to match the new execution-time/cap-5 semantics #3998 intended.
- One stale `gittensory_agent_disposition_total` assertion missing the `repo` label that `selfhost-metrics.test.ts`'s equivalent assertions already correctly include.

## Validation

- `npm run typecheck` — clean (was failing with TS1005/TS1128/TS1109/TS1136 parse errors before this fix).
- `npx vitest run test/unit/queue.test.ts` — full file green after the fixes above (was 6 failing, now 0).

## Notes

This blocks cutting a new self-host release from the current main tip, so treating as urgent.

### DIFF
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -1080,13 +1080,13 @@ export async function processJob(env: Env, message: JobMessage): Promise<void> {
       // One bounded re-gate unit fanned out by the sweep (#audit-sweep-fanout): re-review + stamp a single PR.
       await regatePullRequest(
         env,
-        message.repairHeadSha,
         message.repoFullName,
         message.prNumber,
         message.installationId,
         message.deliveryId,
         message.force,
         message.prCreatedAt,
+        message.repairHeadSha,
       );
       return;
     case "run-agent":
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as redundant — another concurrent effort fixed this exact bug via #4023 (merged 10:46 UTC) plus additional pre-existing drift via #4019 (merged 10:47 UTC), both landing minutes before this PR was ready. Verified origin/main now typechecks cleanly. No need to carry this fix forward.

Issue #4184, opened 2026-07-08
fix(review): visual-capture screenshots render as broken images on self-host (private origin)


JSONbored:
## What happened

On PR #4180's visual-capture review comment, both viewport thumbnails in the "Visual preview" table rendered as broken images, and the "Before (production)" column was blank for both rows.

## Root cause

The comment's `<img src>` pointed directly at `https://edge-us-01.raccoon-bushi.ts.net/gittensory/shot?url=...` — a Tailscale MagicDNS (`*.ts.net`) tailnet hostname. That's only reachable from devices on the same tailnet; GitHub's own servers, which render the comment's images, cannot resolve or fetch it. Every screenshot embedded by this self-host instance has been silently broken the same way.

`PUBLIC_API_ORIGIN` on this instance is set to that private hostname instead of a publicly-reachable origin. Separately, `PUBLIC_SITE_ORIGIN` (the "before/production" screenshot source) was never set at all — it isn't documented anywhere in `.env.example`'s self-host section, so there was nothing prompting an operator to configure it.

## Fix

- Root cause is a live deployment config value, not something fixable from the repo.
- Added a boot-time advisory (`publicOriginReachabilityAdvisory` in `src/selfhost/health.ts`, mirroring the existing `sqliteBackupAdvisory` pattern) that warns loudly — and exposes a Prometheus gauge — when `PUBLIC_API_ORIGIN`/`PUBLIC_SITE_ORIGIN` look like a private/internal hostname, with an explicit `PUBLIC_ORIGIN_ACKNOWLEDGED` escape hatch for a legitimately Funnel-exposed `*.ts.net` node.
- Documented `PUBLIC_SITE_ORIGIN` in `.env.example` for the first time, and corrected `PUBLIC_API_ORIGIN`'s misleading "not needed after setup" framing in both example files.

## Remaining

The live edge-us-01 instance's own `PUBLIC_API_ORIGIN`/`PUBLIC_SITE_ORIGIN` values still need to be corrected on that deployment directly — that's an ops change outside this repo.

RealDiligent:
Quick follow-up: this appears already fixed by merged PR #4185 (commit 68dabac1 on main). I verified upstream/main contains that commit and the related PUBLIC_API_ORIGIN/PUBLIC_SITE_ORIGIN handling paths. Remaining note in the issue body looks deployment-ops specific, not repo code. If that matches maintainer intent, this issue can likely be closed as code-complete.

[...]

PR #4209, opened 2026-07-08
feat(review): upload visual-capture screenshots to a public R2 bucket

JSONbored:
## Summary

- Advances #4184. PR #4185 added a boot-time advisory for a `PUBLIC_API_ORIGIN` that looks private, but for an instance that's deliberately, permanently private (e.g. Tailscale-only, no public HTTP surface by design), no value of `PUBLIC_API_ORIGIN` can ever fix the broken-screenshot symptom — the whole approach of linking through this instance's own origin is the wrong shape for that deployment.
- Adds an optional, config-driven alternative: `R2_PUBLIC_ACCOUNT_ID` / `R2_PUBLIC_BUCKET` / `R2_PUBLIC_ACCESS_KEY_ID` / `R2_PUBLIC_SECRET_ACCESS_KEY` / `R2_PUBLIC_BASE_URL` (all five required together, otherwise not configured — byte-identical to today). When set, every captured screenshot (before/after PNGs, pixel-diff overlays, scroll-through GIFs, and the actions-fallback workflow's own captures) is uploaded a second time to a dedicated, deliberately public Cloudflare R2 bucket, and its direct URL used in the PR comment instead of the private instance's own origin — so the private instance itself never has to answer a single public request.
- New `src/selfhost/r2-public-upload.ts`: a minimal, self-host-only (Node `node:crypto`) AWS SigV4 request signer for R2's S3-compatible API. No new dependency — hand-rolled rather than pulling in `@aws-sdk/client-s3` for the one operation this needs (a single-object PUT).
- A cache-hit is trusted to already be mirrored to the public bucket (since both writes always happen together going forward), so there's no double-upload cost on repeat reviews of the same head SHA — documented as a one-time transition caveat for any local cache entry written before this feature was enabled.

## Scope

- [x] Focused — one new capability (public-bucket mirroring), no unrelated changes.
- [x] Follows `CONTRIBUTING.md`.
- [x] Owner PR — linked issue eligibility gate does not apply; advances #4184 (does not close it — see Notes).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run docs:drift-check`
- [x] `npm run selfhost:env-reference` — regenerated, 5 new vars appear.
- [x] `npm run db:migrations:check`, `npm run cf-typegen:check` — clean (no schema/binding changes; these are plain self-host runtime env vars).
- [x] `npm audit --audit-level=moderate` — 0 vulnerabilities.
- [x] SigV4 signer independently cross-checked against a Python (`hmac`/`hashlib`) reference implementation for a fixed test vector — exact match — before trusting it in any test.
- [x] End-to-end verified against a real R2 bucket before writing any test: created a dedicated bucket, enabled its public dev URL, signed and uploaded a real object, fetched it back publicly, confirmed byte-identical content, then deleted the test object.
- [x] `npx vitest run test/unit/r2-public-upload.test.ts test/unit/visual-capture.test.ts test/unit/queue.test.ts` — all green; 100% line/branch/function coverage on `src/selfhost/r2-public-upload.ts`; every changed line/branch in `src/review/visual/capture.ts` and `src/queue/processors.ts` covered (confirmed via `git diff` cross-referenced against the coverage report — pre-existing gaps in code this PR doesn't touch were left alone).
- [x] `npm run test:changed` (against `origin/main`) — 2181/2181 passing.
- [ ] `npm run test:workers`, `build:mcp`, `test:mcp-pack`, `ui:openapi:check`, `ui:lint`, `ui:typecheck`, `ui:build` — not re-run locally (no worker-entry/MCP/OpenAPI/UI-component code touched); CI runs them authoritatively.

## Safety

- [x] No secrets committed — the R2 credentials used for end-to-end verification were passed only as inline environment variables for one-off local commands, never written to a file in this repo or worktree.
- [x] No behavior change for any repo/instance that doesn't opt in.
- [x] Non-blocking by design: an upload failure falls back to the existing private-origin URL, never breaks the review.

## Notes

Remaining before this actually fixes the live instance: cut a new self-host release image including this code, deploy it, then set the five `R2_PUBLIC_*` values in the live `.env` — none of that is part of this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -182,6 +182,19 @@ GITTENSORY_REVIEW_DRAFT=false
 #                                            #   bare Tailscale MagicDNS `*.ts.net` address) once you've confirmed
 #                                            #   it's genuinely public (Funnel enabled, a reverse proxy in front of
 #                                            #   it, etc.). Default false (warning shown).
+
+# --- Public screenshot bucket (JSONbored/gittensory#4184; optional) ---
+# For an instance that stays deliberately private (e.g. Tailscale-only, no public HTTP surface at all):
+# uploads visual-capture screenshots to a dedicated PUBLIC Cloudflare R2 bucket and links directly to it,
+# instead of requiring PUBLIC_API_ORIGIN itself to be publicly reachable. All five required together; any
+# one missing means not configured (captures keep using PUBLIC_API_ORIGIN, exactly as before this feature).
+# R2_PUBLIC_ACCOUNT_ID=                       # your Cloudflare account ID
+# R2_PUBLIC_BUCKET=                           # a bucket dedicated to this -- never reuse one holding other
+#                                            #   audit/private data, since every object in it is world-readable
+# R2_PUBLIC_ACCESS_KEY_ID=                    # from R2 -> Manage API Tokens -> Create -> "Object Read & Write",
+# R2_PUBLIC_SECRET_ACCESS_KEY=                #   scoped to ONLY the bucket above -- shown once, save it then
+# R2_PUBLIC_BASE_URL=                         # the bucket's public base URL: its r2.dev dev URL (R2 -> bucket
+#                                            #   -> Settings -> Public access -> Allow Access), or a custom domain
 # SELFHOST_SETUP_TOKEN=change-this-long-random-value  # REQUIRED to unlock the first-run /setup wizard. Without it
 #                                            #   /setup returns 400; with it, enter the token in the browser form
 #                                            #   or send an x-setup-token / Bearer header. Never put this token in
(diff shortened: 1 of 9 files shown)


JSONbored:
Closing as superseded by #4203, which landed a more complete version of this same idea (advances #4184) — a generic S3-compatible backend (`aws4fetch`, region-aware, bounded retry budget, `delete()` parity) replacing `REVIEW_AUDIT`'s own backend selection, rather than this PR's parallel-upload-to-a-second-bucket approach. #4203 was opened before this PR and is already merged.

The bucket + credentials created for this PR (`gittensory-visual-capture-public`) can still be reused as the actual S3-compatible bucket #4203's `REVIEW_AUDIT_S3_*` config points at — no need to provision a second one.

Issue #4297, opened 2026-07-08
feat(miner-concurrency): add git-worktree-per-attempt allocator


JSONbored:
Note the scope split up front: the tracker (#2353, Phase 3) separately lists `feat(miner-hands): git-worktree-per-attempt isolation primitive` (still open, not yet released as an individually-filed contributor issue). That item defines the PRIMITIVE — create/teardown one git worktree for one coding-agent attempt. This issue is the different, complementary piece: the ALLOCATOR that manages a POOL of worktrees across multiple concurrent fleet attempts (acquire/release, a concurrency cap, and reclaiming orphaned worktrees after a crash) — the piece fleet mode (this Phase's sibling Dockerfile issue) actually needs in order to run more than one attempt at a time without two attempts colliding on the same working tree. Confirmed by search: there is currently zero worktree-related code anywhere in `packages/gittensory-miner/` or `packages/gittensory-engine/` — this is genuinely new ground, not a rename or extension of something that already exists.

The package's existing local-store convention is the shape to reuse rather than invent a new persistence style: `packages/gittensory-miner/lib/claim-ledger.js`, `lib/run-state.js`, and `lib/portfolio-queue.js` are all plain-JS, `node:sqlite`-backed local stores under `GITTENSORY_MINER_CONFIG_DIR` — claim-ledger.js's own header comment calls this out explicitly as the pattern siblings should mirror ("Mirrors the package's existing local-store pattern (run-state.js, portfolio-queue.js, event-ledger.js) — plain JS + node:sqlite", `claim-ledger.js:9-10`). An allocator needs the same kind of durable bookkeeping — which worktree paths are currently allocated to which attempt IDs — so a crashed/restarted miner process can tell a stale allocation from a live one instead of silently reusing (or permanently leaking) a path.

## Deliverables
- [ ] `packages/gittensory-miner/lib/worktree-allocator.js` (or split pure scheduling logic into `gittensory-engine` with a thin SQLite-bookkeeping wrapper in the miner package — either split is fine as long as the pure allocation logic is unit-testable without touching a real filesystem).
- [ ] `acquire(attemptId, repoFullName)` → allocates an unused worktree path (base directory configurable, defaulting under `GITTENSORY_MINER_CONFIG_DIR`), enforcing a configurable max-concurrency cap.
- [ ] `release(attemptId)` → frees the path for reuse.
[...]

PR #4455, opened 2026-07-09
test(miner-concurrency): parallel-attempt collision regression suite

### LINKED ISSUE #4298: test(miner-concurrency): add parallel-attempt collision regression suite
This depends on the git-worktree-per-attempt allocator (sibling Phase 6 issue, `feat(miner-concurrency): add git-worktree-per-attempt allocator`) landing first — there is nothing to regression-test until `acquire`/`release` exist. Filing this now so the dependency is visible in the tracker, but a contributor should pick up the allocator issue first (or claim both together).

Once the allocator exists, this is a focused regression suite proving the concurrency guarantees the allocator issue's deliverables promise, rather than just its happy path: that concurrent `acquire()` calls never hand out the same worktree path, that `release()` correctly frees a path for reuse, that the configured concurrency cap is actually enforced under simultaneous requests (not just checked sequentially), and that a simulated crash (an allocation recorded as active with no live owner) is reclaimed on the next startup rather than either double-allocated or leaked forever.

## Deliverables
- [ ] `test/unit/miner-worktree-allocator-collisions.test.ts` (or fold into the allocator's own primary test file if that ends up small enough — match whatever filename the allocator issue actually lands, since `test/unit/miner-*.test.ts` is the established location per `test/unit/miner-claim-ledger.test.ts` and siblings).
- [ ] Concurrent-acquire test: fire N simultaneous `acquire()` calls (e.g. via `Promise.all`) and assert all N returned paths are distinct.
- [ ] Cap-enforcement test: request more than the configured concurrency limit simultaneously and assert the excess requests are rejected or queued (per whatever contract the allocator issue settles on), never silently over-allocated.
- [ ] Release/reuse test: `release()` a path, then `acquire()` again, and assert the freed path becomes available (not permanently retired).
- [ ] Crash-recovery test: seed the allocator's persisted state with an "active" allocation that has no live owner, restart/reopen the store, and assert it is reclaimed rather than treated as still-live.
- [ ] Follow the existing miner-package test conventions (`node:sqlite`-backed local-store tests use a temp DB path per test — see `test/unit/miner-claim-ledger.test.ts` / `test/unit/miner-run-state.test.ts` for the setup/teardown pattern already in use).

## References
- Sibling issue: `feat(miner-concurrency): add git-worktree-per-attempt allocator` (Phase 6) — hard dependency, must land first.
- `test/unit/miner-claim-ledger.test.ts`, `test/unit/miner-run-state.test.ts` — existing SQLite-backed local-store test setup/teardown pattern to reuse for the allocator's own store.

kiannidev:
## Summary

- Add `packages/gittensory-miner/lib/worktree-allocator.js` — SQLite-backed `acquire`/`release` pool with configurable concurrency cap and orphan reclamation on startup (#4297 prerequisite).
- Add `test/unit/miner-worktree-allocator-collisions.test.ts` proving:
  - concurrent `acquire()` returns distinct worktree paths
  - excess parallel acquires are rejected at the cap (`worktree_capacity_exceeded`)
  - `release()` frees a path for reuse
  - crash recovery reclaims orphaned active allocations on reopen
- Add `test/unit/miner-worktree-allocator.test.ts` for path resolution and basic allocator behavior.

Closes #4298

Also delivers the allocator foundation required by #4297 (hard dependency noted in the issue).

## Test plan

- [x] `npm run build:miner`
- [x] `npx vitest run test/unit/miner-worktree-allocator.test.ts test/unit/miner-worktree-allocator-collisions.test.ts`
- [ ] CI: `validate`, `codecov/patch`, security

## UI Evidence

N/A — test-only / miner library PR (no visible UI).

Made with [Cursor](https://cursor.com)


JSONbored (commented):
Fix + resubmit: 

```
test/unit/miner-worktree-allocator-collisions.test.ts's 'concurrent'/'simultaneous' acquire tests only chain Promise.resolve().then() calls within a single Node process; since DatabaseSync is synchronous and JS is single-threaded, these calls never actually race, so the BEGIN IMMEDIATE transaction/raced-recheck branch in worktree-allocator.js acquire() (lines ~180-200) is never genuinely exercised — the suite proves the allocator's plain (non-concurrent) logic works, not that it survives real multi-process collisions, which is the actual failure mode the allocator exists to prevent.
```

PR #29, opened 2026-05-29
feat(bounties): add lifecycle watcher signals

Crystora:
## Summary

Close: #23 

- Add bounty lifecycle checks to advisories, preflight, issue quality, opportunity ranking, MCP tools, and agent flows.
- Distinguish active, completed, cancelled, historical, stale, and ambiguous bounty states so stale or inactive bounty work is not treated as an active opportunity.
- Include linked PR context in bounty advisory results and add focused lifecycle coverage in tests.

## Validation

- [x] `npm run test:ci`
- [x] `npm run changelog:check`

## Safety

- [x] Backend-only change
- [x] No secrets, wallet details, user PATs, raw trust scores, or private rankings exposed
- [x] Public text avoids compensation-seeking or optimization-tactic language
- [x] OpenAPI/MCP behavior updated where needed
- [x] Public docs/changelogs updated where needed

PR #4492, opened 2026-07-09
fix(review): keep regate sweep drift backstop

JSONbored:
### Motivation

- The scheduled re-gate sweep must serve as a same-head drift backstop for base/sibling/manifest/settings changes, but the selector was made one-shot and permanently excluded any PR with a `lastRegatedAt` marker, allowing stale successful gate checks to persist.

### Description

- Restore sweep eligibility by treating `lastRegatedAt` as a progress/sort marker rather than a permanent deny-list in `selectRegateCandidates` (`src/settings/agent-sweep.ts`).
- Preserve the `oldest-first` initial-drain semantics by narrowing the ordering pool only during the initial drain and otherwise cycling the full staleness pool; implement an `orderingPool` that filters only for the initial-drain case.
- Update unit tests in `test/unit/agent-sweep.test.ts` to assert that already-regated PRs can be reselected after the freshness window in both staleness and `oldest-first` modes.

### Testing

- Ran the focused selector unit suite with `npx vitest run test/unit/agent-sweep.test.ts`, and all tests passed.
- Ran the targeted sweep integration check with `npx vitest run test/unit/queue.test.ts -t "agent re-gate sweep recomputes stale open PR verdicts"`, which passed for the scoped test(s).
- Attempted coverage with `npx vitest run --coverage test/unit/agent-sweep.test.ts`; tests passed but the local coverage remapper errored with `TypeError: jsTokens is not a function` so coverage reporting was not produced locally.
- Attempted `npm audit --audit-level=moderate`; the registry audit endpoint returned `403 Forbidden`, so the audit step did not complete locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500cdbee00832c90a8b896835996b9)


JSONbored:
Closing — this reverts the fail-closed fix from 4a86e8e26 (2026-07-09, same day), which stopped the confirmed-live incident where the regate sweep endlessly re-checked already-reviewed, permanently-red owner PRs on every tick, driving the primary GitHub rate limit to zero twice in two hours.

This PR's own CI fails the regression guard added alongside that fix: `test/unit/agent-sweep.test.ts:275` (`#never-endless-reregate`) — "PR #29 was already regated pre-test and is not repair-priority, yet was re-selected."

The "drift backstop" concern in the new comment (catching a moved base / merged sibling / changed manifest without a new push) is legitimate, but 4a86e8e26 traded it off deliberately: re-review now happens via the real-time webhook path or an explicit maintainer re-run, not the sweep, specifically to stop the rate-limit exhaustion. If we still want a same-head drift backstop, it needs a bounded/rate-limited re-check (e.g. re-include only after N hours, or cap re-checks per tick) — not unconditional re-inclusion that reopens the exact failure mode the incident fix closed.

Issue #4512, opened 2026-07-09
fix(review): velocity-aware exception in the unlinked-issue-match escalation for machine-cadence (miner) authors


JSONbored:
## Context

A maintainer review-stack × Autonomous Miner System (AMS) integration audit (2026-07-09) found that the unlinked-issue credibility-farming guardrail's repeat-escalation rule is calibrated for human abuse cadence and produces the wrong outcome under machine cadence.

## Evidence

- `src/review/unlinked-issue-guardrail.ts:22-25` — a 90-day repeat window ("a match from a year ago shouldn't silently escalate... forever").
- `src/review/unlinked-issue-guardrail.ts:52-55,76-127` — `hasPriorUnlinkedIssueMatch` is keyed only by `authorLogin` and correlated across **any repo** in the install; the first match holds the PR, a confirmed repeat within the window escalates straight to auto-**close** (lines 113-119).
- `src/db/repositories.ts:2609-2617` — `hasRecentAuditEventForOtherTarget` confirms the cross-repo, no-repo-scoping correlation.
- Issue #3555's own design rationale (the guardrail's origin issue) is written purely in terms of a human being unlikely to repeat the same "forgot to link the issue" mistake twice in 90 days — it never considers a bot/miner author.

## Why this matters for the Autonomous Miner System

A systematic bug in how a miner auto-generates its PR body (e.g., a template that occasionally omits `Closes #N` even though it correctly solved a real, ranked issue) could recur within hours rather than months at machine cadence. Because the correlation key is author login across **all** repos in the install, a miner fleet's very first two legitimate PRs system-wide could trigger hold-then-close, converting one tooling bug into cross-repo closure of otherwise-correct, properly-targeted work — with no cadence-aware distinction between deliberate farming and a fast, legitimate miner repeatedly hitting its own formatting defect. The guardrail is opt-in per repo (`mode: off` by default), which bounds blast radius somewhat, but the core failure mode reproduces within a single opted-in repo with no cross-repo amplification needed.

## Deliverables

- [ ] Add a velocity/cadence-aware exception to the repeat-escalation path: e.g. require a minimum time gap between the first hold and the confirmed repeat before escalating to auto-close, or cap escalation severity (hold again, rather than close) when both matches occur within an implausibly short human window
- [ ] Ensure the exception can't be trivially exploited to launder repeated deliberate abuse (the fix should distinguish "two legitimate fast submissions" from "rapid-fire farming," not just widen the window uniformly)
[...]

PR #4551, opened 2026-07-10
fix(review): per-actor rate ceiling + shared budget check on unlinked-issue AI verifier

### LINKED ISSUE #4515: fix(review): add a per-actor rate ceiling on the unlinked-issue-match AI verifier, ahead of its existing post-hoc repeat-escalation check
## Context

A maintainer review-stack × Autonomous Miner System (AMS) integration audit (2026-07-09) found that the unlinked-issue-match AI verifier — unlike every sibling free-AI-advisory feature — has no cost ceiling on how often it runs per actor.

## Evidence

- `src/review/unlinked-issue-guardrail.ts:76-127` — `resolveUnlinkedIssueMatchDisposition` runs its config/linked-issue-count short-circuits, then unconditionally loops candidates into `verifyUnlinkedIssueMatch` — no actor-level call-frequency check before this loop.
- `src/signals/unlinked-issue-candidates.ts:31` — `MAX_CANDIDATES = 3` bounds cost per PR only, not per actor over time.
- `src/review/unlinked-issue-guardrail.ts:52-55` — the repeat-check (`hasPriorUnlinkedIssueMatch`) only runs **after** a candidate has already cleared the AI verifier — i.e. after the AI cost is already spent, so it cannot prevent the spend.
- By contrast, sibling free-AI-advisory features (`src/services/ai-slop.ts:199-208`, `src/services/linked-issue-satisfaction-run.ts`) explicitly consult the shared `AI_DAILY_NEURON_BUDGET` before spending — `src/review/unlinked-issue-match.ts`'s `verifyUnlinkedIssueMatch` calls `env.AI.run` directly with **no budget check of any kind**, not even the shared global ceiling its siblings use.

## Why this matters for the Autonomous Miner System

For an occasional human abuser this is a non-issue — natural submission rate bounds the cost. For a machine-paced fleet opening many unlinked PRs per hour across an install's repos, this AI-verifier cost is genuinely unbounded per-actor even though it's bounded per-PR: exactly the "built for occasional human abuse, not fleet-paced abuse" pattern that breaks at AMS scale, and a real spend/availability risk independent of any anti-farming concern.

## Deliverables

- [ ] Add a per-actor rate ceiling (or at minimum wire this path into the existing shared `AI_DAILY_NEURON_BUDGET` check its siblings already use) before `verifyUnlinkedIssueMatch` is invoked
- [ ] Decide the fallback behavior when an actor exceeds the ceiling (skip the AI check and fall back to the deterministic disposition, vs. queue/delay)
- [ ] Tests: budget-exceeded fallback path, normal-actor path unaffected, existing per-PR `MAX_CANDIDATES` bound unaffected
- [ ] Docs: note the new ceiling alongside the guardrail's other config knobs

---
_Surfaced by a maintainer review-stack × AMS integration audit (2026-07-09). Related: #3555 (the guardrail's origin issue), #2344 (miner-side rate-limit/backoff wiring — the write-side counterpart to this read/AI-cost-side gap)._

JSONbored:
## Summary

The unlinked-issue-match AI verifier (`src/review/unlinked-issue-match.ts`) had no cost bound of its own — every unlinked PR that produces at least one deterministic pre-filter candidate runs a real AI call, with no per-actor rate awareness and no check against the shared daily neuron budget every other free-tier AI feature (`ai-review.ts`, `ai-slop.ts`) already respects.

This adds two gates in `resolveUnlinkedIssueMatchDisposition` (`src/review/unlinked-issue-guardrail.ts`), both running **before** the candidate-verification loop:

- **Per-actor rate ceiling**: tracks AI-verifier invocations via a new `github_app.unlinked_issue_verify_attempt` audit event (cross-repo, keyed on PR author login) and skips verification once an actor hits 15 attempts in an hour. An unidentifiable author (no login) can't be rate-limited individually, so only the shared budget check applies to them — consistent with how the rest of this file already treats an unknown author.
- **Shared daily neuron budget check**: estimates the worst case for this PR's candidate loop (every candidate needs a fallback retry) and compares against `sumAiEstimatedNeuronsSince`/`AI_DAILY_NEURON_BUDGET`, mirroring `ai-slop.ts`'s own pre-call budget check exactly (same default/clamp) since both draw from the same counter.

Both gates fail toward "proceed as if this layer didn't exist" on any read error — a DB hiccup must never silently disable the guardrail itself, only this additive cost control.

`MAX_CANDIDATES` (signals/unlinked-issue-candidates.ts) and `MAX_TOKENS`/`DIFF_CHAR_BUDGET` (review/unlinked-issue-match.ts) are now exported so the orchestrator sizes its estimate off the same numbers rather than a second, driftable copy.

Fixes #4515

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run test/unit/unlinked-issue-guardrail.test.ts test/unit/unlinked-issue-match.test.ts test/unit/unlinked-issue-candidates.test.ts test/unit/ai-slop.test.ts test/unit/ai-review.test.ts` — 287/287 passing
- [x] Targeted coverage on `src/review/unlinked-issue-guardrail.ts`: 100% branch, 100% lines on all new code (the one function coverage gap reported is a pre-existing #4512 line untouched by this diff)
- [x] New tests cover: rate-ceiling-exceeded skip, rate-ceiling boundary (14 ok, 15 blocks), cross-actor isolation, attempt recording, rate-ceiling read-failure fail-safe, budget-exceeded skip, budget-unset default-to-10M, budget read-failure fail-safe

### DIFF
--- a/src/db/repositories.ts
+++ b/src/db/repositories.ts
@@ -2636,6 +2636,21 @@ export async function mostRecentAuditEventForOtherTarget(env: Env, actor: string
   return rows[0]?.createdAt ?? null;
 }
 
+/** Count-returning, cross-repo variant (#4515): how many recent events of this type has this actor generated,
+ *  across EVERY target (repo/PR), within the recency window? Unlike {@link countRecentAuditEventsForActorAndTarget}
+ *  (scoped to one target thread), this counts an actor's ACTIVITY VOLUME irrespective of which PR/repo each
+ *  event landed on -- backs a per-actor rate ceiling on an expensive operation (e.g. a paid AI call) that a
+ *  single-target-scoped counter would never catch for an actor spreading attempts across many repos. */
+export async function countRecentAuditEventsForActor(env: Env, actor: string, eventType: string, sinceIso: string): Promise<number> {
+  const db = getDb(env.DB);
+  const [row] = await db
+    .select({ count: sql<number>`count(*)` })
+    .from(auditEvents)
+    .where(and(eq(auditEvents.actor, actor), eq(auditEvents.eventType, eventType), gte(auditEvents.createdAt, sinceIso)));
+  /* v8 ignore next -- count(*) always returns exactly one row; the empty-array guard only satisfies the destructure type. */
+  return row?.count ?? 0;
+}
+
 /** Count-returning variant of {@link hasRecentAuditEvent}, additionally scoped to one `targetKey` (e.g. a single
  *  `owner/repo#123` PR/issue) rather than the actor's activity across the whole repo. Backs the review-request
  *  nagging cooldown (#2463): counting how many `@gittensory` pings a contributor has sent on ONE thread within
--- a/src/review/unlinked-issue-guardrail.ts
+++ b/src/review/unlinked-issue-guardrail.ts
@@ -18,12 +18,13 @@ import {
   getFreshOfficialMinerDetection,
   mostRecentAuditEventForOtherTarget,
   listOpenIssues,
+  recordAiUsageEvent,
   recordAuditEvent,
   sumAiEstimatedNeuronsSince,
   upsertOfficialMinerDetection,
 } from "../db/repositories";
 import { fetchOfficialGittensorMiner } from "../gittensor/api";
-import { clampNumber, estimateNeurons, utcDayStartIso } from "../services/ai-review";
+import { BEST_REVIEW_MODELS, RELIABLE_FALLBACK_MODELS, clampNumber, estimateNeurons, utcDayStartIso } from "../services/ai-review";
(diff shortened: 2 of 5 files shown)


JSONbored:
Confirmed — good catch. `verifyUnlinkedIssueMatch`'s actual calls never wrote to `ai_usage_events`, so the budget gate this PR added was reading a counter this feature never contributed to (it respected every other feature's spend but was a free rider on its own).

Fixed in e053ed0: records the same worst-case per-candidate estimate the budget check itself uses into the shared counter, right alongside the existing rate-ceiling attempt recording. Added tests confirming the counter increases after a verification call (including with an unknown author) and that a write failure there is swallowed.

PR #4555, opened 2026-07-10
fix(db): resolve migration number collision at 0131

JSONbored:
## Summary

`db:migrations:check` currently fails on `main`: two already-merged PRs each independently claimed `migrations/0131` —

- #4544 (`fix(review): cache impact-map's per-file vector query results`) — `0131_impact_map_query_cache.sql`
- #4545 (`feat(review): viewport x theme completeness matrix for the screenshot-table gate`) — `0131_screenshot_table_gate_matrix.sql`

Both merged with the same number, so the sequence is no longer contiguous. This renumbers the more-recently-merged file (`impact_map_query_cache`, confirmed via `git log`) to `0133` — the next free slot — restoring a contiguous, collision-free sequence. No column/table/logic changes; a pure rename (`git mv`), and no test or source file references the old filename or number.

Discovered while working on an unrelated feature branch and fixed as its own small, dedicated change per repo convention (own commit for discovered drift, not folded into unrelated feature work).

## Test plan

- [x] `npm run db:migrations:check` — now reports a clean, contiguous `0001..0133` sequence
- [x] `npm run db:schema-drift:check` — clean
- [x] `npx tsc --noEmit` — clean


JSONbored:
Superseded by #4554, which merged the identical fix (renumbering `migrations/0131_impact_map_query_cache.sql` -> `0132_impact_map_query_cache.sql`) before this PR did. This branch is now stale against `main` and would have no effect (or regress newer migrations) if merged. Closing.

Issue #3611, opened 2026-07-05
Enable + validate automated visual capture end-to-end for metagraphed


JSONbored:
Part of #3607. Depends on #3608, #3609, #3610.

## Motivation
Once self-host rendering, preview-URL discovery, and per-repo route mapping are done, the feature still needs to actually be turned on and verified against real traffic before it's trustworthy enough to rely on instead of contributor-submitted screenshots.

## Requirements
- Set `GITTENSORY_REVIEW_SCREENSHOTS=true` on edge-us-01 (metagraphed is already on the `GITTENSORY_REVIEW_REPOS` cutover allowlist — no change needed there).
- Open a real (or deliberately test) PR against metagraphed touching `apps/ui/**` and confirm: the before/after table renders in the gittensory-orb comment, both desktop + mobile shots are present, the "after" shot reflects the actual PR change (not a stale/placeholder image), and a failed/pending preview degrades to the documented placeholder cards rather than a broken image or a thrown error.
- Confirm capture failures are fully fail-safe under real self-host conditions (network timeouts to the browserless service, etc.), not just in unit tests — a broken screenshot must never block or hold an otherwise-mergeable PR (already a design invariant of `buildCapture`/`capturePage`; verify it holds in practice).
- Monitor cost/latency impact for a week of real traffic before calling this done — Browser Rendering (or its self-host chromium equivalent) is explicitly called out in the existing code comments as the most expensive part of this pipeline.

## Deliverables
- [ ] Flag enabled on edge-us-01
- [ ] At least one real metagraphed PR with a verified, correct before/after table
- [ ] A documented rollback plan (flip the flag back off) if something goes wrong
- [ ] A week of production observation with no regressions before calling this shipped

## Expected outcome
Metagraphed contributors touching `apps/ui/**` get automated before/after visual evidence with no action required on their part — the original ask, fully live.

## Effort
S — mostly verification/observation once the prerequisite sub-issues land, not new engineering.

JSONbored:
Update 2026-07-10, post JSONbored/gittensory#4564 (\`review.visual.production_url\` per-repo override + generalized route inference beyond \`apps/gittensory-ui/\`).

[...]

PR #4648, opened 2026-07-10
fix(review): narrow visual screenshot path gate

JSONbored:
### Motivation
- The previous visual-path classifier marked every `apps/*` file as "visual", which caused backend/config/docs files under app folders to trigger expensive browser screenshot captures.
- The intent is to restrict screenshot capture to truly web-visible changes (public assets and frontend source files) while preserving route-file inference for frontend apps.

### Description
- Removed the broad `/^apps\/[^/]+\//i` pattern from `VISUAL_PATTERNS` in `src/review/visual/paths.ts` so only `public/**` and web-visible frontend extensions (e.g. `.tsx`, `.jsx`, `.css`, `.html`, `.svg`, `.astro`, `.vue`, `.svelte`, `.mdx`) are considered visual.
- Left `DEFAULT_ROUTE_FILE` and `mapFilesToRoutes` behavior in `src/review/visual/capture.ts` unchanged so frontend route files in `apps/*/src/routes/*.tsx|jsx` are still mapped via extension-based matching and route inference remains available.
- Updated `test/unit/visual-paths.test.ts` to reflect the narrowed gate and added regression assertions that backend/config/docs files such as `apps/api/src/server.ts`, `apps/ui/components.json`, and `apps/marketing-site/README.md` do not trigger the visual gate.

### Testing
- Ran `git diff --check` which passed with no whitespace or conflict markers.
- Ran targeted unit tests with `npx vitest run test/unit/visual-paths.test.ts test/unit/visual-capture.test.ts --testNamePattern "isVisualPath|mapFilesToRoutes app-folder generalization"`, and the test files executed successfully (unit assertions for `isVisualPath` and `mapFilesToRoutes` passed).
- `npm run typecheck` failed in this environment due to missing optional/local packages (`@anthropic-ai/claude-agent-sdk`, `web-tree-sitter`) causing TypeScript resolution errors.
- `npm run test:coverage` could not complete in this environment because coverage remapping errored with `TypeError: jsTokens is not a function`, so full coverage verification remains blocked here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5099b72db4832c8c139b5e188becd8)

### DIFF
--- a/src/review/visual/paths.ts
+++ b/src/review/visual/paths.ts
@@ -1,17 +1,17 @@
 // Visual-path classifier (reviewbot→gittensory convergence — visual capture port).
 //
 // PORTED VERBATIM from reviewbot's src/agents/gittensory/capabilities.ts `isVisualPath` (the three
-// VISUAL_PATTERNS), with the first pattern's app-folder segment widened to a wildcard (#3611 follow-up) so it
-// isn't gittensory-ui-only — see capture.ts's DEFAULT_ROUTE_FILE for the same generalization. This is the
-// EMPHATIC gate: screenshots fire ONLY for WEB-VISIBLE changes — any frontend app folder (apps/*/**, e.g.
-// apps/gittensory-ui/** or apps/ui/**), a public asset (public/**, e.g. an OG image), or a front-of-house
-// source extension (.tsx/.jsx/.css/.scss/.sass/.less/.html/.svg/.astro/.vue/.svelte/.mdx). A backend change
+// VISUAL_PATTERNS), keeping capture scoped to file types and public assets that are directly web-visible.
+// Route inference is still generalized in capture.ts's DEFAULT_ROUTE_FILE, but this cost gate must not treat
+// every apps/* file as visual: screenshot-enabled repos often keep backend/config/docs files there too. This is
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this reverts deliberate, already-shipped behavior from #4564 (merged the day before this PR was opened).

`e063f55ff` intentionally widened `VISUAL_PATTERNS`'s app-folder pattern from `/^apps\/gittensory-ui\//i` to `/^apps\/[^/]+\//i` specifically so metagraphed's `apps/ui/**` (a different app-folder name using the same TanStack flat-file routing convention) gets web-visible capture coverage — see #3611. That commit even added test cases pinning `apps/gittensory-ui/public/og.png → true` and `apps/ui/components.json → true`.

This PR removes that pattern entirely and flips those exact same test assertions to `false`, undoing the fix rather than fixing a real bug. The "every apps/* file marked visual, causing expensive captures" framing in the PR description is describing the intended outcome of #4564, not a defect.

Thanks for the scan — the underlying "is this gate too broad" question is worth watching, but the correct fix (if capture cost becomes a real problem) would need to preserve `#3611` coverage, not drop it.

PR #4652, opened 2026-07-10
fix(notifications): mark recaps after successful generation

JSONbored:
### Motivation
- Prevent a transient failure during scan/generation/delivery from consuming the per-period dedup marker and suppressing legitimate retries.  
- Ensure delayed queue messages processed after midnight deduplicate against the intended scheduled UTC date rather than processing-time.  

### Description
- Replace the pre-scan conditional `claimMaintainerRecapPeriod` flow with a read-before/mark-after-success approach by adding `getLastMaintainerRecapPeriodKey` and `markMaintainerRecapPeriodSent` in `src/db/repositories.ts`.  
- Change `runMaintainerRecapJob` to consult `getLastMaintainerRecapPeriodKey` (skip if already completed) and to call `markMaintainerRecapPeriodSent` only after the report is generated and the `maintainer_recap_generated` audit event is recorded in `src/review/maintainer-recap-wire.ts`.  
- Thread a `scheduledPeriodKey` through the cron enqueue, job `JobMessage` type, and the queue processor so scheduled jobs carry the intended UTC date (`src/index.ts`, `src/types.ts`, `src/queue/processors.ts`).  
- Add and update unit tests to cover the completed-period marker, the regression where a failed scan must not mark the period complete, and the scheduled-period-key behavior (`test/unit/db-parsers.test.ts`, `test/unit/maintainer-recap-wire.test.ts`, `test/unit/index.test.ts`).  

### Testing
- Ran the focused unit suites with `npx vitest run test/unit/maintainer-recap-wire.test.ts test/unit/db-parsers.test.ts` and both suites passed.  
- Ran TypeScript checks with `npm run typecheck` which completed successfully.  
- Ran repository hygiene checks with `git diff --check` which passed locally.  
- Attempted a targeted `index` test invocation (`npx vitest run test/unit/index.test.ts -t "maintainer recap"`) which timed out in this environment's scheduled-worker harness.  
- `npm audit --audit-level=moderate` could not complete in this environment due to the registry audit endpoint returning `403 Forbidden`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a501e7950d8832c865081df7e8da872)

### DIFF
--- a/src/db/repositories.ts
+++ b/src/db/repositories.ts
@@ -2536,23 +2536,29 @@ export async function claimRegateFanoutSlot(env: Env, now: string, windowMs: num
   }
 }
 
-/** Atomic per-period dedup for the cross-repo maintainer recap digest (#2249): claim `periodKey` (the current
- *  UTC date, "YYYY-MM-DD") as the singleton's last-sent period. Mirrors {@link claimRegateFanoutSlot}: the
- *  conditional UPDATE matches only when the stored period is unset or DIFFERENT from `periodKey`, so a retried
- *  cron tick or a redelivered (at-least-once) queue message for the SAME period gets 0 changes and skips
- *  before any repo scan or Discord send. Fail-open on a driver error (return true → the digest still runs,
- *  degrading to the pre-dedup behaviour rather than silently going dark). */
-export async function claimMaintainerRecapPeriod(env: Env, periodKey: string): Promise<boolean> {
+/** Read the last successfully generated maintainer recap period (#2249). Fail-open on a driver error
+ *  (`null` → the digest still runs), matching the notification job's availability-first posture. */
+export async function getLastMaintainerRecapPeriodKey(env: Env): Promise<string | null> {
   try {
-    const result = await env.DB.prepare(
-      "UPDATE global_agent_controls SET last_recap_period_key = ?1 WHERE id = 'singleton' AND (last_recap_period_key IS NULL OR last_recap_period_key != ?1)",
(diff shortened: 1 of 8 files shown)


JSONbored:
Closing — this trades away the atomicity that makes the dedup safe under Cloudflare Queues' at-least-once delivery.

`f8bd6f798` (merged the day before) deliberately used an atomic conditional `UPDATE ... WHERE (last_recap_period_key IS NULL OR last_recap_period_key != ?1)` — mirroring `claimRegateFanoutSlot`'s established pattern in this codebase — specifically so a retried cron tick or a redelivered queue message for the same period is a guaranteed no-op *before* any repo scan or Discord/Slack send happens.

This PR replaces that with a read-then-mark-after-success pattern (`getLastMaintainerRecapPeriodKey` read, then `markMaintainerRecapPeriodSent` only after the digest completes). The queue consumer runs with `max_concurrency: 3`, so two concurrent redelivered invocations for the same period can both read "not yet sent" before either has marked — both proceed to scan and send, producing duplicate digests. That's the exact scenario the original design's docstring named and closed.

The failure mode this PR is trying to fix (a failed send after claim never retries within that period) is real, but the fix should preserve the atomic claim-before-work guard rather than removing it — e.g. keep the atomic claim but add a separate completion/failure marker, rather than moving the single marker to after-the-fact. Thanks for flagging the lockout-on-failure edge case; happy to see a reworked version that keeps both properties.

Issue #4522, opened 2026-07-09
feat(miner): synthesize PreToolUse deny-hook rules from the review stack's own per-repo close/blocker finding-category history


JSONbored:
## Context

A maintainer review-stack × Autonomous Miner System (AMS) integration audit (2026-07-09) found that the miner's PreToolUse deny-hook rules are a static, hand-authored rule set, even though the review stack already has real per-repo history of what actually gets blocked or closed that could keep those rules current automatically.

## Evidence

- `packages/gittensory-miner/lib/deny-hooks.js:124-141` — `DEFAULT_DENY_RULES` is a hand-authored, hardcoded array; `packages/gittensory-miner/lib/deny-check.js` (the CLI wrapper) only ever calls `evaluateDenyHooks()` with the default rules — no loader for external/dynamic rules exists.
- The taxonomy that actually matters for this synthesis is `blockerCodes`/`scoreBlockers` (`src/rules/advisory.ts`'s `GateCheckEvaluation`, persisted via `blocker_codes_json` in `src/db/schema.ts`) — the review stack's own real, per-repo history of what blocks or closes a PR.
- That blocker-code history is already piped to the miner today, but only as **display text**: `src/services/miner-dashboard-recommendations.ts` renders it as a human-readable "Rerun after validation blockers change: ..." string for a dashboard, not as a machine-enforceable rule.
- Issue #2343 ("PreToolUse-hook-enforced house rules," still open) only asks for rules "sourced from a single, auditable list" — no dynamic-derivation requirement — confirming this synthesis job doesn't exist and isn't yet in scope anywhere.

## Why this matters for the Autonomous Miner System

Gittensory already classifies every review finding and has per-repo history of what blocks a PR (gate-advisory blockers) and what gets a PR closed. Today there's no path from "this repo's live gate has blocked N PRs for touching `CHANGELOG.md`" or "for missing a linked issue" to an automatically-refreshed local deny-hook rule the miner's coding agent is bound by *before* it ever attempts the violating action. This is a differentiator no review-only competitor could build, since it requires controlling both the autonomous submitter's own tool-call interception layer and the PR review side of the same pipeline — gittensory already has both halves, just no synthesis job connecting them.

## Deliverables

- [ ] Build a job that derives candidate deny-hook rules from a repo's own blocker-code/close-reason history (e.g. "N closes for touching path X" → a path-based deny rule), shaped to match `DEFAULT_DENY_RULES`'s existing rule format
[...]

Issue #4267, opened 2026-07-08
feat(miner-hands): Agent-SDK CodingAgentDriver (query() loop)


JSONbored:
A second implementation of the `CodingAgentDriver` interface (#4262, must land first), this one driving the coding agent in-process via the Claude Agent SDK's `query()` function rather than shelling out to a CLI binary. Verified via `grep -rn "claude-agent-sdk\|@anthropic-ai" package.json packages/*/package.json` — zero matches. Neither `@anthropic-ai/claude-agent-sdk` nor any `@anthropic-ai/*` package is a dependency anywhere in this repo today (root `package.json`, `packages/gittensory-engine/package.json`, `packages/gittensory-mcp/package.json`, `packages/gittensory-miner/package.json` are all clean). This issue adds the dependency.

## Deliverables
- [ ] Add `@anthropic-ai/claude-agent-sdk` as a dependency of the package hosting this driver (likely `packages/gittensory-engine`, matching #4262's `src/miner/` home) and wire an implementation around its `query()` async-iterable loop.
- [ ] Map `query()`'s streamed message/tool-use events into the `CodingAgentDriver` result shape from #4262 — success/failure, changed files, transcript/usage summary — without leaking SDK-specific event types into the shared interface.
- [ ] Respect the same scoped-working-directory and acceptance-criteria-file inputs #4262's interface defines, so a caller (the iterate-loop orchestrator in maintainer-only issue #2333) can swap this driver for the CLI-subprocess driver (#4266, this batch) with no caller-side changes — that interchangeability is the entire point of the shared interface.
- [ ] A tool-call interception point compatible with what the `PreToolUse` hook maintainer-only issue #2343 describes ("a PreToolUse hook registered against the CodingAgentDriver's underlying session that intercepts every tool call BEFORE execution") — #2343 names this driver's SDK session as its intended hook attachment point, so the SDK's own hook/callback surface needs to be reachable from outside this module, not fully encapsulated.
- [ ] Tests using the SDK's own test/mock affordances (or a thin injected wrapper around `query()` if none exists) so CI never makes a real model call — mirroring the injected-`SpawnFn` testability convention #4262 establishes for the CLI-subprocess driver.

## References
- Verified via `grep -rn "claude-agent-sdk\|@anthropic-ai" package.json packages/*/package.json` — zero hits repo-wide, confirming this is a net-new dependency
- Depends on #4262 (this batch) — the `CodingAgentDriver` interface this implements
[...]

Issue #2341, opened 2026-07-01
maintainer: global + per-repo kill-switch


JSONbored:
An operator (or the contributor running their own miner) needs one unambiguous way to halt ALL autonomous write activity immediately — analogous to the existing review-stack's global kill-switch (`AGENT_ACTIONS_PAUSED`) but for the miner's own local runtime, plus a per-repo variant so a single misbehaving target can be paused without stopping the whole fleet.

Maintainer-owned: this is the actual trigger surface the Governor consults, and the roadmap treats the kill-switch as core safety-critical infrastructure.

## Deliverables
- [ ] A global kill-switch (env var or a well-known local config flag, e.g. `GITTENSORY_MINER_KILL_SWITCH=true`) that the Governor chokepoint checks FIRST, before any other calculator
- [ ] A per-repo kill-switch resolvable from `.gittensory-miner.yml` (MinerGoalSpec) that halts only that repo's queue items while leaving others running
- [ ] Kill-switch state changes are themselves recorded to the governor ledger (who/what tripped it, when)
- [ ] Tests: global switch denies regardless of per-repo state, per-repo switch denies only its own repo's items, and toggling the switch off resumes exactly where the queue left off (no lost/corrupted in-flight items)

## References
- `src/settings/agent-execution.ts` — the existing review-stack global kill-switch (`AGENT_ACTIONS_PAUSED`) precedent to mirror conceptually for the miner's own runtime
- The Governor chokepoint (companion maintainer issue) — this kill-switch is one of the inputs it consults first
- `.gittensory-miner.yml` (new MinerGoalSpec config file, per the architecture skeleton)

Issue #2333, opened 2026-07-01
maintainer: local create->score->self-review iterate-loop orchestrator (the control-flow core)


JSONbored:
This is the actual autonomous LOOP that Phase 3 exists to build: given a plan/acceptance-criteria packet (from Phase 2, out of scope here) and a resolved `CodingAgentDriver`, repeatedly invoke the driver, self-review the resulting diff against `src/rules/predicted-gate.ts` (the byte-identical predicted-gate target named as THE self-review target for an autonomous miner) plus the acceptance-criteria file, and DECIDE — autonomously, with no human in the loop at this stage — whether to keep iterating, stop and hand off to Phase 4 submission, or abandon the attempt. This decision logic is the precise chokepoint the fixed architecture skeleton's safety-tier system reserves for the owner: it is the trigger surface for "does the system keep trying, or does it eventually open a PR" without a human approving each step.

Tagged `maintainer` per this phase's explicit safety split: the driver seam, drivers, and harness primitives (all prior issues in this batch) are reviewable, no-autonomous-trigger contributor work; this orchestration control flow is not — it is adjacent to the #1 slop-at-scale strategic risk (an autonomous fleet maximizing gate-pass rate can mass-produce gate-passing-but-low-value PRs) and the owner has stated intent to build/review this personally.

## Deliverables
- [ ] `packages/gittensory-engine/src/miner/iterate-loop.ts` implementing the create->score->self-review->decide control flow, consuming a `CodingAgentDriver` + the predicted-gate self-review target
- [ ] A bounded max-iteration/max-cost ceiling enforced INSIDE the loop itself (not left to an external caller to remember to enforce)
- [ ] Every iteration's decision (continue / stop-success / abandon) is recorded via the attempt-log primitive (from the wave-2 batch) so the decision trail is auditable after the fact
- [ ] Default behavior when the loop cannot conclusively determine pass/fail (e.g. self-review itself errors) is to ABANDON, never to optimistically hand off to submission — downgrade-only on ambiguity
- [ ] Owner-reviewed test coverage proving the loop never hands off to Phase 4 submission on a predicted-gate FAIL or an ambiguous self-review result

## References
- `/Users/shadowbook/Documents/gittensory/.claude/worktrees/quizzical-engelbart-34f3be/src/rules/predicted-gate.ts` (263 lines) — the byte-identical self-review target this loop must consult every iteration; see especially `buildPredictedGateVerdict`'s `conclusion`/`blockers` fields
[...]

PR #5031, opened 2026-07-11
feat(miner-governor): PreToolUse-hook-enforced house rules (#2343)

### LINKED ISSUE #2343: maintainer: PreToolUse-hook-enforced house rules (deny even under bypassPermissions)
When the miner's CodingAgentDriver runs a coding agent (e.g. Claude Agent SDK) that itself has arbitrary tool-use ability, prompt-level instructions alone are not a safety boundary — a coding agent can be steered around them, and a bypassPermissions-configured session ignores interactive approval prompts entirely. This issue builds a PreToolUse-hook-enforced denial layer that runs OUTSIDE the coding agent's own reasoning, so house rules (no secrets in diffs, no writes outside the target repo's worktree, no execution of the Governor's OWN write specs without going through the chokepoint) hold even when the underlying agent session has bypassPermissions set.

This was flagged in the roadmap to be "pulled forward from Phase 0/3 if not already started there" — check for any existing scaffold before starting; if none exists, this issue builds it from scratch. Maintainer-owned: this is a hard security boundary, not a bookkeeping task.

## Deliverables
- [ ] A PreToolUse hook registered against the CodingAgentDriver's underlying session that intercepts every tool call BEFORE execution
- [ ] The hook denies (not just warns) any tool call matching a configured house-rule denylist, and this denial is enforced identically whether or not the session was started with bypassPermissions
- [ ] House rules are sourced from a single, auditable list (not scattered inline checks) so they can be reviewed as one unit
- [ ] Tests: a denied tool call is blocked under normal permission mode AND under bypassPermissions, an allowed tool call passes through unmodified, and a hook-internal error fails closed (denies) rather than silently allowing
- [ ] Every denial is recorded to the governor ledger with the specific rule that fired

## References
- `src/selfhost/ai.ts` + `src/selfhost/ai-config.ts` — the existing SelfHostAi/AI_PROVIDER seam this driver mirrors (cited as the template for the whole CodingAgentDriver seam)
- The Governor chokepoint + append-only ledger (companion maintainer/foundation issues) — this hook's denials feed the same ledger
- Check `packages/gittensory-miner/` for any Phase-0-scaffolded hook config before starting; if absent, this issue is the first implementation

JSONbored:
## Summary

- This is a hard security boundary, not a bookkeeping task, per the issue's own framing: prompt-level instructions alone are not a safety boundary for a coding agent with arbitrary tool-use ability, and a `bypassPermissions`-configured session ignores interactive approval prompts entirely.
- **Checked for existing scaffold first, per the issue's explicit instruction — and found two directly relevant pieces already shipped:**
  - `packages/gittensory-miner/lib/deny-hooks.js` (#2295) — a pure `evaluateDenyHooks(toolCall, rules)` decision function plus `DEFAULT_DENY_RULES` (blocks CI workflows, `.env*`, `.dev.vars`, `.npmrc`, secret-bearing paths, private keys/PEM material, force-push). Its own header comment says: *"There is NO live tool-call interception in this phase — a later phase's real coding-agent driver plugs an event source into `evaluateDenyHooks`."*
  - `packages/gittensory-engine/src/miner/agent-sdk-driver.ts` (#4267) already threads an `AgentSdkHooks` passthrough onto the Claude Agent SDK's `query()` options, with a comment naming this exact issue: *"callers pass `hooks` (e.g. a `PreToolUse` matcher, #2343's stated attachment point) and this driver forwards them verbatim."*
  - `packages/gittensory-miner/lib/deny-hook-synthesis.js` (#4522) derives rule *proposals* from blocker history; its header explicitly says it *"Feeds the consumption surface #2343 will wire into `evaluateDenyHooks` — this issue owns derivation + audit, not live hook interception."*
  - So this PR's actual job — confirmed by three separate prior issues all pointing at it — is exactly the missing piece: the live interception wiring itself, not the rule logic (already built) or the rule-derivation logic (already built).
- **Verified the "even under `bypassPermissions`" claim against Anthropic's own current documentation** (`code.claude.com/docs/en/agent-sdk/permissions`) rather than assuming: hooks are evaluated **first**, before deny rules, ask rules, the permission-mode check, and allow rules. Quoting directly: *"Bypass permissions mode... Hooks still execute and can block operations if needed."* and *"Deny rules (`disallowed_tools`), explicit `ask` rules, and hooks are evaluated before the mode check and can still block a tool."* This module does not implement that guarantee — the SDK does. This module's only job is to return a correctly-shaped, fail-closed deny decision every time.
- New `packages/gittensory-miner/lib/pretooluse-hook.js` exports `buildHouseRulesPreToolUseHook(config, options)`, which returns a real Agent SDK `PreToolUse` hook callback: translates the SDK's `{ tool_name, tool_input }` input shape into `evaluateDenyHooks`'s `{ name, input }` shape, returns the exact documented `hookSpecificOutput: { hookEventName: "PreToolUse", permissionDecision: "deny", permissionDecisionReason }` shape on a match (or `{}` to allow), records every denial to the governor ledger with the specific rule's reason, and fails closed (denies) on any internal error — including a ledger-append failure, which must never downgrade a security decision to allow.
- House rules are sourced from the single, already-auditable `DEFAULT_DENY_RULES` list by default, or any caller-supplied effective rule set (e.g. `resolveEffectiveDenyRules` from `deny-hook-synthesis.js`, merging in maintainer-approved synthesized rules) — this module composes whatever rule set it's given; it does not own deriving one.
- Also fixes a small pre-existing gap noticed in passing: `deny-hook-synthesis.js` was missing from the miner package's `node --check` build-validation chain (added alongside the new file, in the same line I was already touching).

## Scope

- [x] Focused: one new file + its test + a two-line build-script fix for a file directly in this issue's own dependency chain.
- [x] Follows `CONTRIBUTING.md`.
- [x] Owner PR — linked issue eligibility gate does not apply. Closes #2343.
- [x] No dependency on the other open Miner Wave 3 PRs — this branches directly off `main` and only imports pre-existing `deny-hooks.js`/`governor-ledger.js`.

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck` — clean.
- [x] `npm run build` in `packages/gittensory-miner` (`node --check`, including the new `pretooluse-hook.js` and the now-checked `deny-hook-synthesis.js`) — clean.
- [x] `vitest run` on the new suite plus every deny-hooks-adjacent suite (`miner-pretooluse-hook`, `miner-deny-hooks`, `deny-hooks-fixtures`, `miner-cli-deny-check`, `miner-deny-hook-synthesis`, `miner-governor-ledger`) — 64/64 passing.
- [ ] `npm run test:coverage`, `test:workers`, `build:mcp`, `test:mcp-pack`, `ui:*`, `npm audit` — not re-run locally; this PR touches only `packages/gittensory-miner/**` and `test/unit/**`, and per `CONTRIBUTING.md` only `src/**` is Codecov-measured. CI runs them authoritatively.

## Safety

- [x] No secrets, wallet/hotkey/trust-score/reward data anywhere.
- [x] Fail-closed is tested, not just asserted: one test forces a real exception through the actual `evaluateDenyHooks` call (a `Proxy` whose `matcher` getter throws, rather than a test-only seam added to production code) and confirms a deny; another forces the governor-ledger `append` itself to throw and confirms the deny still happens.
- [x] Reuses the SDK's own documented deny/allow output shape exactly (verified against current docs, not recalled from training data) — a malformed `hookSpecificOutput` shape would silently fail to block anything, which would be far worse than an obviously-broken build.
- [x] This PR builds the interception primitive but does not wire it as a default anywhere — no current call site constructs a live coding-agent attempt end-to-end yet (that's the not-yet-built orchestration loop, #2333/#2337). `driver-factory.ts`'s existing optional `hooks` parameter is the natural attachment point once that loop exists; forcing a default wiring now would mean editing a file with no real caller yet, which isn't this issue's job.
- [x] No UI changes. No docs/schema changes needed (no new config surface).

## Notes

- Unrelated to my in-flight #2341/#2342/#2340 stack — deliberately branched fresh off `main` rather than stacked, so it can be reviewed and merged independently.


JSONbored:
Pushed a coverage-only follow-up commit (db8710322): measured coverage (vitest v8, scoped include override since this file sits outside the repo's default \`coverage.include\`) found \`pretooluse-hook.js\` at 76.92% branch.

Four of the six gaps were real and got tests:
- The \`input && typeof input === "object"\` boundary check (input is typed \`unknown\`) never saw a malformed non-object input.
- A custom rule omitting \`reason\` never exercised the generic-message fallback.
- A deny with a non-string \`tool_name\` (\`matcher: "*"\` still fires) never exercised the ledger payload's null-toolName fallback.
- The internal-error catch's \`error instanceof Error\` check was only ever exercised with real Error instances, never the \`String(error)\` fallback.

The remaining two (both on \`blockedBy?.matcher ?? null\`) were dead code, not undertested: \`ruleMatches\` gates every match on \`typeof rule.matcher === "string"\`, so a rule can never become \`blockedBy\` without a defined string matcher -- the fallback could never fire. Simplified to \`verdict.blockedBy.matcher\` rather than writing a contrived test to hit unreachable code.

\`pretooluse-hook.js\` now measures 100/100/100/100 (stmts/branch/funcs/lines). 56/56 tests pass across this file and its siblings (deny-hooks, deny-hook-synthesis, deny-hooks-fixtures).

Issue #2338, opened 2026-07-01
maintainer: closed-loop discovery re-entry trigger


JSONbored:
The final piece that makes the whole runtime genuinely AUTONOMOUS/REPEATING rather than a one-shot tool: on a resolved outcome (merged, or rejected-and-disengaged), automatically re-invoke Phase-1 discovery to select the next candidate, consuming the read-only `LoopSummary` object (built in the sibling loop-summary issue) as a feedback signal. This is precisely the "fully-closed discovery-loop re-entry" the phase brief calls out as maintainer-only: it is what turns single runs into an unattended repeat loop, and a bug here (e.g. re-entering too fast, ignoring a circuit-breaker, or looping on a permanently-rejected repo) is the top slop-at-scale risk for the whole miner subsystem.

## Deliverables
- [ ] `packages/gittensory-miner/src/manage/loop-reentry.ts` — on a terminal outcome event, calls into Phase-1 discovery with the `LoopSummary` as input, selecting (or explicitly declining to select) the next candidate
- [ ] A hard rate/session cap (max re-entries per hour/session, configurable, conservative default) enforced independently of any per-repo circuit-breaker
- [ ] Never re-enters on a `disengaged` (rejected) run without the loop-summary's repeated-blocker tally being consulted first — a repo with N consecutive rejections is auto-paused for that repo, not retried
- [ ] Every re-entry decision appended to the event ledger with the LoopSummary that drove it, for full auditability
- [ ] Integration test: merged outcome → re-entry fires once; rejected outcome with high repeated-blocker tally → re-entry is suppressed and the repo is paused
- [ ] Manual owner sign-off before enabling by default in any profile

## References
- `packages/gittensory-miner/src/manage/loop-summary.ts` (new, from the loop-summary issue) — the feedback object this consumes
- `packages/gittensory-miner/src/manage/rejection-state-machine.ts` (new, from the rejection-state-machine issue) — the terminal `disengaged` state this respects
- `src/settings/autonomy.ts` (86 lines) — deny-by-default pattern precedent for the rate/session cap
- New: `packages/gittensory-miner/src/manage/loop-reentry.ts`

Issue #2336, opened 2026-07-01
maintainer: gated-submission trigger — predicted-gate-PASS + slop-under-threshold check before calling open_pr


JSONbored:
THE safety-critical chokepoint of Phase 4: the actual code path that decides "call `gittensory_open_pr` NOW" for an autonomous run. This must conservatively require BOTH a predicted-gate PASS (`src/rules/predicted-gate.ts` / `gittensory_predict_gate`) AND a slop score under a configurable threshold (`gittensory_check_slop_risk`) before the local-write action spec from `buildOpenPrSpec` (`src/mcp/local-write-tools.ts`) is ever invoked. Any ambiguity resolves to NOT submitting. This is exactly the kind of autonomous-write trigger the safety-tier system reserves for maintainer-only work — it is the single most consequential decision point in the whole miner runtime.

## Deliverables
- [ ] `packages/gittensory-miner/src/manage/submission-gate.ts` — `shouldSubmit(candidate) -> { allow: boolean; reasons: string[] }` requiring predicted-gate `conclusion === "merge"`-equivalent AND slop band below the configured threshold
- [ ] Fail-closed on any missing/errored signal (predictor unreachable, slop check errored) — never defaults to allow
- [ ] The actual call site invoking `buildOpenPrSpec` / `gittensory_open_pr` is gated exclusively through this function — no other code path may call it
- [ ] Exhaustive unit tests: pass/pass allows, fail/pass blocks, pass/fail blocks, both-fail blocks, and every fail-closed error path
- [ ] A dry-run/observe mode flag (mirroring `src/settings/autonomy.ts`'s AUTONOMY_LEVELS) that logs the decision without ever calling the write path, for safe rollout

## References
- `src/rules/predicted-gate.ts` (263 lines) — the predicted-gate verdict this reads
- `src/mcp/server.ts:1198` (`gittensory_check_slop_risk`) — the slop-risk check this reads
- `src/mcp/local-write-tools.ts:32` (`buildOpenPrSpec`) — the local-write action spec this gates access to
- `src/settings/autonomy.ts` (86 lines) — the autonomy-dial pattern (`isActingAutonomyLevel`, deny-by-default) to mirror for the dry-run gate
- New: `packages/gittensory-miner/src/manage/submission-gate.ts`

Issue #2334, opened 2026-07-01
maintainer: self-review adapter wiring iterate-loop diffs through predicted-gate + slop scoring


JSONbored:
Build the adapter that takes a driver's raw diff/worktree state each iteration and turns it into the SAME inputs `buildPredictedGateVerdict` and the slop/scoring signals expect (a synthetic PR record, changed paths, title/body), so the iterate loop's self-review call is genuinely byte-identical to what the live maintainer gate would compute post-submission — not an approximation. This is the piece that makes the miner's own self-assessment trustworthy enough to decide autonomously whether to keep iterating, and is inseparable from the trigger-surface concern the orchestration-loop issue covers, so it is held to the same `maintainer` tier.

## Deliverables
- [ ] `packages/gittensory-engine/src/miner/self-review-adapter.ts` exporting a function that builds a `PredictedGateInput`-shaped object (repo, contributor login, title, body, labels, linked issues, changed paths) from an attempt's live worktree diff state
- [ ] Calls `buildPredictedGateVerdict` (imported from wherever the shared-engine extraction issue relocates it, or directly from `src/rules/predicted-gate.ts` if that extraction has not yet landed — coordinate with whichever issue owns moving deterministic logic into `packages/gittensory-engine`) and returns its verdict alongside a slop-signal pass using `src/signals/slop.ts`
- [ ] Explicitly threads `changedPaths` (the real changed file paths from the diff) so the path-dependent predicted-gate checks are evaluated, not skipped — an omitted-paths call silently under-predicts per the existing `PREDICTED_GATE_NOTE_NO_PATHS` disclaimer, which would be a dangerous false-confidence bug if the miner's own loop relied on it
- [ ] A hard requirement (asserted in tests) that a `conclusion` other than a clear pass is treated as "do not hand off to submission" by any caller (this issue, plus the orchestration-loop issue, both enforce this independently — defense in depth)
- [ ] Owner-reviewed tests using real predicted-gate fixtures (a genuinely passing synthetic diff and a genuinely blocked one) proving the adapter's verdict matches calling `buildPredictedGateVerdict` directly with the same inputs

## References
[...]

Issue #781, opened 2026-06-17
feat(agent): harness adapter ('connect a Claude/Codex subscription') + driving loop


JSONbored:
Parent epic: #770 (Phase 2).

## Goal
Extend init-client into a first-class **auto-dev profile** that wires the local harness (Claude Code/Codex/Cursor) to the gittensory MCP + miner-planner profile + a plan→implement→push driving loop.

## Deliverables
- [ ] auto-dev profile generator (per harness)
- [ ] Driving loop: decision_pack → reward_risk → pick action → local_branch plan → (harness implements) → N8 write-tools push
- [ ] Reward-aware: respect open-PR gate, credibility floor, time-decay

## Context
packages/gittensory-mcp/bin (init-client, agent profiles), src/services/decision-pack.ts, reward-risk.ts.

Milestone: M4.

PR #5054, opened 2026-07-11
feat(miner-governor): wire the gated-submission trigger into the driving loop's handoff signal (#2337)

### LINKED ISSUE #2337: maintainer: wire the submission-gate trigger into the harness driving loop
Connects the gated-submission decision function (built in the sibling submission-gate issue) to the ACTUAL harness driving loop (`src/selfhost/ai.ts`-style seam / the miner-auto-dev harness adapter referenced by #781) so that a real run can, end to end, decide to submit and invoke the local-write spec. This is the live actuation wiring itself — the exact trigger surface the safety-tier system reserves for maintainer review, since a bug here means an autonomous write happens when it should not have.

## Deliverables
- [ ] Harness driving-loop call site that, after a candidate fix is prepared, calls `shouldSubmit()` and ONLY on `allow: true` proceeds to build + surface the `open_pr` local-write spec to the contributor's own execution environment
- [ ] A hard circuit-breaker: N consecutive submission-gate `allow:false` decisions in one session pauses the run entirely pending human review (never silently loops trying to force a pass)
- [ ] Every decision (allow or block) is appended to the local event ledger with full reasons, for audit
- [ ] Integration test simulating a full candidate → gate-check → submit (or correctly-blocked) cycle
- [ ] Manual owner sign-off on the wiring before this ships to any default-on profile

## References
- `packages/gittensory-miner/src/manage/submission-gate.ts` (new, from the submission-gate issue) — the decision function this wires in
- `src/selfhost/ai.ts` (754 lines) — the SelfHostAi/AI_PROVIDER seam pattern the harness driving loop mirrors
- `packages/gittensory-miner/src/state/event-ledger.ts` (new, from the event-ledger issue) — the audit trail this writes to
- git history: `feat(agent): harness adapter — miner-auto-dev profile + driving loop (#781)` — the existing driving-loop this wiring attaches to

JSONbored:
## Summary

Advances #2337.

Adds `evaluateHarnessSubmissionTrigger` (`packages/gittensory-engine/src/miner/harness-submission-trigger.ts`) + `evaluateAndRecordHarnessSubmissionTrigger` (`packages/gittensory-miner/lib/harness-submission-trigger.js`): connects `shouldSubmit` (#2336) to the actual driving loop's own handoff signal -- iterate-loop.ts's (#2333) `HandoffPacket`, produced the moment a real run's self-review reaches a clean predicted-gate pass.

**Why "Advances" not "Closes"**: this issue's own text asks for a real run to, "end to end, decide to submit **and invoke the local-write spec**." This PR delivers the decision-wiring half genuinely, but does *not* build or invoke `buildOpenPrSpec` (`src/mcp/local-write-tools.ts`) -- that lives in the private root `src/` tree, unreachable from this portable package for the same cross-package-boundary reason `self-review-adapter.ts`'s slop injection exists (#2334). A real call site (root-side server/CLI integration, wired in a later issue) would consult this function's `allow: true` before it builds that spec itself. Flagging this explicitly rather than claiming full closure.

**The session-level circuit breaker** this issue's own deliverable calls for, distinct from `shouldSubmit`'s per-candidate signal checks: N consecutive `allow:false` decisions in one session pauses the run entirely pending human review. Checked FIRST, before ever consulting `shouldSubmit` -- once tripped, only a human clearing the session's own tally can un-trip it, unlike a per-candidate block a later, different candidate can clear on its own merits. Session-scoped (not per-repo), matching "pauses the run entirely" -- deliberately distinct from #2338's loop-reentry circuit breaker, which is per-repo (a rejection streak on one repo must not pause unrelated repos).

Every decision (allow or block) is appended to the local event ledger with full reasons, whether or not the circuit breaker was involved -- per this issue's own audit deliverable.

**On the referenced "harness driving loop"**: the issue's references section cites both the old M4-era `gittensory-mcp` harness adapter (#781, a *different* package, closed a while back under a different epic) and this session's own `submission-gate.ts`. Wired to `iterate-loop.ts`'s `HandoffPacket` instead of #781's system, since that's the actual, current, same-wave (#2333) driving loop #2336 was explicitly built to compose with -- reaching into the older `gittensory-mcp` package would be a much larger, differently-scoped undertaking than this maintainer-only issue's batch suggests.

## Validation

Measured, not assumed -- across both the engine's own `node:test` suite AND the actual codecov-visible root vitest path:

```
# engine package
npx tsc -p tsconfig.json --incremental false && npx tsc -p tsconfig.test.json --incremental false && node --experimental-test-coverage --test "dist-test/**/*.test.js"
# root, the actual codecov-visible path
npx vitest run test/unit/miner-harness-submission-trigger.test.ts --coverage --coverage.include="packages/gittensory-miner/lib/harness-submission-trigger.js"
npx vitest run test/unit/miner-harness-submission-trigger.test.ts --coverage --coverage.include="packages/gittensory-engine/src/miner/harness-submission-trigger.ts"
```

- Engine: 447/447 tests pass; `harness-submission-trigger.js` (compiled from the `.ts` source) 100.00% lines/branch/funcs.
- Root vitest (codecov's real measurement path): 5/5 tests pass; **both** the miner-lib wrapper and the engine policy independently confirmed at 100.00% stmts/branch/funcs/lines.
- `packages/gittensory-miner`'s `node --check` build script passes cleanly with the new file registered.

## Test plan

Per the issue's own explicit deliverable:
- [x] "Integration test simulating a full candidate -> gate-check -> submit (or correctly-blocked) cycle" -- both scenarios, named verbatim in the test file.
- [x] Circuit breaker: N consecutive blocks pauses even an otherwise-clean handoff; a single allow resets the streak.
- [x] Defense in depth: a handoff whose own verdict fails predicted-gate or exceeds the slop threshold is still blocked by `shouldSubmit`'s own re-check, not blindly trusted just because a handoff happened.
- [x] Observe mode forces `allow: false` regardless of the underlying signals.
- [x] Fails closed on every malformed-input/missing-dependency path.


JSONbored:
Pushed a coupling fixup (0eb3704a1) + rebased onto #5057 (#2339, kill-switch propagation): #2339 makes \`SubmissionGateCandidate.killSwitchScope\` required on \`shouldSubmit\`, which this PR's own \`evaluateHarnessSubmissionTrigger\` constructs internally -- threaded \`killSwitchScope\` through from this function's own candidate rather than duplicating the kill-switch check itself (\`shouldSubmit\`'s own guard covers it whenever the session circuit breaker hasn't already short-circuited first). Base retargeted to \`feat/miner-killswitch-manage-loop-2339\` so this PR's diff stays scoped to its own changes.

Also closed a real, pre-existing codecov gap discovered while verifying the fixup: \`submission-gate.ts\` had zero root-vitest scenarios for its null-predictedGateVerdict, null-slopAssessment, slop-exceeds-threshold, and observe-mode-would-have-blocked/would-have-allowed paths -- it's only reachable from the root vitest suite through this file's own integration test (\`test/unit/miner-harness-submission-trigger.test.ts\`), and that test never exercised them. Added all five scenarios; \`submission-gate.ts\` now measures 100% branch coverage via the actual codecov-visible path.

409 -> now includes 12/12 tests in the root integration test file, all three touched files (harness-submission-trigger.js, harness-submission-trigger.ts, submission-gate.ts) independently confirmed at 100% stmts/branch/funcs/lines.

Issue #5000, opened 2026-07-11
fix(orb): /v1/orb/token still returns 500 after the #orb-broker-500 fix (95 Sentry events)


JSONbored:
## Summary

`POST /v1/orb/token` (the Orb token-broker exchange) returns a bare HTTP 500 — 95 events over 10+ days.

## Evidence (Sentry)

[GITTENSORY-J](https://jsonbored.sentry.io/issues/GITTENSORY-J) — `orb_broker_unavailable: Orb broker token exchange failed (500).` First seen 2026-06-29, last seen 2026-07-09.

## Root cause — already investigated, this is NOT a simple "wrap in try-catch" fix

Sentry Seer's summary ("Wrap the entire `/v1/orb/token` handler in a top-level try-catch so unhandled stream/DB errors return 503 instead of a framework 500") **does not match the current code and must not be blindly applied**:

- The handler (`src/api/routes.ts:3311`) already wraps its `brokerOrbToken` call in try/catch, returning 503 on any thrown error.
- `readOrbRelayRegisterBody` (`src/orb/relay.ts:175`), the one call in this handler made *before* that try/catch, was **already hardened for exactly this failure mode** in commit `65de78ec6e7bf5350e217bc3aa4c62dda9911c88` (2026-07-05, tagged `#orb-broker-500` in its own doc comment) — its own comment literally describes fixing "a bare framework 500 instead of the route's own clean 4xx/503 JSON error response."
- That fix commit is confirmed to be an ancestor of every commit that has run on the self-host box since (verified via `git merge-base --is-ancestor`).
- **Yet this Sentry issue's last occurrence (2026-07-09) is 4 days *after* that fix was live in production.**

This means the 2026-06-29–07-03 events are likely the original, already-understood bug. The 2026-07-05–07-09 events (after the fix deployed) are either a **regression**, an **incomplete fix** (a different unhandled-throw path in `brokerOrbToken` itself, or in Hono's own routing/middleware layer before this handler is even reached), or evidence the fix's deploy didn't reach production as early as the commit timestamp suggests. Needs to be distinguished by re-checking Sentry's event timestamps against exact deploy windows, not assumed.

## Requirements

1. Confirm precisely which events (by timestamp) occurred before vs. after the 07-05 fix actually went live on the box.
2. If post-fix events remain: find the actual remaining unhandled-throw path (check `brokerOrbToken` itself and anything Hono-level ahead of this route) rather than re-applying the same already-shipped fix.
[...]

PR #5109, opened 2026-07-11
fix(auth): fail closed on rate limiter errors

JSONbored:
### Motivation
- A global rate-limit middleware was failing open when the Durable Object check threw, which unthrottled strict/expensive and unauthenticated endpoints during limiter outages and created an abuse-control bypass. 
- The change ensures the middleware preserves the global abuse-control boundary by returning a structured fail-closed response instead of allowing the request to proceed.

### Description
- Change `enforceRateLimit` in `src/auth/rate-limit.ts` to return a structured `503` JSON response `{ error: "rate_limiter_unavailable" }` with a `Retry-After: 5` header when the Durable Object `idFromName/get/fetch` path throws. 
- Preserve best-effort audit semantics so a failed `rate_limit.denied` audit write does not convert an already-denied `429` into a crash. 
- Update `test/unit/auth.test.ts` to assert the new fail-closed behavior for both `Error` and non-`Error` Durable Object failures and adjust test names/comments to match the structured response expectation.

### Testing
- Ran the targeted regression suite: `npx vitest run test/unit/auth.test.ts -t "REGRESSION (#5000)" --reporter=verbose`, which passed the updated regression checks. 
- Ran the full auth unit file: `npx vitest run test/unit/auth.test.ts --reporter=verbose`, which passed (all `auth.test.ts` cases passed locally). 
- Attempted full coverage with `npm run test:coverage` in this environment but the complete repo-wide coverage run could not be completed here due to unrelated long-running/timeout failures in other unit shards; recommend running the full local gate (`npm run test:ci` / `npm run test:coverage`) before pushing.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52358aa760832c917453bb36642740)

### DIFF
--- a/src/auth/rate-limit.ts
+++ b/src/auth/rate-limit.ts
@@ -66,13 +66,19 @@ export async function enforceRateLimit(c: Context<{ Bindings: Env }>, routeClass
       body: JSON.stringify({ key, ...config }),
     });
   } catch (error) {
-    // Fail OPEN (#5000): this middleware runs on every route ahead of the handler's own try/catch, and no
-    // app.onError is registered anywhere -- an uncaught Durable Object hiccup (eviction, migration, a
-    // rolling-deploy blip) previously escaped as Hono's bare, unstructured 500 for whatever route the caller
-    // happened to be hitting, indistinguishable from a real application bug in that route. The rate limiter
-    // exists to protect the app, not crash the request it's supposed to be gating.
+    // Fail closed with a structured response (#5000): limiter outages should not escape as Hono's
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing: reverses #5000's deliberate fail-open rationale for this middleware (it runs ahead of every route's own error handling, with no app.onError registered). Failing closed turns any transient rate-limiter Durable Object hiccup — eviction, migration, rolling deploy — into a structured 503 across every gated route instead of a brief unprotected window. Availability takes priority over abuse-control here; keeping fail-open.

Issue #4996, opened 2026-07-11
fix(review): RAG embedding calls rejected with HTTP 400 (485 Sentry events, ongoing)


JSONbored:
## Summary

RAG context-fetch embedding calls are rejected with HTTP 400 — 485 events, ongoing for 2 weeks, still firing this hour as of filing.

## Evidence (Sentry)

[GITTENSORY-D](https://jsonbored.sentry.io/issues/GITTENSORY-D) — `review_context_fetch_failed: Error: ai_embed_http_400`, first seen 2026-06-28, last seen 2026-07-11 (today, ongoing). Tags: `contextType: rag`, `ev: rag_embed_error`. No stack trace attached to the captured event (structured-log-forwarded, not a raw exception — see the companion observability-hygiene issue for the general pattern).

## Root cause (needs independent investigation — do not trust Seer blind)

Seer's summary for this issue says: "Add a circuit-breaker flag in enrichment-wire.ts: disable all /v1/enrich calls for the process lifetime when the startup probe confirms a 401/403 auth rejection." **This does not match the actual error** — the tagged error is `ai_embed_http_400` (Bad Request), not 401/403 (auth). A 400 usually means a malformed request (wrong payload shape, oversized input, bad model/param), not an auth problem. Seer's summary appears to have pulled context from a different, adjacent issue. Start from the actual embedding-request code (search for `rag_embed_error` / the RAG context-fetch path), not from this summary.

## Requirements

1. Find the actual embedding-call code path (`rag_embed_error` / `review_context_fetch_failed`) and reproduce or trace what request is actually being rejected with 400.
2. Determine whether this is a malformed-payload bug (fixable), an input-size limit being exceeded (needs truncation/chunking), or a genuinely bad config (model name, endpoint) — each has a different fix.
3. Since this has been failing continuously for 2 weeks, check whether RAG-backed review context has been silently degraded (missing context) for every review in that window, and whether that's visible anywhere it should have been caught sooner.

## Deliverables

- [ ] Root cause identified from the actual code/request, not assumed from Seer's mismatched summary.
- [ ] Code fix with a regression test.
- [ ] A note on how long review quality was degraded and whether any additional signal (test, log) should have caught this sooner.

PR #5111, opened 2026-07-11
fix(selfhost): keep embedding errors status-only

JSONbored:
### Motivation

- Prevent provider-controlled embedding response bodies from being embedded in thrown Error values and then promoted into operational logs/Sentry, which can leak private request/input material. 

### Description

- Stop reading and propagating non-OK `/embeddings` response bodies by throwing a status-only error in `src/selfhost/ai.ts` (now throws `ai_embed_http_<status>` instead of including `res.text()` content). 
- Update unit tests in `test/unit/selfhost-ai.test.ts` to assert that embedding errors surface only the status code and that provider-controlled error bodies are not read. 
- Preserve existing chat-provider behavior and only alter the embeddings error path so functionality and fail-safe semantics remain unchanged except for the diagnostic detail removed.

### Testing

- Ran the targeted unit tests with `npx vitest run test/unit/selfhost-ai.test.ts -t "embeddings response|provider-controlled embeddings" --pool=forks`, and the focused test suite passed. 
- Ran `git diff --check` to verify whitespace/conflict hygiene, and it succeeded. 
- Note: a full `npm run test:ci` / unsharded coverage run was not executed in this environment; only the targeted unit tests listed above were run automatically and passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52358b6574832c80e232d3f9ae2516)

### DIFF
--- a/src/selfhost/ai.ts
+++ b/src/selfhost/ai.ts
@@ -302,15 +302,9 @@ export function createOpenAiCompatibleAi(opts: {
           body: JSON.stringify({ model: opts.embedModel ?? "bge-m3", input: options.text }),
           signal: AbortSignal.timeout(120_000),
         });
-        // #4996: the error previously carried only the status code, with no detail from the response body --
-        // diagnosing a real production ai_embed_http_400 required SSHing into the box and reasoning about
-        // likely causes from first principles, since the actual rejection reason (e.g. Ollama's own "input
-        // length exceeds..." message) was thrown away. Bounded read, best-effort (a body-read failure must
-        // never mask the original status in the thrown error).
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing: reverses #4996, which added response-body detail specifically to fix a real production debugging pain point (SSHing in to diagnose ai_embed_http_400). The stated info-disclosure concern doesn't hold up — Sentry here is the self-host operator's own opt-in DSN (src/selfhost/sentry.ts), not a shared or third-party sink, and the operator already has full access to the repo content being embedded. Keeping the response-body detail for debuggability.

PR #5208, opened 2026-07-12
fix(selfhost): keep secret files host-private

JSONbored:
### Motivation

- Revert the change that created world-readable secret placeholders so populated secret files are not encouraged to remain host-world-readable and leak high-value secrets to other local users.
- Preserve the original intent: the container must be able to read secret files while avoiding making populated files globally readable on the host.

### Description

- `scripts/selfhost-init-secrets.sh`: create missing placeholders with owner-only mode `600` and heal existing empty placeholders to `600` instead of `644`, leaving any non-empty file untouched so operators' populated files (and their chosen modes/ownership) are preserved.
- `secrets/README.md`: replace `>`/`cp` population guidance with an explicit install workflow using `sudo install -o 1000 -g 1000 -m 0400` to create container-readable, non-world-readable files and warn against redirection/copying into placeholders which would preserve insecure host modes.
- Regenerated `worker-configuration.d.ts` via the repo's CF typegen tooling to satisfy the `cf-typegen` check (no functional runtime changes were made beyond the generated types).

### Testing

- A shell regression check that exercised placeholder creation, empty-placeholder healing, and non-empty-file preservation using a temporary tree succeeded.
- `npm run cf-typegen` was run and succeeded, and the generated `worker-configuration.d.ts` was updated accordingly.
- The full local gate `npm run test:ci` was executed but failed during `tsc --noEmit` due to pre-existing/unrelated type errors (`Cannot find module '@jsonbored/gittensory-engine'` and implicit-`any` test errors), so the gate did not complete green.
- `npm audit --audit-level=moderate` was attempted but the audit endpoint returned `403 Forbidden`, so that automated dependency-review step did not complete successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5354bf3530832c878e5a8ba313583a)


JSONbored:
Closing: this reverses the fix for a documented, real production incident on the self-host dedicated server (#secrets-uid-mismatch) — the removed comment states plainly: "644 is the minimum that works portably across arbitrary host/container uid pairs... confirmed against a real deploy — this is exactly what happened the first time this shipped: every secret read failed with selfhost_secret_file_unreadable."

The security rationale for tightening to 600 is real, but the PR provides no safety net for the exact failure mode it reintroduces: an operator following the previously-documented `cp`/`>` population pattern (now removed from the README, but still muscle memory / cached in tutorials, old docs, existing automation) will silently reproduce selfhost_secret_file_unreadable for a *populated* secret — worse than the original bug, which only affected an unused placeholder. No code-level detection or clearer diagnostic was added to catch that case. Keeping 644 with the documented tradeoff.

PR #5211, opened 2026-07-12
fix(chat): target publicSafeSummary in the private-lane-signal redaction test

JSONbored:
## Summary
- Fixes a test currently failing on `main` (CI red since the #5149/#5106 merge sequence).
- #5149's private-lane-signal redaction test seeded the private-lane phrasing into the action's `why` field. #5106 (merged earlier the same session) had already dropped `why`/`blockedBy` from the grounding bundle entirely as a stronger data-minimization boundary. Neither PR conflicted textually, but combined, the test now asserts on content that never reaches the prompt.
- Retargets the test fixture to `publicSafeSummary`, the one action field #5106 still forwards through `redactGroundingText` — restores the test's actual intent (private-lane phrasing gets redacted before the prompt is built) against current behavior.

## Validation
- `npx vitest run test/unit/ai-chat-qa.test.ts` — 31/31 passing (was 1 failing on main).
- `npm run typecheck` — clean.
- Test-only change; no `src/**` production code touched.


JSONbored:
Superseded — #5201 fixed the same root cause (ai-chat-qa.test.ts's why→publicSafeSummary field) and merged first. Closing this one.

Issue #4878, opened 2026-07-11
Verify `git mv` rename recognition in coverage/diff tooling


JSONbored:
**Problem:** Whether the coverage/diff tooling correctly treats a moved file as a rename (versus a delete-plus-add, which would wrongly zero out its prior coverage history and could trip the patch-coverage gate on an otherwise-unchanged file) is unverified today — including for the already in-flight queue decomposition work.

**Area:** ORB / CI tooling

**Proposal:** Do a small, throwaway test move of a real file and confirm how Codecov and the diff tooling represent it, before any real extraction work in this milestone begins.

**Deliverables:**
- A documented, verified answer: does a plain `git mv` read as a rename to the coverage tooling, or does it need to be structured differently (e.g. moved in a commit with no other changes) to be recognized correctly?

**Acceptance criteria:**
- [ ] A real test move is confirmed to either preserve coverage history correctly, or a documented workaround is established for every subsequent phase to follow.

**Boundaries:**
- This is a one-time verification, not itself a real code move — revert the throwaway test move once the answer is confirmed.

PR #5212, opened 2026-07-12
test(#4878): throwaway git-mv rename verification, DO NOT MERGE

JSONbored:
Throwaway verification PR for #4878 — do not merge, do not review as real work. Opened as a draft specifically so the gate cannot auto-merge it.

Pure `git mv` of `src/review/inline-finding-category-parse.ts` (zero content change) plus the two trivial import-path updates it requires (`src/services/ai-review.ts`, its own test file), to observe whether Codecov's patch-coverage report represents this as a rename (coverage history preserved) or as a delete+add (coverage zeroed, which would wrongly flag an otherwise-unchanged file).

Will close this without merging once the Codecov report is observed.


JSONbored:
Closing per this PR's own note — codecov/patch reported (SUCCESS), observation complete.

Issue #5178, opened 2026-07-12
Add `<NAME>_FILE` secrets-file indirection for GITHUB_TOKEN and coding-agent credentials in fleet mode


JSONbored:
### Context
ORB already has a proven pattern for Docker/Swarm/K8s-secret-mount indirection in `src/selfhost/load-file-secrets.ts`, which lets an operator supply a credential via a file path instead of a plaintext env var. `gittensory-miner` has no equivalent: DEPLOYMENT.md's fleet-mode example only shows `-e GITHUB_TOKEN` as a plain env var, which means the token (and any coding-agent credential) is visible in plaintext via `docker inspect` on any host running the miner. This issue ports that pattern into the miner package.

### Dependencies
None — independently shippable. It does not require `docker-compose.miner.yml` (a separate sibling issue in this batch) to exist first — the `_FILE` indirection is a config-resolution change usable with a plain `docker run` today. It does, however, add new env-var names, so if the generated env-var reference issue in this same batch (`npm run miner:env-reference`) lands first, this PR should run that generator as part of its own commit to keep the reference in sync; if this PR lands first, note in its own follow-up that the reference generator will need to pick these up.

### Requirements
1. In the module(s) under `packages/gittensory-miner/lib/` that currently resolve `GITHUB_TOKEN` and the coding-agent credential env vars (per `MINER_CODING_AGENT_PROVIDER`: an API-key env var, or an auth-file path for claude-cli/codex-cli), mirror `src/selfhost/load-file-secrets.ts`'s pattern.
2. For each supported credential, check for a companion `<NAME>_FILE` variable (e.g. `GITHUB_TOKEN_FILE`) in addition to the plain `<NAME>` variable.
3. If `<NAME>_FILE` is set and the plain `<NAME>` is unset, read the file's contents (trimmed of trailing whitespace/newlines) and use that as the effective credential value.
4. If both `<NAME>` and `<NAME>_FILE` are set, follow the same precedence rule as `load-file-secrets.ts` (state explicitly, in code comments and docs, which one wins) rather than silently picking one.
5. If `<NAME>_FILE` is set but the file is missing or unreadable, fail with a clear, actionable error identifying the file path — never silently fall through to an empty/undefined credential.
6. Apply this to `GITHUB_TOKEN` and to whichever credential(s) the currently-configured `MINER_CODING_AGENT_PROVIDER` requires.
[...]

PR #5265, opened 2026-07-12
feat(miner): add <NAME>_FILE secrets-file indirection for fleet-mode credentials (#5178)

dhgoal:
## What

Ports ORB's proven `<NAME>_FILE` secrets-file indirection to **gittensory-miner** (Closes #5178). In fleet mode, `-e GITHUB_TOKEN` (and any coding-agent credential) sits in plaintext env, visible to anyone who can `docker inspect` the host. This lets operators supply credentials via a mounted Docker/Swarm/K8s secret instead: `-e GITHUB_TOKEN_FILE=/run/secrets/github_token`.

Mirrors `src/selfhost/load-file-secrets.ts`, with one deliberate miner-side difference: an unreadable `_FILE` path is a **hard error**, never a silent empty credential.

## Changes

- **`lib/load-file-credentials.js`** (+ `.d.ts`, new) — `loadFileCredentials(env?, readFile?)` resolves any `<NAME>_FILE` into `<NAME>` in place. Injectable `env`/`readFile` for tests; defaults to `process.env` + `readFileSync`.
- **`bin/gittensory-miner.js`** — calls `loadFileCredentials()` once at startup, **before** any CLI reads a credential, so `discover`/`manage`/`attempt`/etc. transparently pick up the resolved value.
- **`DEPLOYMENT.md`** — fleet-mode example with the `_FILE` + secret-mount alternative and the precedence/error rules.
- **`test/unit/miner-load-file-credentials.test.ts`** (new) — 8 tests.

## Behavior (per the acceptance criteria)

- **Precedence:** an explicit plain `<NAME>` always **wins** over `<NAME>_FILE` (documented, never a silent pick).
- **Hard error** on a missing/unreadable `_FILE` path, naming the var and path — no silent fallthrough to an empty credential.
- **Never logs the secret value** — only mutates `env` in place (invariant-tested).
- **Compose-safe:** `COMPOSE_FILE` / `COMPOSE_ENV_FILE` (Compose's own reserved `_FILE` vars) are never dereferenced.
- **No regression:** an operator's existing plain-env-var setup is byte-identical.
- **No control-flow touched** — governor/attempt/claim untouched; this is credential-resolution plumbing only.

## Test coverage

8 tests covering every branch: plain-only (no regression), file-only, both-set (plain wins, no file read), neither, empty file → `""`, unreadable → throws with the path, Compose-reserved skipped, and a **no-secret-logging invariant** (spies `console.error`/`console.log`, asserts the value never appears).

`npx vitest run test/unit/miner-load-file-credentials.test.ts` → 8/8 green · `tsc` clean.

> Note: per the issue, `packages/gittensory-miner/**` currently sits outside vitest's `coverage.include`, so `codecov/patch` can't measure it yet — the tests above still hold the 100%-including-invariants bar by hand.

Closes #5178

### DIFF
--- a/packages/gittensory-miner/DEPLOYMENT.md
+++ b/packages/gittensory-miner/DEPLOYMENT.md
@@ -61,10 +61,24 @@ docker run --rm -it \
   doctor
 ```
 
+**Fleet mode — avoid plaintext credentials.** A plain `-e GITHUB_TOKEN` is visible via `docker inspect` on the host. Any credential can instead be supplied via a `<NAME>_FILE` companion pointing at a mounted secret (Docker/Swarm/K8s secret at `/run/secrets/<name>`); the miner reads and trims the file at startup into the plain `<NAME>`:
+
+```sh
+docker run --rm -it \
+  -e GITTENSORY_MINER_CONFIG_DIR=/data/miner \
+  -e GITHUB_TOKEN_FILE=/run/secrets/github_token \
+  --mount type=bind,src=/host/secrets/github_token,dst=/run/secrets/github_token,ro \
+  -v miner-data:/data/miner \
+  gittensory-miner:latest \
+  doctor
+```
+
+Precedence: an explicit plain `GITHUB_TOKEN` always **wins** over `GITHUB_TOKEN_FILE`. An unreadable `_FILE` path is a hard startup error (never a silent empty credential). The same `_FILE` convention applies to the coding-agent credential env vars your configured `MINER_CODING_AGENT_PROVIDER` requires.
+
 The image entrypoint is `gittensory-miner`; pass subcommands after the image name (`status`, `doctor`, `claim`, …).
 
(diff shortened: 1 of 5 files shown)


JSONbored (commented):
Closing: 

```
Review summary
Ports the ORB `<NAME>_FILE` secrets-file indirection to gittensory-miner: a new `loadFileCredentials()` iterates `env` keys, resolves any unset `<NAME>` from a readable `<NAME>_FILE`, excludes Compose's own reserved `_FILE` vars, and throws hard on an unreadable file rather than silently continuing. It's wired in once at the very top of `bin/gittensory-miner.js` before any CLI dispatch, so all subcommands transparently see the resolved value, and the 8 unit tests cover precedence, trimming, the Compose exclusion, and the never-logs invariant against real code paths (not fabricated ones). The logic correctly mirrors the documented self-host loader and closes https://github.com/JSONbored/gittensory/issues/5178 with a narrow, single-purpose diff.

Blockers

packages/gittensory-miner/lib/load-file-credentials.js:28 treats `env[target]` as the precedence check, so an explicitly set empty plain variable such as `GITHUB_TOKEN=""` loses to `GITHUB_TOKEN_FILE`, violating the documented contract; change it to a presence check such as `if (Object.prototype.hasOwnProperty.call(env, target)) continue;` or explain why empty explicit env vars should be treated as unset.
```

PR #5266, opened 2026-07-12
feat(miner-deployment): add docker-compose.miner.yml for AMS fleet mode

### LINKED ISSUE #5177: Add `docker-compose.miner.yml` for AMS fleet mode with named services and restart policy
### Context
packages/gittensory-miner already ships a Dockerfile, but DEPLOYMENT.md currently asks operators to hand-assemble their own `docker run` invocation to run fleet mode (multiple parallel miner workers). No docker-compose file exists for AMS today: the repo-root `docker-compose.yml` is ORB's own review-stack compose, and the separate miner-ui daemon/launcher issue in Wave 5 covers the webapp, not this CLI.

### Dependencies
None — independently shippable; only requires the existing Dockerfile in `packages/gittensory-miner`, which is already shipped.

### Requirements
1. Add `packages/gittensory-miner/docker-compose.miner.yml`.
2. Define a `miner` service built from the existing `packages/gittensory-miner/Dockerfile`.
3. Declare a named volume mapped to the package's documented state-dir path (e.g. `/data/miner`) for persistence across container restarts.
4. Set `restart: unless-stopped` on the service.
5. Document — in a comment block within the compose file, and in DEPLOYMENT.md — how to use `docker compose -f docker-compose.miner.yml up --scale miner=N` to run N parallel workers, including how to partition each worker's config/state dir so parallel workers don't collide on the same SQLite files.
6. Reference environment variables via an env file rather than hardcoding credentials or provider config directly into the compose file (pointing at the `.gittensory-miner.env.example` starter file from this same batch, if it has already merged, or at the existing env documentation otherwise).
7. Update DEPLOYMENT.md's fleet-mode section to point operators at this compose file instead of (or in addition to) the current hand-rolled `docker run` instructions.

### Deliverables / Acceptance Criteria
- [ ] New file `packages/gittensory-miner/docker-compose.miner.yml`
- [ ] `miner` service built from the existing Dockerfile
- [ ] Named volume for the state-dir path with `restart: unless-stopped`
- [ ] `--scale` usage documented, including per-worker config/state partitioning
- [ ] DEPLOYMENT.md updated to reference the new compose file
- [ ] No credentials or secrets hardcoded in the compose file

### Test Coverage Requirements
This PR must ship with full test coverage for every changed line and branch — the repo's Codecov patch gate requires 99%+ coverage and the house standard is to aim for 100%, including both sides of every conditional/nullish-coalescing branch introduced. Because the primary artifact is a docker-compose YAML file rather than application logic, add: (1) a validation test (e.g. invoking `docker compose -f docker-compose.miner.yml config` or an equivalent YAML-schema check) covering both a default single-instance invocation and a `--scale`-style multi-instance invocation, to confirm the file parses and the volume/service definitions are well-formed, (2) an invariant test for "the compose file never contains a literal credential value" (a grep-style check for common secret/token patterns run against the committed file), and (3) a regression test is not applicable since this is net-new tooling, not a bug fix — state that explicitly in the PR.

**Codecov visibility note:** `packages/gittensory-miner/**`, `apps/gittensory-miner-ui/**`, and `apps/gittensory-miner-extension/**` currently sit entirely outside vitest's `coverage.include` glob, so `codecov/patch` cannot measure changes there yet — closing that gap is exactly what #4864 and #4865 (in this same milestone) do. Any part of this change under `packages/gittensory-engine/src/**` or the repo's own `src/**` remains fully Codecov-instrumented as usual. Either way, this does not lower the bar: treat the 100%-including-invariants-and-regression target above as the enforced house standard regardless of what Codecov can currently see, and it becomes gate-enforced for real once #4864/#4865 ship.

### Expected Outcome
An operator running AMS in fleet mode can bring up N parallel miner workers with one `docker compose` command and a documented `--scale` pattern, instead of hand-assembling repeated `docker run` invocations.

### Links & Resources
- packages/gittensory-miner/Dockerfile
- packages/gittensory-miner/DEPLOYMENT.md
- See the `.gittensory-miner.env.example` issue in this same batch
- Theme: Self-host packaging & docs

davion-knight:
Adds `packages/gittensory-miner/docker-compose.miner.yml` so operators run the miner as a long-lived worker instead of a hand-assembled `docker run` (#5177). Built on the existing package Dockerfile.

## Contents
- **`docker-compose.miner.yml`** — a `miner` service built from the Dockerfile (monorepo-root context), `command: ["run"]` (continuous worker), `restart: unless-stopped`, SQLite state on a named `miner-data` volume, and credentials via `env_file` (not inlined).
- **`.gittensory-miner.env.example`** — env template with **empty** values (real values go in an uncommitted `.gittensory-miner.env`; nothing credential-like is committed).
- **`DEPLOYMENT.md`** — a Docker Compose section: `up` usage + scaling. It documents `--scale miner=N` **and** the honest caveat that a shared `miner-data` volume corrupts the miner’s SQLite ledgers (not safe for concurrent access), with the per-worker isolation patterns: separate compose projects (`-p miner-1 …`) or distinct `GITTENSORY_MINER_CONFIG_DIR` mounts — and points at the `k8s/` StatefulSet for built-in isolated scaling.

## Validation
- **`test/unit/miner-docker-compose.test.ts`** (4 tests): the service is built from the package Dockerfile with the named volume + `unless-stopped` restart, credentials come from an env file (no secret-named key assigned a value in the compose), and the env example ships only empty placeholders. The existing `miner-deployment-doc.test.ts` still passes (required DEPLOYMENT.md sections preserved).
- YAML parses; `tsc` typecheck clean; prettier-clean; no hardcoded credentials. Static packaging only — no `src/**` logic, no runtime control-flow touched.

Closes #5177

### DIFF
--- a/packages/gittensory-miner/.gittensory-miner.env.example
+++ b/packages/gittensory-miner/.gittensory-miner.env.example
@@ -1,0 +1,11 @@
+# Env file for docker-compose.miner.yml (#5177). Copy to `.gittensory-miner.env`, fill in real values, and do
+# NOT commit the filled-in copy. Values are intentionally EMPTY here (a real-looking value in a committed file
+# trips secret scanners). Compose reads `KEY=value` lines; leave a key blank or delete it if unused.
+#
+# Required: a GitHub token the miner uses to read issues and open PRs.
+GITHUB_TOKEN=
+# Optional: only for the coding-agent providers you enable (claude-cli / codex-cli / agent-sdk).
+ANTHROPIC_API_KEY=
+OPENAI_API_KEY=
+# Optional: override the on-container state dir (defaults to /data/miner, mapped to the miner-data volume).
+GITTENSORY_MINER_CONFIG_DIR=/data/miner
--- a/packages/gittensory-miner/DEPLOYMENT.md
+++ b/packages/gittensory-miner/DEPLOYMENT.md
@@ -2,13 +2,13 @@
 
 Two form factors for running `@jsonbored/gittensory-miner`: **laptop mode** (single machine, zero Docker) and **fleet mode** (containerized workers with a shared data volume). Both are 100% client-side for core operation — the miner never uploads source and never requires a hosted Gittensory callback to boot. Credentials (GitHub tokens, etc.) stay on the operator's machine or in their own secret store; nothing is baked into images.
 
-| | Laptop mode | Fleet mode |
-|---|---|---|
-| **Best for** | One contributor machine, local experimentation | Many parallel miner attempts on a host or small cluster |
-| **Dependencies** | Node.js `>=22.13.0` only | Docker (or compatible runtime) + Node image or custom image |
-| **State** | SQLite files under `~/.config/gittensory-miner/` (override with `GITTENSORY_MINER_CONFIG_DIR`) | Same SQLite layout on a mounted `/data` (or `GITTENSORY_MINER_CONFIG_DIR`) volume |
-| **Setup** | `npm install -g @jsonbored/gittensory-miner` or workspace build | `docker build` + `docker run` with env + volume (see below) |
-| **Footprint** | One Node process, local disk for ledgers/queues | One container per worker; scale horizontally by adding containers |
+|                  | Laptop mode                                                                                    | Fleet mode                                                                        |
+| ---------------- | ---------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- |
+| **Best for**     | One contributor machine, local experimentation                                                 | Many parallel miner attempts on a host or small cluster                           |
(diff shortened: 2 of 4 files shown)


JSONbored (commented):
Closing:

```
packages/gittensory-miner/docker-compose.miner.yml:36 hardcodes GITTENSORY_MINER_CONFIG_DIR in the compose environment, so an operator following packages/gittensory-miner/.gittensory-miner.env.example:11 to override that variable through the env file will be silently ignored by Compose's precedence rules; remove the hardcoded environment entry or change it to an interpolated default such as `GITTENSORY_MINER_CONFIG_DIR: ${GITTENSORY_MINER_CONFIG_DIR:-/data/miner}`.
```

PR #5347, opened 2026-07-12
feat(miner): add schema-version migration runner across local stores

### LINKED ISSUE #4832: Add schema versioning to the local SQLite stores
**Problem:** Every one of the miner's local SQLite stores only ever does `CREATE TABLE IF NOT EXISTS` — there's no `user_version`/migration mechanism at all, unlike the main product's `migrations/` discipline.

**Area:** AMS / Miner

**Proposal:** Add a lightweight schema-version column and migration-runner convention shared across all the miner's local stores.

**Deliverables:**
- A schema-versioning convention applied consistently across every store.

**Acceptance criteria:**
- [ ] Opening an older-schema database file triggers a migration rather than silent incompatibility or data loss.

**Boundaries:**
- Keep it lightweight — this doesn't need the full weight of the main product's migration tooling, just a real version check.

real-venus:
## Summary

The miner's local SQLite stores only ever ran `CREATE TABLE IF NOT EXISTS`, with **no `user_version`/migration mechanism at all** — so an older on-disk database file was silently reused with a stale shape, unlike the main product's `migrations/` discipline. This adds a lightweight, shared schema-versioning convention across every store.

**New:** `packages/gittensory-miner/lib/schema-version.js` (+ `.d.ts`) exporting `applySchemaMigrations(db, migrations)`:
- Each store's bootstrap schema is treated as **version 1** (`BASELINE_SCHEMA_VERSION`); a store's `migrations` array holds **only post-baseline changes** — `migrations[i]` upgrades the schema from version i+1 to i+2.
- The runner reads `PRAGMA user_version`, runs exactly the **pending** migrations in order, then stamps the new version — so opening an older file runs its outstanding migrations instead of continuing on an incompatible shape. Re-opening an up-to-date file runs none.
- A pre-versioning file (`user_version` 0) already carries the baseline tables (the idempotent `CREATE TABLE IF NOT EXISTS` ran), so it advances from the baseline.
- Pure control flow over an injected `DatabaseSync` handle — deterministic, no IO of its own beyond the PRAGMA read/write.

**Wired into all seven stores:** claim-ledger, event-ledger, governor-ledger, plan-store, run-state, laptop-init, portfolio-queue.

**Portfolio-queue's existing ad-hoc `leased_at` ALTER is expressed as its first real migration (1 → 2)** — kept defensive (checks `table_info`) so a version-0 file that already ran the pre-convention ad-hoc ALTER is not re-altered into a duplicate-column error. This both removes the ad-hoc branch and demonstrates the convention on a real migration.

Kept lightweight per the issue's boundary — no heavyweight migration framework.

## Scope

- [x] Narrow, one coherent change — one shared runner + uniform per-store wiring + tests
- [x] In scope (`packages/`), no `blockedPaths`, no secrets/private terms
- [x] Hand-authored `.d.ts` companion added, matching the package's existing convention

## Validation

- [x] `npm run typecheck`
- [x] `npm run test:coverage` (full unsharded suite)
- [x] New unit test `test/unit/miner-schema-version.test.ts` (in-memory `DatabaseSync`): baseline stamp with no migrations, full run in order on a pre-versioning file, idempotent re-apply, partial catch-up from an intermediate version, and coercion of an absent/non-integer/negative `user_version`
- [x] Every existing store test still passes (verified against a clean-`main` baseline: identical results — the Windows-only chmod/path test failures are pre-existing and unrelated); portfolio-queue's `leased_at`/`markFailed` tests confirm the migrated behavior is preserved

## Safety

- [x] Behaviour-preserving: stores gain a `user_version` stamp and a migration hook; the only converted logic (portfolio-queue `leased_at`) keeps its exact defensive semantics
- [x] No secrets, tokens, wallets, trust scores, or reward values in code, tests, or this description

Closes #4832

### DIFF
--- a/packages/gittensory-miner/lib/claim-ledger.js
+++ b/packages/gittensory-miner/lib/claim-ledger.js
@@ -1,4 +1,5 @@
 import { normalizeLocalStoreDbPath, openLocalStoreDb, resolveLocalStoreDbPath } from "./local-store.js";
+import { applySchemaMigrations } from "./schema-version.js";
 
 // The miner's local soft-claim ledger (#2314): a 100% client-side record of "I'm working on issue #N in repo X",
 // so Phase 2's soft-claim adjudication (sibling issues) has somewhere to persist claims. Schema + CRUD only — no
@@ -71,6 +72,8 @@ export function openClaimLedger(dbPath = resolveClaimLedgerDbPath()) {
       UNIQUE (repo_full_name, issue_number)
     )
   `);
+  // Schema-version convention (#4832): stamp the baseline and run any post-baseline migrations (none yet).
+  applySchemaMigrations(db, []);
 
   // Idempotent claim in ONE atomic statement: insert a new active claim, or — only if the existing row is NOT
   // already active — re-activate it (a released/expired claim can be re-claimed). The `WHERE status <> 'active'`
--- a/packages/gittensory-miner/lib/event-ledger.js
+++ b/packages/gittensory-miner/lib/event-ledger.js
@@ -1,5 +1,6 @@
 import { isDeepStrictEqual } from "node:util";
 import { normalizeLocalStoreDbPath, openLocalStoreDb, resolveLocalStoreDbPath } from "./local-store.js";
+import { applySchemaMigrations } from "./schema-version.js";
 
 // The miner's local, append-only event ledger (#2290): an immutable audit trail of every significant miner-loop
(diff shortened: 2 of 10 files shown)


JSONbored (commented):
Closing: 

```
packages/gittensory-miner/lib/schema-version.js:42 applies multiple migrations before stamping once at line 51, so if migration N succeeds and migration N+1 throws, the next open reruns migration N against an already-mutated schema and can fail permanently; wrap the migration loop and final `PRAGMA user_version` in one transaction, or stamp after each successful migration with a transaction around each migration plus its stamp.
```

PR #5352, opened 2026-07-12
fix(github): fallback to public Orb slug when GITHUB_APP_SLUG is unset

JSONbored:
### Motivation
- The stale-approval dismissal path computed the bot login from `env.GITHUB_APP_SLUG` unguarded, which becomes `undefined[bot]` when the env var is removed and prevents dismissal of real Orb approvals, weakening branch-protection integrity. 
- The change restores the intended security behavior by falling back to the publicly-installable Orb slug while preserving self-host behavior when an operator provides their own `GITHUB_APP_SLUG`.

### Description
- Add a `PUBLIC_GITHUB_APP_SLUG = "gittensory-orb"` constant and compute the effective bot login with `env.GITHUB_APP_SLUG?.trim() || PUBLIC_GITHUB_APP_SLUG` inside `dismissLatestBotApproval` so dismissal still targets the real Orb bot when the env var is absent. 
- Update the `dismissLatestBotApproval` implementation to use the trimmed/ fallback slug before scanning reviews. 
- Extend `test/unit/github-pr-actions.test.ts` by allowing env overrides in `envWithKey(...)` and adding regression tests that cover the unset and blank `GITHUB_APP_SLUG` branches which assert that `gittensory-orb[bot]` approvals are still dismissed. 
- Preserve existing behavior when `GITHUB_APP_SLUG` is configured so self-host deployments continue to prefer their configured slug.

### Testing
- Ran the targeted unit test file with `npx vitest run test/unit/github-pr-actions.test.ts`, and the file's tests passed (42 tests passed). 
- Attempted a full coverage run with `npm run test:coverage`, which exercised the test harness but failed global coverage thresholds and encountered long-running unrelated test suites in this environment, so the full unsharded coverage job did not report a green result. 
- Ran `npm run typecheck`, which failed due to pre-existing/unrelated typing issues in other packages and files and not due to the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a53996c9a20832ca6e0d5a4cd7c8cfd)

### DIFF
--- a/src/github/pr-actions.ts
+++ b/src/github/pr-actions.ts
@@ -10,6 +10,7 @@ const ISSUE_EVENTS_RECENT_PAGE_LIMIT = 10;
 // earliest reviews on a PR with a long review history and could dismiss (or miss) the wrong one.
 const REVIEW_PAGE_SIZE = 100;
 const REVIEW_PAGE_LIMIT = 10;
+const PUBLIC_GITHUB_APP_SLUG = "gittensory-orb";
 
 // The GitHub write primitives the maintainer auto-maintain layer (#778) uses to act on a PR's STATE — never
 // its source. Thin wrappers over the installation-scoped REST API, mirroring labels.ts / comments.ts. Each
@@ -128,7 +129,8 @@ export async function dismissLatestBotApproval(env: Env, installationId: number,
     const { owner, repo } = splitRepo(repoFullName);
     return await withInstallationTokenRetry(env, installationId, async (token) => {
       const octokit = makeInstallationOctokit(env, token, "live", githubRateLimitAdmissionKeyForInstallation(installationId));
-      const botLogin = `${env.GITHUB_APP_SLUG}[bot]`;
+      const botSlug = env.GITHUB_APP_SLUG?.trim() || PUBLIC_GITHUB_APP_SLUG;
+      const botLogin = `${botSlug}[bot]`;
       // Reviews are returned oldest-first; the LAST matching entry across ALL pages is the bot's most recent
       // APPROVE. Stopping at page 1 would find (or miss) the wrong review on a PR with >100 total reviews.
       let latestApprovalId: number | undefined;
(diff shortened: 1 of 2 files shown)


JSONbored:

`${env.GITHUB_APP_SLUG}[bot]` unset does become the literal string `"undefined[bot]"` — but that's already a safe no-op (no real GitHub bot is named `undefined`), not an active bypass. It's the same fail-closed *outcome* as the explicit pattern in `isOwnGitHubAppCheckRun`/`isOwnReviewThreadAuthor` (`src/github/backfill.ts`), just implemented implicitly via string coercion instead of an explicit early-return.

More importantly, this PR's fix inverts a deliberate design decision from #5269 (d56f4eccd), which was already merged into `main` *before* this branch was created. That commit established two distinct patterns on purpose: own-bot-identity checks (this one's category, per `isOwnReviewThreadAuthor`'s own comment listing `pr-actions.ts` as a sibling) degrade to "no match" when `GITHUB_APP_SLUG` is unset — never a hardcoded fallback — while `PUBLIC_GITHUB_APP_SLUG = "gittensory-orb"` is reserved for a completely different, public-facing "which app to install" descriptor (`src/services/subnet-interface.ts`). This PR imports that public constant into an own-identity comparison, which #5269 explicitly reasoned against ("decoupled from whatever credentials this Worker happens to hold for its own operational purposes").

It also only patches 1 of 5 byte-identical `${env.GITHUB_APP_SLUG}[bot]` sites (`comments.ts`, `project-tracker-adapter.ts`, `processors.ts`, `review-evasion.ts` are untouched), so even taken at face value it wouldn't close the claimed gap.

If we want the explicit fail-closed idiom applied consistently, the right shape is extending `backfill.ts`'s existing pattern (or factoring a shared `resolveOwnBotLogin(env)` helper) across all 5 call sites — not a hardcoded slug fallback. No action needed right now; flagging for whoever revisits this.

PR #5491, opened 2026-07-12
fix(rees): require auth for /metrics and guard server startup

JSONbored:
### Motivation
- The REES `/metrics` endpoint was exposed without authentication and could be scraped remotely, revealing aggregate telemetry and analyzer health; the service needs to enforce the same shared-bearer auth used by its private API.
- Tests and imports should be able to import the Hono `app` without starting a network listener so route-level tests can run in-process.

### Description
- The `/metrics` route now checks `normalizeSharedSecret(process.env.REES_SHARED_SECRET)` and calls `verifyBearer(...)` before returning Prometheus text via `renderMetrics()` so unauthenticated scrapes are rejected.
- Server startup side effects (calling `serve()` and installing process signal handlers) are wrapped behind a direct-execution guard using `if (import.meta.url === pathToFileURL(process.argv[1] ?? "").href)` so test imports do not open a listener.
- Added a regression test `review-enrichment/test/server-metrics-auth.test.ts` that asserts unauthenticated scrapes receive `401` and authenticated scrapes receive metrics, and adjusted imports (`pathToFileURL`) needed for the startup guard.

### Testing
- Ran `git diff --check` which produced no trailing-whitespace or conflict-marker errors and succeeded.
- Ran the REES test suite via `npm run rees:test` which completed with all tests passing.
- Ran the new route tests directly with `node --test --experimental-strip-types test/server-metrics-auth.test.ts` and they passed locally.
- Attempted `npm audit --audit-level=moderate` but the registry audit endpoint returned `403` in this environment, so the audit could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a53eb1cc258832cae39aa507a96859b)

### DIFF
--- a/review-enrichment/src/server.ts
+++ b/review-enrichment/src/server.ts
@@ -8,6 +8,7 @@
 // Transport + contract here; the analysis lives in brief.ts (orchestrator) + analyzers/*, with each analyzer
 // filling one findings key for renderer/prompt consumption.
 import { serve } from "@hono/node-server";
+import { pathToFileURL } from "node:url";
 import { Hono } from "hono";
 import { normalizeSharedSecret, verifyBearer } from "./auth.js";
 import { buildBrief } from "./brief.js";
@@ -46,7 +47,13 @@ app.get("/health", (c) =>
   c.json({ status: "ok", service: "review-enrichment" }),
 );
 app.get("/ready", (c) => c.json({ ready: true }));
-app.get("/metrics", (c) => c.text(renderMetrics()));
+app.get("/metrics", (c) => {
+  const secret = normalizeSharedSecret(process.env.REES_SHARED_SECRET);
+  if (!secret) return c.json({ error: "service_not_configured" }, 503);
+  if (!verifyBearer(c.req.header("authorization"), secret))
+    return c.json({ error: "unauthorized" }, 401);
+  return c.text(renderMetrics());
+});
 
 function recordEnrichOutcome(status: string, startedAtMs: number): void {
   incr("rees_enrich_requests_total", { status });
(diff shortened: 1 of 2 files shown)


JSONbored:

The premise — that REES's `GET /metrics` is internet-exposed and "could be scraped remotely" — doesn't hold under this repo's actual, deliberate architecture:

- `docker-compose.yml`'s `rees` service uses `expose: ["8080"]` with an explicit comment: *"Never published to the host — reached only over the internal docker network at http://rees:8080, same in-network-only posture as postgres/redis/qdrant/browserless."*
- The Railway deployment path uses private networking — `review-enrichment/README.md`'s own Deploy section states no public domain is required.
- `prometheus.yml`'s `rees` scrape job has zero bearer-token configuration, consistent with all 11 other scrape jobs in that file.
- The main app's own `GET /metrics` (`src/server.ts:829`) is likewise unauthenticated, by the same deliberate convention.

More concretely, this would break already-shipped, same-day work: #5412 (`aec16727a`, already an ancestor of this branch) wired a live Prometheus scrape + a dedicated Grafana dashboard against this exact unauthenticated endpoint, and its commit message documents it was verified end-to-end against a real deployment ("exporter -> scrape -> query"). Merging this PR's 401 gate would silently break that dashboard's data on any correctly configured `--profile rees --profile observability` self-host install, since Prometheus has no bearer-token wiring to satisfy it.

The code itself is clean and well-tested (the `import.meta.url` main-module guard is a reasonable pattern) — the underlying premise just doesn't match this repo's actual deployment model. No action needed; flagging for visibility.

PR #5496, opened 2026-07-12
fix(miner): halt loop reentry for unresolved PRs

JSONbored:
### Motivation
- The autonomous miner loop could re-enter and submit additional PRs when a prior submitted PR was still open or its PR number was unparseable, risking unwanted GitHub write spam.
- The change prevents the loop from advancing until a submitted PR reaches a closed (terminal) disposition or a parseable PR number is available.

### Description
- Add a guard in `packages/gittensory-miner/lib/loop-cli.js` that, when an attempt submitted a PR but the disposition is not `closed` (or the PR number could not be parsed), records a declined re-entry cycle and halts the loop with an explicit reason.
- Surface explicit decline reasons: `submitted_pr_disposition_unresolved` when an observed PR remains open and `submitted_pr_number_unresolved` when no parseable PR number is exposed.
- Add regression unit tests in `test/unit/miner-loop-cli.test.ts` that assert the loop halts re-entry for an open PR and for a submitted attempt that exposes no parseable PR number, and that the queue/ledger state is left consistent.

### Testing
- Ran `npx vitest run test/unit/miner-loop-cli.test.ts`, and the updated unit test file passed (`17 tests` passed).
- Ran `npm run build:miner`, which completed successfully (`node --check` passed for the miner package files).
- Began the full gate with `git diff --check && npm run test:ci`; the run progressed through many checks but stopped at `cf-typegen:check` due to a pre-existing `worker-configuration.d.ts is stale; run npm run cf-typegen` issue unrelated to the miner change.
- `npm audit --audit-level=moderate` could not complete against the registry in this environment (registry returned `403 Forbidden`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a53a9c65b20832c90cb6779132a5256)

### DIFF
--- a/packages/gittensory-miner/lib/loop-cli.js
+++ b/packages/gittensory-miner/lib/loop-cli.js
@@ -410,6 +410,23 @@ export async function runLoop(args, options = {}) {
       );
       sinceSeq = loopSummary.lastSeq;
 
+      if (submitted && prDisposition?.state !== "closed") {
+        const reason = prNumber === null ? "submitted_pr_number_unresolved" : "submitted_pr_disposition_unresolved";
+        cycles.push({
+          cycle: cycleIndex,
+          outcome: "attempted",
+          repoFullName: claimed.repoFullName,
+          identifier: claimed.identifier,
+          attemptOutcome,
+          reentryOutcome,
+          prNumber,
+          reentered: false,
+          reasons: [reason],
+        });
+        haltReason = `reentry_declined:${reason}`;
+        break;
+      }
+
       const reentry = attemptLoopReentryFn(
         { killSwitchScope: killSwitch.scope, repoFullName: claimed.repoFullName, outcome: reentryOutcome },
         { eventLedger, portfolioQueue, runState, nowMs: nowMsFn(), sessionStartMs, loopSummary },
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing: this inverts a deliberate, same-day design decision rather than fixing a real gap.

`loop-cli.js`'s reentry wiring (together with `pr-disposition-poller.js`'s `classifyPrDisposition`) was authored by #5303 (commit dc6fa7a73, merged same day). Its own doc comments are explicit that a still-open/unresolved submitted PR classifies as `'other'` and must NOT be treated as a failure or blocking condition — "not an error", "not fail", "nothing to re-enter on yet" is intentional: the loop is meant to keep making forward progress on other work while a submitted PR is still pending review, not halt entirely.

This PR's guard turns "still open" into an unconditional full-loop halt, directly reversing that same-day intent. Separately, the double-submission concern the PR's motivation describes doesn't hold up: the specific (repo, identifier) pair is already marked done via `portfolioQueue.markDone(...)` before any PR-disposition handling runs, so it can't be re-attempted regardless of the PR's later outcome — there's no double-submission path to guard against.

PR #5497, opened 2026-07-12
fix(miner): reject symlinked goal specs

JSONbored:
### Motivation

- The local MinerGoalSpec resolver previously followed repository-controlled symlinks and used `readFileSync` on the joined worktree path, allowing a malicious repo to cause a DoS or unbounded read by pointing `.gittensory-miner.yml` at a blocking or large local target. 
- The intent of this change is to harden the resolver so it never follows symlinks, validates the discovered entry is a regular file, and bounds the on-disk read before loading content into memory.

### Description

- Added a safe reader `readRegularUtf8File` in `packages/gittensory-miner/lib/miner-goal-spec.js` that uses `lstat`, opens with `O_NOFOLLOW`, `fstat`-validates regular file semantics, enforces a `MAX_MINER_GOAL_SPEC_BYTES` size bound, and closes the descriptor before returning content. 
- Replaced the previous direct `readFileSync(join(repoPath, relativePath), "utf8")` call with the new bounded/no-follow read path in `resolveMinerGoalSpec`. 
- Updated the resolver types in `packages/gittensory-miner/lib/miner-goal-spec.d.ts` to expose the injected filesystem hooks (`lstatSync`, `openSync`, `fstatSync`, `readFileSync` accepting a fd, and `closeSync`). 
- Added regression tests in `test/unit/miner-goal-spec-resolver.test.ts` to assert that symlinked `.gittensory-miner.yml` files are rejected and that oversized files are ignored before being read into memory. 
- Regenerated Cloudflare worker runtime types (`worker-configuration.d.ts`) so `cf-typegen:check` remains up-to-date with the current Wrangler/workerd output.

### Testing

- Ran the new unit tests with `npx vitest run test/unit/miner-goal-spec-resolver.test.ts`, and the test file passed. 
- Built the miner with `npm run build:miner` and ran `npm run cf-typegen:check`, both of which completed successfully. 
- Attempted the full local gate `npm run test:ci`, which progressed to `test:coverage` but did not complete in this environment due to unrelated long-running queue suites timing out on retries. 
- `npm audit --audit-level=moderate` was attempted but the registry audit endpoint returned `403 Forbidden` in this environment, so that automated audit could not complete here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a53ace13980832c95ab362f3671bc74)


JSONbored:
The symlink-rejection fix itself (\`readRegularUtf8File\`: lstat pre-check, O_NOFOLLOW open, fstat size+regular-file validation, proper fd cleanup) is correct and its regression tests were already passing. The only thing failing CI was an unrelated, incidentally-bundled \`worker-configuration.d.ts\` regeneration — it was generated against a newer wrangler/workerd than CI's pinned version produces, which is what tripped \`validate-code\`'s cf-typegen drift check. Pushed a follow-up commit reverting that file to \`origin/main\`'s content; the goal-spec resolver changes are untouched.

PR #5499, opened 2026-07-12
fix(miner): clamp repo AMS policy defaults

JSONbored:
### Motivation

- Prevent target repositories from loosening an operator's execution-risk policy by supplying permissive `.gittensory-ams.yml` content; repo-controlled values must not be allowed to enable live submission or raise resource caps above operator-safe defaults.

### Description

- Clamp repo-provided AMS policy fields before they become effective by adding `clampRepoAmsPolicySpec` and helpers (`SLOP_THRESHOLD_RANK`, `clampNumberToDefault`) and applying it to repo fallbacks in `resolveAmsPolicy` in `packages/gittensory-miner/lib/ams-policy.js`.
- Keep the operator's local policy authoritative and unchanged, and document that repo files can only propose a more restrictive fallback.
- Add regression/unit tests in `test/unit/miner-ams-policy.test.ts` that assert permissive repo policies are clamped, restrictive repo policies are allowed, and local files still fully replace repo content.

### Testing

- Ran the unit test file `npx vitest run test/unit/miner-ams-policy.test.ts --pool=forks --reporter=verbose` and all tests in that file passed (10/10).
- Verified `npm run build:miner` and `git diff --check` completed successfully in this environment.
- Attempted full local gate via `npm run test:ci` but the run could not complete due to an environment/registry issue (`worker-configuration.d.ts` drift requiring `npm run cf-typegen` and `npm audit` failing with a registry `403 Forbidden`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a53ace1521c832ca47d85c1053c1b72)

### DIFF
--- a/packages/gittensory-miner/lib/ams-policy.js
+++ b/packages/gittensory-miner/lib/ams-policy.js
@@ -6,8 +6,9 @@ import { resolveLocalStoreDbPath } from "./local-store.js";
 // (ams-policy-spec.ts, engine package) is the type/parser surface; this module is the actual fetch+resolve
 // caller, mirroring `.gittensory.yml`'s own established self-host precedent (src/selfhost/private-config.ts's
 // `makeLocalManifestReader`): the operator's own local file, when present, FULLY REPLACES whatever the
-// target repo's own file says -- never a field-by-field merge. The repo's file is only ever consulted as a
-// fallback default for an operator who hasn't set their own local policy.
+// target repo's own file says -- never a field-by-field merge. The repo's file can only ever propose a
+// more restrictive fallback for an operator who hasn't set their own local policy; repo-controlled content
+// is clamped to the engine's safe deny-by-default policy before it becomes effective.
 //
 // This is deliberately NOT the same resolution shape as self-review-context.js/rejection-signal.js, which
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing: this reverts deliberate, same-day work rather than fixing a live gap.

The vulnerability this targets (a target-repo `.gittensory-ams.yml` loosening operator-side AMS execution policy) was already fixed — more thoroughly — in #5351 (commit 5f2851923, merged ~82 minutes before this PR's commit). That fix removed `fetchRepoAmsPolicyContent`/`AMS_POLICY_SPEC_FILENAMES`/the outbound fetch of target-repo content entirely, so repo-provided policy is never read at all (`resolveAmsPolicy` on main is fully operator-local with no repo-content code path). It also added a standing regression test asserting `fetchImpl` is never called.

This PR's diff is built against a pre-#5351 snapshot and reintroduces the fetch/clamp machinery that fix deliberately deleted — merging it would be a functional regression to a strictly weaker mitigation (clamping repo-provided values) for a threat model #5351 already closed off entirely (never trusting repo-provided values at all). Its CONFLICTING mergeable state against main reflects exactly this: main no longer has the symbols this diff's hunks reference.

Issue #4842, opened 2026-07-11
Add conditional-GET caching for policy-doc fetches


JSONbored:
**Problem:** Policy documents (contribution/AI-usage guidance) are refetched from GitHub on every single `discover` invocation even though they rarely change, burning rate-limit budget on essentially static content.

**Area:** AMS / Performance

**Proposal:** Add ETag/If-None-Match conditional-GET support to the policy-doc fetch path.

**Deliverables:**
- Conditional-GET wired into the policy-doc fetch, with a local cache of the last-known ETag/content.

**Acceptance criteria:**
- [ ] A repeated `discover` run against an unchanged repo makes a conditional request and consumes no additional rate-limit budget when the doc hasn't changed.

Scope note: this issue is limited to the HTTP-level fetch mechanism (ETag / If-None-Match) for the policy document. It must not change how the fetched content is parsed or how a policy verdict is derived from it — that logic stays out of scope. On any error, timeout, or ambiguous cache state, the fetch must fail safe by refetching the full document rather than serving a possibly-stale cached copy: silently serving stale policy content that could let the autonomous loop treat a now-opted-out repo as eligible is out of scope for this issue and stays maintainer-owned.

PR #5516, opened 2026-07-13
feat(miner): persist resolved policy verdicts across discover runs

### LINKED ISSUE #4843: Add a persisted cross-invocation cache of resolved policy verdicts
**Problem:** Even with conditional-GET caching (above), the resolved AI-usage-policy verdict for a repo is recomputed fresh on every invocation rather than persisted across runs.

**Area:** AMS / Performance

**Proposal:** Persist the resolved policy verdict per repo locally, invalidated when the underlying doc's ETag changes.

**Deliverables:**
- A small local cache of resolved policy verdicts, keyed by repo and doc ETag.

**Acceptance criteria:**
- [ ] Repeated runs against the same unchanged repo skip re-resolution entirely.

JSONbored:
## Summary
- Adds `policy-verdict-cache.js`, a local cache of resolved `AiPolicyVerdict` objects keyed by repo + the ETag of whichever doc (AI-USAGE.md or CONTRIBUTING.md) decided the verdict, following the same `local-store.js`/`schema-version.js` conventions as every other store in this package.
- `fetchRepoDoc` (opportunity-fanout.js) now returns `{content, etag}` instead of bare content, so `resolveRepoAiPolicy` can see the ETag of whichever doc it fetched.
- `resolveRepoAiPolicy` checks the verdict cache before calling `resolveAiPolicyVerdict`: a hit requires BOTH the decisive doc name and its ETag to match the cached entry (guards against a repo whose decisive doc changes between runs, e.g. AI-USAGE.md going from absent to present); a miss always resolves fresh and re-caches.
- Same fail-open discipline as #4842's policy-doc cache throughout: a cache read/write failure degrades to "resolve fresh," never affects discovery, and `discover-cli.js` opens the store in its own try/catch so a corrupt/unwritable cache DB can't abort a run.

## Test plan
- [x] `npx vitest run test/unit test/contract` — 780 files / 15197 tests passing, 0 regressions (one pre-existing unrelated flake in `github-graphql-cache.test.ts` reproduced only under full-suite parallel load, confirmed passing in isolation and untouched by this diff)
- [x] `npm run build --workspace @jsonbored/gittensory-miner` (`node --check` over every lib file, including the new one)
- [x] `npx tsc --noEmit` clean
- [x] `npm run docs:drift-check && npm run manifest:drift-check && npm run engine-parity:drift-check && npm run miner:env-reference:check`
- [x] Dedicated tests for: cold-cache resolve+cache, cache-hit reuse (proven via a deliberately wrong planted verdict), recompute on ETag change, recompute on decisive-doc mismatch even with a colliding ETag, CONTRIBUTING.md-decisive path, cache read/write failure fail-open, no-etag-no-cache, inert-without-a-cache, real on-disk persistence, default-store open/close, and the corrupt-cache-degrades-not-fails regression

Closes #4843


JSONbored:
Fixed the cross-host cache-key issue: the policy-verdict cache is now keyed by \`\${apiBaseUrl}::\${repoFullName}\` instead of a bare \`owner/repo\`, matching \`policy-doc-cache.js\`'s own precedent of keying on the full request URL. Renamed the store's column/field from \`repo_full_name\` to \`repo_scope\` to reflect that it's now a caller-owned composite key. Added a dedicated regression test proving two different tenant forge hosts with the same \`owner/repo\` (and even a colliding ETag string) get independent cache entries. Full local gate re-run clean; force-pushed the rebased branch.

Issue #5192, opened 2026-07-12
Persist a savedAt timestamp with rankedCandidates and render a 'last synced' label on the extension's opportunity badge


JSONbored:
### Context
apps/gittensory-miner-extension/options.js currently persists rankedCandidates to chrome.storage.local whenever a contributor pastes a new discover-run JSON blob, but it doesn't record when that save happened. opportunity-badge.js renders the badge from rankedCandidates but has no way to tell the contributor how stale that data is, unlike ORB's shared RefreshMeta component, which already gives users a relative-time trust signal elsewhere in the product.

### Dependencies
None -- independently shippable. This is distinct from all five already-ticketed extension items (#4860-#4863, #4866); none of those touch data-freshness UX, so this can land in any order relative to them.

### Requirements
1. In apps/gittensory-miner-extension/options.js, when a contributor pastes a new discover-run JSON blob and rankedCandidates is saved to chrome.storage.local, also write a savedAt timestamp (e.g. Date.now()) to the same storage entry/key structure.
2. Ensure savedAt is updated every time rankedCandidates is overwritten (re-paste), not just on first save.
3. In opportunity-badge.js, read savedAt from chrome.storage.local alongside rankedCandidates.
4. Render a relative-time label (e.g. 'last synced 3m ago', 'last synced 2h ago') next to the badge, using the same relative-time formatting/trust-signal visual pattern as ORB's shared RefreshMeta component (mirror its thresholds/format, e.g. seconds/minutes/hours/days buckets).
5. Handle the case where savedAt is missing entirely (e.g. data saved before this change shipped, or storage was cleared) -- the badge must degrade gracefully (e.g. omit the label or show a neutral 'sync time unknown' state) rather than rendering NaN or crashing.
6. This is UI/storage-schema plumbing only -- do not touch any ranking/scoring logic, governor state, or attempt state; rankedCandidates' contents and meaning must be unchanged, only the accompanying savedAt metadata is new.

### Deliverables / Acceptance Criteria
- [ ] options.js writes savedAt alongside every rankedCandidates save
- [ ] opportunity-badge.js reads savedAt and renders a relative-time 'last synced' label
- [ ] Label formatting mirrors ORB's RefreshMeta relative-time conventions
- [ ] Missing/absent savedAt is handled gracefully with no crash and no NaN/invalid-date rendering
- [ ] Before/after screenshots of the badge (with and without a fresh sync) included in the PR description

### Test Coverage Requirements
[...]

PR #5571, opened 2026-07-13
feat(miner-extension): add a standard icon set to the manifest (#4862)

### LINKED ISSUE #4862: Add icons to the extension manifest
**Problem:** The extension's manifest has no icons defined.

**Area:** AMS / Extension

**Proposal:** Add a standard icon set to the manifest.

**Deliverables:**
- Icons at the standard required sizes, referenced correctly in the manifest.

andriypolanski:
Closes #4862

## Summary

`apps/gittensory-miner-extension/manifest.json` declared no icons, so the browser fell back to a generic default
in the extensions list and the toolbar. This adds a standard MV3 icon set (16/32/48/128) referenced from both the
extension-level `icons` and the toolbar `action.default_icon`, plus a committed, dependency-free generator so the
raster assets are reproducible rather than opaque binaries.

## What's in it

- **`apps/gittensory-miner-extension/icons/icon-{16,32,48,128}.png`** — the standard required icon sizes, as real
  8-bit RGBA PNGs.
- **`apps/gittensory-miner-extension/icons/generate-icons.mjs`** — the generator that produces them. Pure Node
  (only `node:zlib` for PNG DEFLATE), no external deps; draws the extension's mint "loop" mark (`#7ee7bc`) on a
  dark rounded tile (`#111317`), matching the opportunity badge's palette in `styles.css`. Uses 4×4 supersampling
  with alpha-weighted downsampling so the ring and rounded corners stay crisp at 16px. Output is deterministic
  (byte-identical on re-run).
- **`apps/gittensory-miner-extension/manifest.json`** — adds `icons` and `action.default_icon`, each mapping
  `16/32/48/128` to `icons/icon-<size>.png`.
- **`package.json`** — adds `miner:extension:icons` to regenerate the set.
- **`apps/gittensory-miner-extension/README.md`** — documents the icon set and how to regenerate it.

Scope note: this is the AMS/miner contributor extension (`apps/gittensory-miner-extension`), the one the linked
issues (#5192/#5193) and this milestone reference. The separate maintainer overlay
(`apps/gittensory-extension`) is out of scope.

## Testing

- `test/unit/miner-extension-content.test.ts` — a new case asserts `manifest.icons` and
  `manifest.action.default_icon` both declare the four standard sizes at `icons/icon-<size>.png`, and that each
  referenced file exists, begins with the PNG signature, and has an IHDR width/height matching its declared size
  (so a manifest key can never lie about the asset behind it).
- Verified locally (Node 18 sandbox): `node --check` on the generator; ran it and confirmed all four files are
  valid RGBA PNGs at exact sizes; re-ran and confirmed byte-identical (deterministic) output; the manifest is
  valid JSON and the new test assertions pass via a standalone driver; no linter errors.

## Note for CI

`npm run test:ci` / `npm run test:coverage` must be run under Node >= 22.13 (this sandbox is Node 18, so vitest
could not be executed here). The new files live under `apps/**`, outside the Codecov-measured `src/**` / `lib/**`
surface, so they add no patch-coverage burden; the manifest/icon assertions run as part of the existing unit suite.


JSONbored (commented):
Not sure what these icons are, can you please generate something more fitting? Also, why is there an icon generator script included?

Closing, not a fan.

PR #5605, opened 2026-07-13
feat(miner-ui): require a same-origin session cookie on the local API

### LINKED ISSUE #4858: Add auth to the local miner-ui API
**Problem:** The local miner-ui API currently relies on undocumented, implicit loopback-only trust — there's no actual authentication, and this becomes more important once the UI can take real actions (above).

**Area:** AMS / Miner UI

**Proposal:** Add a minimal auth mechanism (e.g. a locally-generated token) to the miner-ui's API, documented clearly.

**Deliverables:**
- A working, documented auth mechanism for the local API.

**Acceptance criteria:**
- [ ] An unauthenticated request to the miner-ui API is rejected.

Scope note: this auth mechanism is for the local, loopback-only miner-ui API only. It must not touch the review gate's own authentication/bypass-resistance mechanisms or any multi-tenant security architecture, which stay maintainer-only.

JSONbored:
## Summary

- Adds `apps/gittensory-miner-ui/vite-auth.ts`: a random token generated once per dev-server process, delivered to the browser as an `HttpOnly; SameSite=Strict` cookie on every response, and required on every `/api/*` request. An unauthenticated request gets `401`.
- Registered as the **first** plugin in `vite.config.ts`, ahead of `runStateApiPlugin`/`portfolioQueueApiPlugin`/`ledgersApiPlugin` — an unauthenticated `/api/*` request is rejected before any of their middlewares run, so any future `/api/*` endpoint (e.g. a write action) is covered automatically with no per-endpoint wiring.
- No client-side fetch code changes needed: because the token rides on the browser's own cookie jar, a browser that has loaded this app's own page (`/`) already carries the cookie on every subsequent same-origin `fetch()` call.
- Documents the mechanism in the app's `README.md`.

## Why a cookie instead of a header/query token

The previous state was undocumented, implicit, and inconsistent: `vite-run-state-api.ts` had its own `isLoopbackAddress` check, but `vite-portfolio-queue-api.ts` and `vite-ledgers-api.ts` had no gate at all. Loopback-IP checking also isn't real authentication — it doesn't stop another local process, or a malicious page open in the same browser, from hitting the API.

`HttpOnly` keeps the token unreachable from any XSS in the SPA itself. `SameSite=Strict` means a cross-origin page — including a DNS-rebinding attempt, which resolves an attacker-controlled hostname to `127.0.0.1` rather than reusing this dev server's own origin — never has the cookie attached automatically by the browser. This satisfies the issue's literal acceptance criterion (an unauthenticated request is rejected) without touching the three existing read-only API files or any client fetcher.

Closes #4858

## Test plan

- [x] `npm --workspace @loopover/ui-miner run typecheck` (via `npm run ui:typecheck`, which builds `@loopover/ui-kit` first)
- [x] `npm --workspace @loopover/ui-miner run lint` — 0 errors (5 pre-existing, unrelated warnings)
- [x] `npm --workspace @loopover/ui-miner run test` — 72/72 passing (15 new in `src/auth.test.ts`, covering cookie parsing edge cases, the 401/fall-through decision, and the actual Vite middleware wiring for both `configureServer` and `configurePreviewServer`)
- [x] `npm --workspace @loopover/ui-miner run build`
- [x] Manually started the real dev server and curled it end-to-end: an unauthenticated `GET /api/portfolio-queue` returns `401` with the documented error body; loading `/` captures the `Set-Cookie`; replaying that cookie on `GET /api/portfolio-queue` returns `200` with the real summary payload.

This change lives entirely under `apps/gittensory-miner-ui/**`, which `vitest.config.ts`'s root `coverage.include` explicitly excludes (`apps/**`) — Codecov's patch gate does not apply here, matching the precedent set by #5479 (the PR this issue was blocked on).

### DIFF
--- a/apps/gittensory-miner-ui/README.md
+++ b/apps/gittensory-miner-ui/README.md
@@ -15,3 +15,14 @@ Phase 6 data views (run history, portfolio cards) land in follow-up issues after
 | Env var                     | Required | Description                                                                                                                                                                                                                                            |
 | --------------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `VITE_MINER_UI_GRAFANA_URL` | No       | If set (and non-empty), renders a footer link to your ORB/Grafana dashboard at this URL. Unset ⇒ no link. Must be `VITE_`-prefixed so Vite exposes it to the client bundle. It is a plain navigational link — no token or credential is ever appended. |
+
+## Local API authentication
+
+`/api/*` (run-state, portfolio-queue, ledgers, and any future endpoint under that prefix) requires a
+same-origin session cookie — an unauthenticated request is rejected with `401`. The dev/preview server
+(`vite-auth.ts`) generates a random token once per process and sets it as an `HttpOnly; SameSite=Strict`
+cookie on every response; a browser that has loaded this app's own page (`/`) already carries the cookie
+automatically on every subsequent same-origin `fetch()` call, so none of the client-side data fetchers need
+to know about it. A request from another local process, or from a different page/origin the user has open
+(including a DNS-rebinding attempt), has no way to obtain the cookie and is rejected. There is nothing to
+configure — this is always on for both `vite dev` and `vite preview`.
--- a/apps/gittensory-miner-ui/src/auth.test.ts
+++ b/apps/gittensory-miner-ui/src/auth.test.ts
@@ -1,0 +1,150 @@
+import { describe, expect, it } from "vitest";
+
+import { authPlugin, handleAuthRequest, isAuthenticatedRequest } from "../vite-auth";
+
+const TOKEN = "deterministic-test-token";
+
+describe("isAuthenticatedRequest (#4858)", () => {
+  it("rejects a request with no Cookie header at all", () => {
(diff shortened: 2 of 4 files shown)


JSONbored:
Fixed in 01c0ad505. `Set-Cookie` was applied unconditionally before the auth check ran, so an unauthenticated `/api/*` request could read the valid token straight off its own 401 and replay it — you're right, that defeats the mechanism entirely. Reordered so the auth check runs first: a rejection short-circuits before any `setHeader("Set-Cookie", ...)` call, and the cookie is only stamped on a response the caller is already authorized to receive (an authenticated `/api/*` request, or any non-`/api/*` page/asset load).

Added a regression test asserting `Set-Cookie` is absent from the 401 response, plus verified live: curled `/api/portfolio-queue` unauthenticated and confirmed no `Set-Cookie` header on the 401, then confirmed the legitimate page-load → cookie → authenticated-request flow still works end-to-end.

The `ui:version-audit` CI failure (MCP npm dist-tag drift) looks unrelated to this diff — #5608 (opened around the same time, untouched by this bug) already shows that check passing on a later run, so it looks like transient/external drift that should clear on this push's fresh CI run.

Issue #5681, opened 2026-07-14
Wire AMS's dead orb-export.js telemetry stub to a real central ingest endpoint


JSONbored:
### Problem
`packages/gittensory-miner/lib/orb-export.js` already builds a correctly-shaped, HMAC-anonymized batch of PR-outcome telemetry mirroring Orb's own self-host collector — but per its own comments (lines 16-17, 161-163) it deliberately never performs the network send. AMS today has zero visibility into miner-side growth/usage/outcome metrics beyond what an operator can see on their own single machine, unlike Orb's `orb_signals`/`orb_pr_outcomes` central tables, which already give gittensory/LoopOver real adoption and outcome data for Orb.

### Area
gittensory-miner + central ingest (src/orb)

### Proposal
Finish wiring `orb-export.js`'s existing batch-building logic to a real network send, mirroring Orb's self-host collector's cadence and endpoint pattern (`ORB_COLLECTOR_URL`, default `https://api.loopover.ai/v1/orb/ingest`). Decide and implement the central-side ingest surface during build: either a new `ams_signals`/`ams_instances` table pair mirroring `orb_signals`/`orb_instances`, or a shared `product` discriminator column added to the existing tables so one ingest pipeline and one `analytics.ts`-style aggregation serves both products without duplicating the D1 schema — prefer the discriminator-column approach unless a concrete field mismatch forces separate tables. Reuse the shared anonymization primitive from the companion extraction issue rather than reimplementing HMAC logic a second time.

### Deliverables
- `orb-export.js` wired to a real `fetch` POST against the ingest endpoint, using the shared anonymization primitive
- A contiguous `migrations/NNNN_*.sql` for whichever schema decision is made
- Central ingest route (mirroring `src/orb/ingest.ts`'s validation/insert pattern) accepting AMS's batch shape
- `docs/observability.md` updated to describe the new central export path alongside the existing local-only Prometheus/Grafana docs
- Full unit test coverage for the new send path: success, network failure fails open (matches the existing local-only paths' failure posture), opt-out respected

### Resources
- `packages/gittensory-miner/lib/orb-export.js` (existing dead stub to finish)
- `src/selfhost/orb-collector.ts` + `src/orb/ingest.ts` (Orb's live send/ingest pattern to mirror)
- The companion shared-anonymization-primitive issue (sequence this after it merges)

### Boundaries
[...]

PR #5683, opened 2026-07-14
refactor(engine): extract shared telemetry-anonymization primitive

### LINKED ISSUE #5680: Extract shared telemetry-anonymization primitive into @loopover/engine
### Problem
Orb's self-host telemetry collector hand-rolls its own per-instance-secret HMAC anonymization (`getOrCreateAnonSecret`/`hmacField`, `src/selfhost/orb-collector.ts:63-93`) with no reusable module — confirmed no other subsystem imports it as a client library. AMS is about to need the identical pattern (per-instance secret + HMAC'd identifiers) to safely export its own telemetry without leaking identifiable repo/PR data (see the companion issue wiring `orb-export.js` live). Building AMS's version independently would put two independently-maintained implementations of the same security-sensitive anonymization primitive in the codebase — a divergence between them (weaker hash, reused secret, different field set) would be a real privacy bug, not style debt.

### Area
gittensory-engine / shared core

### Proposal
Extract Orb's existing per-instance-secret + HMAC-anonymization logic into a new pure, DI-based module in `packages/gittensory-engine/src/telemetry/anonymize.ts`, following the same "engine computes, product persists" split already used for governor/portfolio-queue. The engine function takes a secret and a raw value and returns the anonymized value — pure, no I/O, no file access. Each product keeps its own local secret storage/generation (paths differ per product), calling into the shared primitive only for the hash itself. Re-wire `orb-collector.ts` onto the shared primitive with zero behavior change, verified by a regression test asserting identical output pre/post-refactor for the same secret+input.

### Deliverables
- `packages/gittensory-engine/src/telemetry/anonymize.ts` — pure `hmacAnonymize(secret, value)` plus any small pure helpers `orb-collector.ts` needs, with full engine-package test coverage (node:test)
- `orb-collector.ts` refactored to import and call the shared primitive instead of its inline HMAC code
- Root vitest mirror test (this repo's `gittensory-engine` node:test is invisible to Codecov otherwise)

### Resources
- `src/selfhost/orb-collector.ts:63-93` (existing implementation to extract)
- `packages/gittensory-engine/src/governor/self-plagiarism.ts` (recent precedent for this exact engine-side pure-function extraction pattern)

### Boundaries
- Extraction and refactor only — must NOT change Orb's anonymization algorithm, secret rotation policy, or opt-out semantics.
- Must NOT touch AMS's `orb-export.js` — wiring AMS onto the shared primitive is the separate follow-up issue below, since it also needs its own local-secret storage and network-send logic built first.

JSONbored:
## Summary
- Orb's self-host collector (`src/selfhost/orb-collector.ts`) hand-rolled its own per-instance-secret HMAC anonymization with no reusable module.
- AMS is about to need the identical pattern to safely export its own telemetry (#5681) -- extracting it now means one implementation instead of two independently-maintained copies of the same security-sensitive logic.
- New `packages/gittensory-engine/src/telemetry/anonymize.ts` exports `generateAnonSecret`/`hmacAnonymize` (pure, no I/O); `orb-collector.ts` re-wired onto it with a fixed-vector regression test asserting identical output to the pre-extraction implementation.

## Test plan
- [x] `npm run build --workspace @loopover/engine` clean
- [x] Engine's own `node:test` suite: 550/550 pass
- [x] `npx tsc --noEmit --incremental false` clean
- [x] Root vitest: `test/unit/engine-telemetry-anonymize.test.ts` + `test/unit/selfhost-orb-collector.test.ts` (21/21 pass, confirms zero behavior change to Orb's live export)
- [x] `npm run test:coverage` (unsharded): both changed files at 100% line/branch/function coverage per `coverage/lcov.info`

Closes #5680

### DIFF
--- a/packages/gittensory-engine/src/index.ts
+++ b/packages/gittensory-engine/src/index.ts
@@ -672,3 +672,8 @@ export {
   SECRET_PATTERNS,
   redactSecrets,
 } from "./subprocess-env.js";
+
+// Shared telemetry-anonymization primitive (#5680) — one source of truth for the per-instance-secret HMAC
+// hashing both Orb's self-host collector and AMS's export path use before repo/PR identifiers leave the
+// instance.
+export { generateAnonSecret, hmacAnonymize } from "./telemetry/anonymize.js";
--- a/packages/gittensory-engine/src/telemetry/anonymize.ts
+++ b/packages/gittensory-engine/src/telemetry/anonymize.ts
@@ -1,0 +1,29 @@
+// Shared telemetry-anonymization primitive (#5680): one source of truth for the per-instance-secret HMAC
+// hashing every self-hosted product (Orb's `src/selfhost/orb-collector.ts`, and AMS's own export path) uses
+// to anonymize repo/PR identifiers before they leave the instance. Extracted out of Orb-only code so a second,
+// independently-maintained implementation never drifts from this one -- a weaker hash or a reused secret in
+// only one of the two would be a real privacy bug, not style debt.
+//
+// Deliberately narrow: secret PERSISTENCE (where/how it's stored -- D1 for Orb, local SQLite for AMS) stays in
+// each product's own store, since that's genuinely different per product. Only the pure hash/generate math
+// lives here.
+import { createHmac, randomBytes } from "node:crypto";
+
+/**
+ * Generate a fresh, single-purpose 256-bit anonymization secret (64 hex chars). Each product persists this
+ * once per instance in its own store and reuses it on every export, so the same raw value always hashes the
+ * same way. Never derived from, or shared with, any other credential (App private keys, webhook secrets) --
+ * key separation means a leaked anonymization secret can't be used to forge or decrypt anything else.
+ */
+export function generateAnonSecret(): string {
+  return randomBytes(32).toString("hex");
+}
+
+/**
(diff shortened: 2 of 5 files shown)


JSONbored:
Re: the flagged "blocker" about the relative-path import being the likely cause of the `build + boot smoke test` failure — that's not what happened here.

The actual failure (see the job log) was `ERROR: write /app/node_modules/@opentelemetry/sdk-logs/build/.../InMemoryLogRecordExporter.d.ts: no space left on device` during Docker layer import — a runner disk-space flake, unrelated to any source change. I re-ran the identical job with zero code changes and it passed clean (`build + boot smoke test  pass  4m47s`), which rules out the import path as the cause: a real build-time defect wouldn't fix itself on an unchanged re-run.

The relative-path import itself (`../../packages/gittensory-engine/src/telemetry/anonymize.js`) is also this codebase's existing, deliberate convention for `src/` consumers of engine code, not something introduced by this PR — see `src/signals/reward-risk.ts:22`, `src/signals/duplicate-winner.ts:15`, and `src/mcp/issue-rag.ts:6`, all importing engine internals the same way (the published `@loopover/engine` package is what `packages/gittensory-miner` uses instead, since that's a separately-published consumer). Switching this one call site to the package import would be inconsistent with every sibling file in `src/selfhost`/`src/signals`.

Leaving the code as-is. Held for manual review per the guarded-path policy, which is correct — over to a maintainer to merge.

PR #6018, opened 2026-07-15
fix(review): honor ContentRepoSpec.urlFields in directory-index synthesis

### LINKED ISSUE #5941: fix(review): contentSignalSourceFromDirectoryEntry ignores ContentRepoSpec, breaking URL duplicate-detection for custom specs
## Context

`src/review/content-lane/duplicates.ts` synthesizes frontmatter for directory-index entries so they can be run through the same duplicate-detection signal extraction as real content files. Two code paths disagree on which URL fields are considered:

`extractContentDuplicateSignals` (around line 270) correctly honors the caller-supplied `spec: ContentRepoSpec`:
```ts
const urls = [
  ...new Set(
    Object.entries(fields)
      .filter(([key]) => spec.urlFields.has(key))
      .map(([, value]) => normalizeUrl(value, spec))
      .filter(Boolean),
  ),
];
```

But `contentSignalSourceFromDirectoryEntry` (around line 549), which builds the synthetic frontmatter that later gets parsed and fed into `extractContentDuplicateSignals`, does not take a `spec` parameter at all — it always emits only a hardcoded field list:
```ts
const DIRECTORY_ENTRY_URL_SIGNAL_FIELDS = [/* ...9 hardcoded field names... */, "sourceUrl", "websiteUrl"] as const;

function contentSignalSourceFromDirectoryEntry(entry: DirectoryIndexEntry): string {
  const lines = ["---", `title: ${yamlScalar(entry.title)}`, ...];
  for (const field of DIRECTORY_ENTRY_URL_SIGNAL_FIELDS) {
    const value = entry[field];
    if (value) lines.push(`${field}: ${yamlScalar(value)}`);
  }
  lines.push("---", "");
  return lines.join("\n");
}
```

`directoryIndexToSignals` (around line 591) does thread `spec` through to `extractContentDuplicateSignals`, but passes `contentSignalSourceFromDirectoryEntry(entry)` (no `spec`) as the `content` argument:
```ts
extractContentDuplicateSignals(
  { filePath, content: contentSignalSourceFromDirectoryEntry(entry), label: ..., url: ... },
  spec,
)
```

For a self-hosted content repo whose `ContentRepoSpec.urlFields` uses field names other than the 9 hardcoded ones in `DIRECTORY_ENTRY_URL_SIGNAL_FIELDS`, none of that repo's URL fields ever make it into the synthesized frontmatter — `urls` in the resulting `ContentDuplicateSignals` is silently empty for every directory-index entry, breaking URL/domain-based duplicate detection against the accepted-entries corpus for any non-default `ContentRepoSpec`.

`test/unit/content-lane-duplicates.test.ts` only tests removing a URL field from the default spec, never a differently-named custom field, so this drift is untested.

## Requirements

- `contentSignalSourceFromDirectoryEntry` must accept a `spec: ContentRepoSpec` parameter (defaulting to `AWESOME_CLAUDE_CONTENT_SPEC`, matching the existing default used elsewhere in this file) and emit frontmatter lines for the fields named in `spec.urlFields` instead of (or in addition to, if a safe superset approach is simpler) the hardcoded `DIRECTORY_ENTRY_URL_SIGNAL_FIELDS` list.
- `directoryIndexToSignals` must pass its own `spec` argument through to `contentSignalSourceFromDirectoryEntry`, so the synthesized frontmatter and the downstream `extractContentDuplicateSignals` filtering agree on which fields are URL signals.
- Preserve behavior for the default `AWESOME_CLAUDE_CONTENT_SPEC` case exactly (same synthesized frontmatter output as today) — this is a fix for the custom-spec case, not a behavior change for the default lane.
- Keep `DirectoryIndexEntry`'s field access safe for entries missing a given custom field (same `if (value) ...` guard pattern already in use).

## Deliverables

- [ ] `contentSignalSourceFromDirectoryEntry(entry, spec)` synthesizes URL frontmatter lines using `spec.urlFields` instead of a fixed hardcoded list.
- [ ] `directoryIndexToSignals` passes its `spec` argument through to `contentSignalSourceFromDirectoryEntry`.
- [ ] A regression test in `test/unit/content-lane-duplicates.test.ts` using a `ContentRepoSpec` with a custom (non-default) `urlFields` set, asserting the resulting `ContentDuplicateSignals.urls` for a directory-index entry actually contains the custom field's URL.
- [ ] A test confirming the default-spec output is unchanged from before this fix (no regression for `AWESOME_CLAUDE_CONTENT_SPEC` callers).

## Test Coverage Requirements

Aim for 99%+ Codecov patch coverage (100% including the new branch/invariant) on the touched lines in `src/review/content-lane/duplicates.ts`. This is a fix for silently-empty duplicate-detection signals on a real self-host configuration path, so the custom-spec regression test is required, not just incidental line coverage.

## Expected Outcome

A self-hosted content repo with a custom `ContentRepoSpec.urlFields` gets working URL/domain-based duplicate detection against accepted directory-index entries, instead of `urls` being silently empty because the synthesized frontmatter only ever emitted the default spec's hardcoded field names.

## Links & Resources

- `src/review/content-lane/duplicates.ts` (`extractContentDuplicateSignals`, ~line 260-280; `DIRECTORY_ENTRY_URL_SIGNAL_FIELDS` + `contentSignalSourceFromDirectoryEntry`, ~line 535-561; `directoryIndexToSignals`, ~line 566-599)
- `src/review/content-lane/content-repo-spec.ts` (`ContentRepoSpec`, `AWESOME_CLAUDE_CONTENT_SPEC`)
- `test/unit/content-lane-duplicates.test.ts` (existing test suite)

jsdevninja:
## Summary

- `contentSignalSourceFromDirectoryEntry` now takes a `ContentRepoSpec` and emits URL frontmatter from `spec.urlFields` instead of a hardcoded field list.
- `directoryIndexToSignals` passes its `spec` through so synthesis and `extractContentDuplicateSignals` agree on URL keys.
- Regression tests cover a custom non-default `urlFields` set (corpus `urls` must contain the custom field) and unchanged default `AWESOME_CLAUDE_CONTENT_SPEC` camelCase behavior.

## Scope

- [x] Conventional Commit title
- [x] Focused bug fix for #5941
- [x] Linked open issue

## Validation

- [x] `npx vitest run test/unit/content-lane-duplicates.test.ts` — 83 passed
- [x] Coverage on `src/review/content-lane/duplicates.ts` — 100% stmts/branches/funcs/lines for this file under that suite

## Safety

- [x] No secrets / wallets / scoring surfaces
- [x] No changelog / site edits

## UI Evidence

N/A — review content-lane backend fix; no UI change.

Closes #5941

### DIFF
--- a/src/review/content-lane/duplicates.ts
+++ b/src/review/content-lane/duplicates.ts
@@ -537,8 +537,10 @@ function yamlScalar(value: unknown): string {
  * URL lines come from `spec.urlFields` (not a hardcoded list) so they agree with
  * `extractContentDuplicateSignals`'s filtering — custom ContentRepoSpec URL keys must
  * reach the synthesized frontmatter or corpus `urls` stay silently empty (#5941).
+ *
+ * Exported for unit tests so the default-`spec` parameter branch is exercised (#5941 Codecov).
  */
-function contentSignalSourceFromDirectoryEntry(
+export function contentSignalSourceFromDirectoryEntry(
   entry: DirectoryIndexEntry,
   spec: ContentRepoSpec = AWESOME_CLAUDE_CONTENT_SPEC,
 ): string {
--- a/src/review/content-lane/duplicates.ts
+++ b/src/review/content-lane/duplicates.ts
@@ -528,25 +528,20 @@ export type DirectoryIndexEntry = Record<string, unknown> & {
   canonicalUrl?: unknown;
 };
 
-/** URL signal fields read off a directory entry, in order. */
-const DIRECTORY_ENTRY_URL_SIGNAL_FIELDS = [
-  "documentationUrl",
-  "docsUrl",
-  "downloadUrl",
-  "githubUrl",
-  "packageUrl",
-  "repoUrl",
-  "repositoryUrl",
-  "sourceUrl",
-  "websiteUrl",
-] as const;
-
 function yamlScalar(value: unknown): string {
   return JSON.stringify(String(value || ""));
 }
 
-/** Synthesize the per-entry frontmatter block. */
-function contentSignalSourceFromDirectoryEntry(entry: DirectoryIndexEntry): string {
+/**
+ * Synthesize the per-entry frontmatter block for directory-index corpus extraction.
+ * URL lines come from `spec.urlFields` (not a hardcoded list) so they agree with
+ * `extractContentDuplicateSignals`'s filtering — custom ContentRepoSpec URL keys must
+ * reach the synthesized frontmatter or corpus `urls` stay silently empty (#5941).
+ */
+function contentSignalSourceFromDirectoryEntry(
+  entry: DirectoryIndexEntry,
+  spec: ContentRepoSpec = AWESOME_CLAUDE_CONTENT_SPEC,
+): string {
   const lines = [
     "---",
     `title: ${yamlScalar(entry.title)}`,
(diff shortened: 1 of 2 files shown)


JSONbored:
Please, do not use draft - it messes with the bot, wastes resources, and generally slows down the pipeline/creates potential merge conflicts due to risks with out of order reviews/merges.

PR #123, opened 2026-05-30
feat(data): add contributor reconciliation reports

oktofeesh1:
## Summary
- add private contributor reconciliation reports
- compare official contributor totals with cached GitHub context
- separate maintainer-lane history from normal contributor evidence

## What changed
- builds per-repo official, cached, and effective totals
- explains mismatch reasons such as lookback, stale/closed context, unavailable official data, and maintainer-owned history
- threads reconciliation into contributor outcome history with deterministic ordering
- keeps readiness integration fixtures current so freshness SLO assertions remain deterministic in CI

## Why
- private decision data should explain why official and cached GitHub histories differ without inflating contributor evidence.

## Validation
- git diff --check
- npx tsc --noEmit
- npx vitest run test/unit/signals-v2.test.ts
- npm run test:coverage
- GitHub validate, Contributor trust, and Superagent Security Scan pass on the latest head

Closes #108

PR #6050, opened 2026-07-15
fix(review): track lockfile entry nesting so a nested dependencies object can't hide tamper (#5837)

### LINKED ISSUE #5837: fix(review): lockfile-tamper check loses entry tracking across a nested dependencies sub-object, evadable by key reordering
## Context

`scanPackageLockPatch` in `src/review/lockfile-tamper.ts` (function starts line 114) is a heuristic line-based scanner for `package-lock.json` diffs that flags the classic supply-chain tell: a `resolved`/`integrity` value changed without that same lockfile entry's own `"version"` field genuinely changing (see the module's header comment for the full rationale).

It tracks "which package entry am I currently inside" with a single `currentEntryKey`/`currentPackageName` pair and a `sawPackagesEntry` flag. The state machine resets `currentEntryKey`/`currentPackageName` to `null` in two places:
- Line 150-151: when an object-header line's key is neither a `node_modules/...` path nor (once `sawPackagesEntry` is true) a recognized container key.
- Line 156: unconditionally, on any closing-brace line (`}` or `},`).

The bug: a real `package-lock.json` entry frequently contains its **own nested** `"dependencies"` / `"devDependencies"` / `"optionalDependencies"` sub-object (this repo's own `package-lock.json` has many such entries, e.g. `apps/loopover-miner-ui/node_modules/@asamuzakjp/css-color`). Once inside the `"packages"` tree (`sawPackagesEntry === true`), hitting that nested `"dependencies": {` header falls into the `else` branch at line 150 (because `npmPackageFromNodeModulesPath("dependencies")` returns `null` and `CONTAINER_KEYS.has("dependencies")` is only checked when `!sawPackagesEntry`), resetting `currentEntryKey`/`currentPackageName` to `null` **without ever restoring them** when the nested object closes. Any `resolved`/`integrity`/`version` line for that same outer entry appearing *after* the nested sub-object in the diff is then silently skipped by the `if (!currentEntryKey || !currentPackageName || line.sign === " ") continue;` guard (line 158-159) — exactly the fields this module exists to police.

Today's real npm output happens to always order `version`/`resolved`/`integrity` *before* `dependencies` within an entry (confirmed against this repo's own `package-lock.json`), so legitimate diffs don't trip this. But nothing in the parser enforces that ordering, and an attacker crafting a hand-edited tamper diff specifically to evade this exact check would simply reorder the entry's keys so `dependencies` precedes `resolved`/`integrity` — defeating the tamper detector entirely for that entry. `test/unit/lockfile-tamper.test.ts` has no case with a `node_modules/...` entry containing a nested `dependencies` sub-object positioned before its own `resolved`/`integrity`/`version` fields, so this branch is untested today.

## Requirements

- Fix `scanPackageLockPatch` so that a nested `"dependencies"`/`"devDependencies"`/`"optionalDependencies"` sub-object inside a `node_modules/...` entry does not permanently lose track of that entry — subsequent `resolved`/`integrity`/`version` lines belonging to the same outer entry (appearing after the nested object closes) must still be attributed to it, regardless of key order within the entry.
- The fix must be a parsing/state-tracking correction only — it must not change `LockfileTamperCandidate`'s shape, the module's public API (`scanPackageLockPatch`'s exported callers), or how a detected candidate is turned into a finding/blocker (`isConfiguredGateBlocker` / gate-mode wiring stays untouched).
- Preserve the existing behavior for the legacy lockfileVersion 1 bare-name `"dependencies"` container tree (the `!sawPackagesEntry` branch) exactly as documented in the module's header comment — this fix is scoped to the `node_modules/...`-keyed lockfileVersion 2/3 path.
- Must not regress the existing entry-keying fix from `#2692`/`#2563` (full `node_modules/...` path as the map key, not bare package name) — the fix should track entry *nesting depth*, not just presence/absence of a current key.

## Deliverables

- [ ] `scanPackageLockPatch` correctly attributes `resolved`/`integrity`/`version` lines to their owning entry even when a nested `dependencies`/`devDependencies`/`optionalDependencies` sub-object appears earlier in that same entry's diff hunk.
- [ ] A regression test in `test/unit/lockfile-tamper.test.ts` with a synthetic diff where a `node_modules/foo` entry's `dependencies` sub-object appears before its `resolved`/`integrity`/`version` lines (key order reversed from real npm output), asserting the tamper signal (`resolvedOrIntegrityChanged` without a matching `versionChanged`) is still detected.
- [ ] A second test covering the combination noted by the audit: two different nested entries sharing a bare package name (e.g. `node_modules/foo` and `node_modules/bar/node_modules/foo`), each also carrying its own `dependencies` sub-object, to confirm the entry-keying fix from `#2692` and this fix compose correctly.

## Test Coverage Requirements

Aim for 99%+ Codecov patch coverage (100% including the new branch) on the touched lines in `src/review/lockfile-tamper.ts` (`src/**` is in `coverage.include`). This is a security-relevant regression fix — a dedicated test for the exact bypass scenario (reordered keys defeating detection) is required, not just incidental line coverage.

## Expected Outcome

A `package-lock.json` diff where a tampered entry's `resolved`/`integrity`/`version` lines appear after that entry's own nested `dependencies` object is now still correctly flagged by the lockfile-tamper check — closing a state-tracking gap that could otherwise let a reordered-key tamper diff slip past this advisory/gate signal undetected.

## Links & Resources

- `src/review/lockfile-tamper.ts` (`scanPackageLockPatch`, function starting line 114; the reset bug at lines 145-158; `entryFor`, lines 120-134; `CONTAINER_KEYS`, near the top of the file)
- `test/unit/lockfile-tamper.test.ts` (existing test suite)
- Prior related fix: `#2563` (original lockfile-tamper-risk gate check) and its `#2692` gate-review follow-up (full-path entry keying, documented in this same function's header comment)

shin-core:
## Summary

Closes #5837.

`scanPackageLockPatch` in `src/review/lockfile-tamper.ts` is the heuristic line-based scanner that flags the
classic supply-chain tell: a `resolved`/`integrity` value changed for a `package-lock.json` entry **without** that
same entry's own `"version"` genuinely changing. It tracked "which entry am I inside" with a single flat
`currentEntryKey`/`currentPackageName` pointer.

A real `node_modules/...` entry frequently contains its **own nested** `dependencies` /
`devDependencies` / `optionalDependencies` sub-object. Once inside the `packages` tree, hitting that nested
container header fell into the `else` branch and reset the pointer to `null` — and it was **never restored** when
the nested object closed. Any `resolved`/`integrity`/`version` line for the same outer entry appearing *after* the
nested sub-object was then silently skipped by the current-entry guard.

Today's npm output always orders `version`/`resolved`/`integrity` **before** `dependencies` within an entry, so
legitimate diffs never trip this. But nothing enforces that ordering: an attacker hand-crafting a tamper diff
could simply **reorder the entry's keys** so `dependencies` precedes `resolved`/`integrity`, defeating the tamper
detector entirely for that entry.

**Fix:** replace the flat pointer with a **stack of brace scopes**. Each object-header line pushes a frame (an
entry, or `null` for a container/root/other header); each closing brace pops one. Closing a nested sub-object
therefore restores the **enclosing entry** as the current one, so subsequent `resolved`/`integrity`/`version`
lines are still attributed to it regardless of key order. This is a parsing/state-tracking correction only:

- Entries stay keyed by their full `node_modules/...` path (#2692), so two same-named entries under different
  nesting paths remain independent.
- The legacy lockfileVersion-1 bare-name `dependencies` tree (`!sawPackagesEntry` branch) is unchanged.
- `LockfileTamperCandidate`'s shape, the module's public API, and the gate/finding wiring
  (`isConfiguredGateBlocker`) are untouched.

## Scope

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally; **`codecov/patch` requires ≥99% coverage of the lines AND branches you changed** (aim for **100%** on your diff so CI variance does not fail near the threshold). Global coverage is a non-blocking trend with a loose 90% backstop, not the gate.
- [x] `npm run test:workers`
- [x] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [x] `npm run ui:lint`
- [x] `npm run ui:typecheck`
- [x] `npm run ui:build`

- Ran the full `npm run test:ci` gate plus `npm audit --audit-level=moderate` (0 vulnerabilities). `scanPackageLockPatch` is at 100% branch coverage on the changed lines (unsharded `npm run test:coverage`); the two new regression tests were confirmed to FAIL against the pre-fix source, so they genuinely pin the bug.

## Safety

Notes on the Safety boxes: this is a backend-only parsing correction to a deterministic supply-chain scanner — no UI, API/OpenAPI, or auth/CORS/session surface. The negative-path testing that applies is the scanner's own: the added tests cover the tamper-detected path (reordered keys), and the existing suite already covers the no-finding legitimate-bump path (unchanged and still passing). The fix strengthens a security check rather than relaxing one.

## UI Evidence

Not applicable — backend-only change to a diff parser; no visible UI, frontend, docs, or extension surface.

## Notes

- Regression tests (`test/unit/lockfile-tamper.test.ts`): (1) a `node_modules/lodash` entry whose `dependencies` sub-object precedes its `resolved`/`integrity`/`version` lines (key order reversed from real npm output) — asserts the unbumped-integrity tamper signal is still detected; (2) two distinct entries sharing the bare name `foo` (`node_modules/foo` and `node_modules/bar/node_modules/foo`), each with its own `dependencies` sub-object — the legitimate bump on the top-level `foo` does not mask the unbumped integrity edit on the nested `foo`, confirming the #2692 full-path keying and this nesting fix compose. Both new tests were verified to fail on the pre-fix source.

### DIFF
--- a/src/review/lockfile-tamper.ts
+++ b/src/review/lockfile-tamper.ts
@@ -113,8 +113,14 @@ function versionChanged(removed: string | undefined, added: string | undefined):
  *  actively maintained repo's lockfile and is out of scope here. */
 function scanPackageLockPatch(path: string, patch: string): LockfileTamperCandidate[] {
   const byEntry = new Map<string, MutableCandidate>();
-  let currentEntryKey: string | null = null;
-  let currentPackageName: string | null = null;
+  // A single flat currentEntryKey pointer loses the outer entry when a `node_modules/...` entry contains its own
+  // nested `dependencies`/`devDependencies`/`optionalDependencies` sub-object: the nested container header reset
+  // the pointer to null and never restored it, so any resolved/integrity/version line for the SAME outer entry
+  // appearing after the nested object was silently skipped — an attacker could reorder an entry's keys to put
+  // `dependencies` before `resolved`/`integrity` and evade the tamper check entirely (#5837). Track a stack of
+  // brace scopes instead: each object-header line pushes a frame (an entry, or null for a container/root/other),
+  // each closing brace pops one, so closing a nested sub-object restores the enclosing entry as the current one.
+  const scopeStack: Array<{ entryKey: string; packageName: string } | null> = [];
   let sawPackagesEntry = false;
 
   const entryFor = (entryKey: string, packageName: string): MutableCandidate => {
@@ -140,23 +146,26 @@ function scanPackageLockPatch(path: string, patch: string): LockfileTamperCandid
       const key = objectHeader[1]!;
       const nodeModulesPackage = npmPackageFromNodeModulesPath(key);
       if (nodeModulesPackage) {
-        currentEntryKey = key;
-        currentPackageName = nodeModulesPackage;
+        scopeStack.push({ entryKey: key, packageName: nodeModulesPackage });
         sawPackagesEntry = true;
       } else if (!sawPackagesEntry && !CONTAINER_KEYS.has(key)) {
-        currentEntryKey = key;
-        currentPackageName = key;
+        scopeStack.push({ entryKey: key, packageName: key });
       } else {
-        currentEntryKey = null;
-        currentPackageName = null;
+        // A container/root/unrecognized header (incl. a nested dependencies sub-object) — not an entry itself,
+        // but still a brace scope, so push a null frame that a matching closing brace will pop back off.
+        scopeStack.push(null);
       }
       continue;
     }
     if (body === "}" || body.startsWith("},")) {
-      currentEntryKey = null;
-      currentPackageName = null;
+      scopeStack.pop();
     }
-    if (!currentEntryKey || !currentPackageName || line.sign === " ") continue;
+    // The current entry is the innermost open scope IF that scope is an entry (we're directly in its body). A
+    // resolved/integrity/version line nested inside a sub-object (top frame null) is correctly not attributed.
+    const current = scopeStack.length > 0 ? scopeStack[scopeStack.length - 1] : null;
+    if (!current || line.sign === " ") continue;
+    const currentEntryKey = current.entryKey;
+    const currentPackageName = current.packageName;
 
     const resolvedMatch = /^"resolved"\s*:\s*"([^"]*)"/.exec(body);
     const integrityMatch = /^"integrity"\s*:\s*"([^"]*)"/.exec(body);
--- a/test/unit/lockfile-tamper.test.ts
+++ b/test/unit/lockfile-tamper.test.ts
@@ -95,6 +95,71 @@ describe("lockfileTamperRiskFinding", () => {
     expect(finding?.detail).toContain("lodash");
   });
 
+  // #5837: a nested dependencies sub-object inside an entry, appearing BEFORE the entry's own
+  // resolved/integrity/version lines (key order reversed from real npm output), must not make the parser lose
+  // track of the outer entry — otherwise a reordered-key tamper diff evades detection entirely.
+  it("still detects tamper when the entry's dependencies sub-object precedes its resolved/integrity/version (regression for #5837)", () => {
+    const lockPatch = [
+      '@@ -100,12 +100,12 @@',
+      '     "node_modules/lodash": {',
+      '       "dependencies": {',
+      '-        "helper": "^1.0.0"',
+      '+        "helper": "^1.0.1"',
+      '       },',
+      '-      "version": "4.17.20",',
+      '-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.20.tgz",',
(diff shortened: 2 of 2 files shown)


shin-core:
Closing as a duplicate — #5837 was already resolved upstream by #6038 (commit 497505c2), which landed while this was in progress. The two fixes are functionally equivalent (nesting-depth tracking so a nested dependencies sub-object no longer drops the outer entry). No action needed.

Issue #6297, opened 2026-07-16
test(ci): check-mcp-package.mjs has no unit-test seam, unlike its sibling check-miner-package.mjs


JSONbored:
## Context

`scripts/check-miner-package.mjs` exports `validateMinerPackFileList`/`runMinerPackCheck` with injectable `readContent`/`pack` parameters, letting `test/unit/check-miner-package.test.ts` exercise its logic without a real `npm pack`. `scripts/check-mcp-package.mjs` is a monolithic top-level script (`spawnSync` + `throw` at module scope, no exports, no test file) — exactly why the sibling "hand-duplicated regex" issue (filed alongside this one) went uncaught by any test.

## Requirements

- Refactor `check-mcp-package.mjs` to expose its core validation logic as testable, injectable functions, matching `check-miner-package.mjs`'s existing pattern.
- Add a test file (`test/unit/check-mcp-package.test.ts`) covering its actual validation behavior, including the allowlist/forbidden-content checks the sibling issues in this batch touch.

## Test Coverage Requirements
99%+ Codecov patch coverage on the refactored, now-testable logic.

## Deliverables
- [ ] `check-mcp-package.mjs`'s core logic is testable via injectable dependencies, matching `check-miner-package.mjs`.
- [ ] A real test file exercising it.

## Expected Outcome
Future drift in this script's validation logic (like the regex-duplication and allowlist-gap issues found this session) gets caught by CI instead of requiring manual code review to find.

## Links & Resources
- `scripts/check-mcp-package.mjs`
- `scripts/check-miner-package.mjs`, `test/unit/check-miner-package.test.ts` (the pattern to replicate)

PR #6399, opened 2026-07-16
fix(ci): check-mcp-package imports the shared secret-shape detector

### LINKED ISSUE #6290: fix(ci): check-mcp-package.mjs hand-duplicates forbidden-content.mjs's regex instead of importing it
## Context

`scripts/forbidden-content.mjs:1-8` states it's the "Single source of truth for the miner package's secret-shape detector... importing it here rather than hand-duplicating the regex keeps the two byte-for-byte in sync." `scripts/check-mcp-package.mjs:19` re-declares the identical regex literal instead of importing `FORBIDDEN_CONTENT`, while `scripts/check-miner-package.mjs:6,37` imports it correctly. No test asserts the two definitions stay equal.

## Requirements

- Update `check-mcp-package.mjs` to import `FORBIDDEN_CONTENT` from `forbidden-content.mjs` instead of re-declaring the regex, matching `check-miner-package.mjs`'s pattern.
- Add a test asserting the two consumers use the identical exported constant (not just visually-identical literals), so this can't silently drift again.

## Deliverables
- [ ] `check-mcp-package.mjs` imports the shared regex.
- [ ] A test guarding against future drift.

## Expected Outcome
There's exactly one definition of this secret-shape detector, matching the file's own "single source of truth" claim.

## Links & Resources
- `scripts/forbidden-content.mjs:1-8`, `scripts/check-mcp-package.mjs:19`, `scripts/check-miner-package.mjs:6,37`

dhgoal:
## Summary

Closes #6290

- `scripts/forbidden-content.mjs` declares itself the "single source of truth" for the packaged secret-shape detector, and `check-miner-package.mjs` imports it. `check-mcp-package.mjs` instead re-declared the identical regex as its own local literal, so the two could drift apart silently — and nothing asserted they stayed equal.
- `check-mcp-package.mjs` now imports `FORBIDDEN_CONTENT`, matching its sibling. The detector is **stateless** (no `global`/`sticky` flag), so sharing a single instance across both checkers is safe — a `/g` regex would carry `lastIndex` between `.test()` calls and make shared use order-dependent.
- Adds a drift guard covering **both** checkers: each must import the shared constant and declare no local detector of its own.
- **1 file changed + 1 test added; the duplicated literal is deleted, not reworded.**

## Scope

## Validation

- [x] `git diff --check`
- [ ] `npm run actionlint`
- [ ] `npm run typecheck`
- [ ] `npm run test:coverage` locally; **`codecov/patch` requires ≥99% coverage of the lines AND branches you changed** (aim for **100%** on your diff so CI variance does not fail near the threshold). Global coverage is a non-blocking trend with a loose 90% backstop, not the gate.
- [ ] `npm run test:workers`
- [ ] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [ ] `npm run ui:openapi:check`
- [ ] `npm run ui:lint`
- [ ] `npm run ui:typecheck`
- [ ] `npm run ui:build`
- [ ] `npm audit --audit-level=moderate`

- **Verified the guard actually guards**: temporarily restored `main`'s duplicated version and confirmed the new test *fails* against it, then passes with the fix. It is a real drift guard, not a tautology.
- `npm run test:mcp-pack` passes (exit 0) — the real checker still validates the live MCP package end-to-end through the shared constant, so this is a behaviour-preserving swap.
- Ran the affected suites with vitest, all green (**49 tests**): the new `forbidden-content` guard plus `check-miner-package` and `miner-mcp-contract`, the two existing consumers of this constant.
- The new assertions are **source-level by necessity**: both checkers run their `npm pack` dry-run at import time and export nothing, so a test cannot import them to compare the constants by identity. (The missing env-var test seam is separately tracked in #6297 and deliberately out of scope here.)

## Safety

- [ ] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests.
- [ ] API/OpenAPI/MCP behavior is updated and tested where needed.
- [ ] UI changes use live API data or real empty/error/loading states, not production mock/demo fallbacks.
- [ ] Visible UI changes include a `UI Evidence` section below with JPG/JPEG or PNG screenshots arranged as organized, captioned, clickable thumbnails. SVG screenshots are not used as review evidence. Review-only screenshots or recordings are not committed to the repository.
- [ ] Public docs/changelogs are updated where needed; changelogs are only edited for release-prep PRs.

This **strengthens** a packaging guard: the MCP package checker is now provably driven by the same detector as the miner one, so a future edit to the shared pattern can no longer protect one package while leaving the other behind. No credential-shaped literal is added anywhere — the new test asserts on the regex's flags and on import structure, never on secret-shaped sample values.

## Notes

Closes #6290

### DIFF
--- a/scripts/check-mcp-package.mjs
+++ b/scripts/check-mcp-package.mjs
@@ -2,6 +2,7 @@
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
 import { spawnSync } from "node:child_process";
+import { FORBIDDEN_CONTENT } from "./forbidden-content.mjs";
 import { MCP_PACKAGE_ALLOWED_FILE_PATTERNS } from "./mcp-package-allowlist.mjs";
 
 const result = spawnSync("npm", ["pack", "--workspace", "@loopover/mcp", "--dry-run", "--json"], {
@@ -17,7 +18,6 @@ const [pack] = JSON.parse(result.stdout);
 const files = pack.files.map((file) => file.path).sort();
 const allowed = MCP_PACKAGE_ALLOWED_FILE_PATTERNS;
 const forbiddenPath = /(^|\/)(\.dev\.vars|\.env|\.npmrc|.*\.pem|.*private.*key.*|.*secret.*)$/i;
-const forbiddenContent = /(BEGIN (RSA |EC |OPENSSH )?PRIVATE KEY|github_pat_[A-Za-z0-9_]+|gh[pousr]_[A-Za-z0-9_]+|gts_[0-9a-f]{64}|[A-Z0-9_]*(TOKEN|SECRET|PRIVATE_KEY)=)/;
 const stalePackageText = /(private beta|zeronode\.workers\.dev|preview URL)/i;
 
 for (const file of files) {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts: 

```
Use the [web editor](https://github.com/JSONbored/loopover/pull/6399/conflicts) or the command line to resolve conflicts before continuing.

scripts/check-mcp-package.mjs
```

PR #6426, opened 2026-07-16
fix(ci): check-mcp-package imports the shared secret-shape detector

dhgoal:
## Summary

Closes #6290

- `scripts/forbidden-content.mjs` declares itself the "single source of truth" for the packaged secret-shape detector, and `check-miner-package.mjs` imports it. `check-mcp-package.mjs` still declared the same regex as its own local constant, so the two could drift apart with nothing asserting they stayed equal.
- `check-mcp-package.mjs` now imports `FORBIDDEN_CONTENT`, matching its sibling. The detector is **stateless** (no `global`/`sticky` flag), so a single shared instance is safe across both callers — a `/g` regex would carry `lastIndex` between `.test()` calls and make shared use order-dependent.
- Adds a drift guard covering **both** checkers on both axes:
  - **structural** — each imports the shared constant and declares no local copy;
  - **behavioral** — each actually rejects what the shared detector matches, exercised through the injectable seams (`validateMcpPackFileList` / `validateMinerPackFileList`).
- Net: **-1 line of duplication** in the script, plus the guard.

## Scope

- [x] The PR title follows `type(scope): short summary` Conventional Commit format, for example `fix(api): restore profile access checks`.
- [x] This PR is focused and does not mix unrelated backend, UI, MCP, docs, dependency, and deploy changes.
- [x] This follows `CONTRIBUTING.md` and does not reintroduce GitHub Pages, VitePress, `site/`, or `CNAME`.
- [x] I linked a currently open issue this PR resolves (e.g. `Closes #123`) — a linked open issue is required for every contributor PR.

## Validation

- [x] `git diff --check`
- [ ] `npm run actionlint`
- [ ] `npm run typecheck`
- [ ] `npm run test:coverage` locally; **`codecov/patch` requires ≥99% coverage of the lines AND branches you changed** (aim for **100%** on your diff so CI variance does not fail near the threshold). Global coverage is a non-blocking trend with a loose 90% backstop, not the gate.
- [ ] `npm run test:workers`
- [ ] `npm run build:mcp`
- [x] `npm run test:mcp-pack`
- [ ] `npm run ui:openapi:check`
- [ ] `npm run ui:lint`
- [ ] `npm run ui:typecheck`
- [ ] `npm run ui:build`
- [ ] `npm audit --audit-level=moderate`
- [x] New or changed behavior has unit/integration tests for new branches, fallback paths, and sanitizer boundaries

If any required check was skipped, explain why:

[...]

PR #6427, opened 2026-07-16
fix(ci): import shared secret-shape detector in check-mcp-package

### LINKED ISSUE #6290: fix(ci): check-mcp-package.mjs hand-duplicates forbidden-content.mjs's regex instead of importing it

nghetienhiep:
## Summary

Closes #6290

`scripts/forbidden-content.mjs` declares itself the "single source of truth" for the packaged
secret-shape detector, and `check-miner-package.mjs` imports it. `check-mcp-package.mjs` instead
re-declared the identical regex as its own local literal, so the two could drift apart silently —
and nothing asserted they stayed equal.

- `check-mcp-package.mjs` now imports `FORBIDDEN_CONTENT` from `./forbidden-content.mjs`, matching
  its sibling; the duplicated literal is deleted. The detector is **stateless** (no `/g` or `/y`
  flag), so sharing one instance across both checkers is safe — a global/sticky regex would carry
  `lastIndex` between `.test()` calls and make shared use order-dependent.
- Adds a drift guard (`test/unit/forbidden-content.test.ts`) covering **both** consumers: each must
  import the shared constant and declare no local detector of its own, plus a check that the shared
  export stays stateless.

## Scope

- [x] I linked a currently open issue this PR resolves (`Closes #6290`).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:mcp-pack` (exit 0 — the live checker still validates the real MCP package through the shared constant)
- [x] Ran the affected suites (`forbidden-content`, `check-mcp-package`, `check-miner-package`, `miner-mcp-contract`) — all green.
- [x] New drift guard has unit tests for both consumers and the shared export's flags.

Notes on scope of checks:

- **Verified the guard actually guards**: temporarily restored the duplicated local literal (dropping
  the shared import) and confirmed the new test *fails* against it, then passes with the fix. It is a
  real drift guard, not a tautology.
- The assertions are **source-level by necessity**: both checkers run their `npm pack` dry-run behind
  an `import.meta.url` main-guard and export nothing, so a test cannot import them to compare the
  constants by identity.
- The change is confined to `scripts/**` and a `test/**` file, so there is no `src/**` diff for
  `codecov/patch` to measure.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, PATs, private keys, raw trust scores, or private
  rankings are exposed. The new test asserts on the regex's flags and on import structure, never on
  secret-shaped sample values.
- [x] Public GitHub text stays sanitized and low-noise.

This **strengthens** a packaging guard: the MCP package checker is now provably driven by the same
detector as the miner one, so a future edit to the shared pattern can no longer protect one package
while leaving the other behind.

### DIFF
--- a/scripts/check-mcp-package.mjs
+++ b/scripts/check-mcp-package.mjs
@@ -4,10 +4,9 @@ import { join } from "node:path";
 import { spawnSync } from "node:child_process";
 import { fileURLToPath } from "node:url";
 import { MCP_PACKAGE_ALLOWED_FILE_PATTERNS } from "./mcp-package-allowlist.mjs";
+import { FORBIDDEN_CONTENT } from "./forbidden-content.mjs";
 
 const FORBIDDEN_PATH = /(^|\/)(\.dev\.vars|\.env|\.npmrc|.*\.pem|.*private.*key.*|.*secret.*)$/i;
-const FORBIDDEN_CONTENT =
-  /(BEGIN (RSA |EC |OPENSSH )?PRIVATE KEY|github_pat_[A-Za-z0-9_]+|gh[pousr]_[A-Za-z0-9_]+|gts_[0-9a-f]{64}|[A-Z0-9_]*(TOKEN|SECRET|PRIVATE_KEY)=)/;
 const STALE_PACKAGE_TEXT = /(private beta|zeronode\.workers\.dev|preview URL)/i;
 
 export function validateMcpPackFileList(files, readContent) {
--- a/test/unit/forbidden-content.test.ts
+++ b/test/unit/forbidden-content.test.ts
@@ -1,0 +1,37 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, it } from "vitest";
+import { FORBIDDEN_CONTENT } from "../../scripts/forbidden-content.mjs";
+
+// forbidden-content.mjs bills itself as the "single source of truth" for the packaged secret-shape detector.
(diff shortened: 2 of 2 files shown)


JSONbored (commented):
Duplicates #6426 - closing.

PR #6431, opened 2026-07-16
fix(ci): capture review-enrichment's node:test coverage for Codecov

### LINKED ISSUE #6250: fix(ci): review-enrichment's real node:test coverage is invisible to Codecov (83 of 83 files effectively unmeasured)
## Context

`review-enrichment/` (`@loopover/rees`, 83 `.ts` files under `src/`) is a standalone service — deliberately not an npm workspace member (its own `package.json`/`package-lock.json`), deployed independently on Railway. Its real test suite runs via Node's own test runner (`"test:node": "node --test --experimental-strip-types \"test/**/*.test.ts\""`), wired into CI via `npm run rees:test` (part of the root `test:ci` pipeline, `.github/workflows/ci.yml`). Tests genuinely run and — as far as CI green/red goes — pass.

But **Codecov never sees any of it**: the root `vitest.config.ts`'s `coverage.include` array lists exactly one file from this entire package — `review-enrichment/src/analyzers/codeowners.ts` — out of 83. And even that single file isn't imported anywhere in the root `src/**` or `test/**` trees (confirmed via repo-wide grep for `analyzers/codeowners`), meaning nothing vitest runs actually exercises it either — the one "included" file is realistically also unmeasured. Net effect: this entire 83-file package, including a secret-scanning analyzer (`secret-scan.ts`), shows as completely uncovered in Codecov, even though real tests for it exist and run in CI.

This is the same failure mode as a previously-hit issue in this repo where a `node:test`-based suite's real coverage was invisible to Codecov because nothing bridged it into the vitest/lcov pipeline that Codecov actually reads.

## Requirements

- Confirm exactly which of `review-enrichment/test/**/*.test.ts` currently pass in CI (they should — this is a visibility gap, not (necessarily) a real test-coverage gap) before assuming zero tests exist.
- Wire `review-enrichment`'s real coverage into a form Codecov can ingest. The two realistic approaches, in likely order of preference given `review-enrichment`'s deliberate standalone-deployability (do not force it into the root npm workspace or vitest runner as a side effect of this fix):
  1. Run `node --test --experimental-test-coverage` (Node 20+ supports this) with an LCOV-producing reporter, and upload that as a **separate Codecov flag** alongside the existing vitest upload, so Codecov merges both into one project view and — critically — applies patch-coverage gating to `review-enrichment/**` changes too.
  2. If Node's native coverage output can't produce a Codecov-compatible format cleanly, investigate `c8`/`v8` coverage collection wrapping `node --test` instead.
- Remove or fix the single stale `review-enrichment/src/analyzers/codeowners.ts` entry in `vitest.config.ts`'s `coverage.include` — it doesn't reflect reality either way (nothing vitest runs imports it).
- Update `codecov.yml` if a second upload flag is introduced, so patch-coverage gating actually applies to `review-enrichment/**` PRs (today, a `review-enrichment`-only PR likely gets a green/no-op Codecov check regardless of what it changes, since none of its lines are in any tracked report — confirm this is really happening today as part of this issue's own verification).

## Test Coverage Requirements
This issue's own success criterion IS improved coverage visibility — after this lands, a PR touching `review-enrichment/src/**` should show real, non-zero patch coverage in its Codecov check, and a deliberately-added untested branch in that package should cause a real patch-coverage failure (verify this as part of the PR, e.g. with a temporary throwaway branch during testing, not left in the final diff).

## Deliverables
- [ ] `review-enrichment`'s real `node --test` coverage is captured and uploaded to Codecov.
- [ ] `vitest.config.ts`'s stale/misleading single-file `coverage.include` entry for `review-enrichment` is corrected.
- [ ] Verified: a PR touching only `review-enrichment/src/**` now gets a real, non-trivial Codecov patch-coverage check instead of an effectively-vacuous one.

## Expected Outcome
`review-enrichment`'s genuine test coverage (which already exists and runs in CI) becomes visible to Codecov, closing a blind spot where a real regression in an 83-file package — including its secret-scanning analyzer — could currently ship without the coverage gate ever seeing it.

## Links & Resources
- `review-enrichment/package.json` (`test`, `test:node` scripts)
- `vitest.config.ts` (`coverage.include`, currently lists one stale/unexercised file)
- `.github/workflows/ci.yml:330-353` (how `rees:test` already runs in CI today)
- `review-enrichment/test/` (the real, already-passing test suite this makes visible)

nghetienhiep:
## What

`review-enrichment` (`@loopover/rees`) is a standalone, non-workspace package with its own
`node:test` suite that runs in CI (`npm run rees:test`), yet **none** of its 83 `src/**` files were
visible to Codecov. The root `vitest.config.ts` listed exactly one file from it
(`review-enrichment/src/analyzers/codeowners.ts`) — and nothing vitest runs even imports that file,
so it was reported at 0% too. Net effect: a real regression anywhere in this package (including its
`secret-scan.ts` analyzer) could ship without the patch-coverage gate ever seeing it.

This wires the package's genuine coverage into Codecov as a separate `rees` flag, without pulling
`review-enrichment` into the root npm workspace or the vitest runner.

## How

- **`rees:coverage` script (root `package.json`)** — re-runs the existing `node --test` suite under
  `c8`. The tests execute the built `dist/**/*.js`, so `c8` remaps V8 coverage back to
  `review-enrichment/src/**` through the build's source maps (already emitted, `sourceMap` +
  `inlineSources`). Run from the repo root so lcov paths are `review-enrichment/src/...` — exactly
  what Codecov needs, no path rewriting. `c8`'s default exclude drops `*-test.*`/`*.test.*` names,
  which wrongly caught the real `flaky-test.ts` analyzer, so the exclude is narrowed to `**/*.d.ts`
  (test files live in `test/**`, never in the `dist/**` include, so none leak in). Result: all 83
  `src` files reported, 0 `dist`/`test` leaks.
- **`.github/workflows/ci.yml`** — a `REES coverage` step (after the existing, authoritative
  uninstrumented `REES ... tests` step) harvests the lcov, a verify step fails CI if it is missing,
  and two upload steps (trusted + fork-tokenless, mirroring the shard uploads) send it under the
  `rees` flag. `c8` instrumentation inflates the suite's timing-sensitive linear-time/ReDoS-guard
  assertions, so the harvest tolerates the child exit — pass/fail is already gated by the
  uninstrumented step, and a broken suite would show as ~0% patch regardless.
- **`validate-tests` now also runs on `rees`-only changes** — Codecov holds `codecov/patch` until
  `after_n_builds` (6) coverage uploads land; a `rees`-only PR doesn't otherwise touch `backend`, so
  without the shards it would upload a single report and the status would never post (the same
  vacuous check being fixed). The shards guarantee the 6-upload floor; the fast `rees` upload from
  `validate-code` is already merged in before the 6th shard trips the threshold.
- **`codecov.yml`** — registers the `rees` flag (scoped to `review-enrichment/`, `carryforward`) and
  documents why `after_n_builds` stays 6.
- **`vitest.config.ts`** — drops the stale, unexercised single-file `coverage.include` entry.

## Verification

- `npm run rees:coverage` → 83 `src` files reported (was 0 tracked), paths `review-enrichment/src/...`,
  no `dist`/`test` leaks, ~91.8% lines / 86.5% branches.
- Uncovered branches are recorded as real `DA:line,0` misses in the lcov, so an untested branch in a
  `review-enrichment/src/**` change now produces a genuine `codecov/patch` shortfall instead of the
  previous no-op.
- `npm --prefix review-enrichment test` (authoritative correctness gate) → 1335 pass, 0 fail.
- `npm run actionlint`, `npm run typecheck`, `test/unit/codecov-policy.test.ts` (+ the other
  CI-policy unit tests), `npm ci --prefix review-enrichment`, and `npm audit --audit-level=moderate`
  (0 vulnerabilities) all green; `git diff --check` clean.

Closes #6250

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -364,6 +364,50 @@ jobs:
       - name: REES build, source-map validation, and tests
         if: ${{ github.event_name == 'push' || needs.changes.outputs.rees == 'true' }}
         run: npm --prefix review-enrichment test
+      # review-enrichment's node:test suite runs against the built dist/, so its real coverage was invisible
+      # to Codecov (which only ever saw vitest's src/** lcov). `npm run rees:coverage` re-runs that suite under
+      # c8 (v8 coverage) and remaps the dist hits back to review-enrichment/src/** through the build's source
+      # maps, producing an lcov Codecov can gate. The authoritative pass/fail for REES stays the uninstrumented
+      # step above: c8's instrumentation inflates that suite's timing-sensitive linear-time/ReDoS-guard
+      # assertions past their (uninstrumented) budgets, so a non-zero child exit here is expected -- this
+      # harvest only needs to emit lcov, and a genuinely broken suite is already caught above (and would show
+      # as ~0% patch anyway). The verify step below fails CI if no report lands.
+      - name: REES coverage
+        if: ${{ github.event_name == 'push' || needs.changes.outputs.rees == 'true' }}
+        run: npm run rees:coverage || true
+      - name: Verify REES coverage report exists
+        if: ${{ github.event_name == 'push' || needs.changes.outputs.rees == 'true' }}
+        run: |
+          if [ ! -s review-enrichment/coverage/lcov.info ]; then
+            echo "::error title=REES coverage::review-enrichment/coverage/lcov.info is missing or empty"
+            exit 1
+          fi
+      # Separate `rees` flag so codecov/patch actually gates review-enrichment/** changes -- before this the
+      # package was in no coverage report at all, so a REES-only PR got a vacuous patch check. Same trusted vs.
+      # fork-tokenless split and override_* handling as the shard uploads in validate-tests below.
+      - name: Upload REES coverage to Codecov
+        if: ${{ (github.event_name == 'push' || needs.changes.outputs.rees == 'true') && github.event.pull_request.head.repo.fork != true }}
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./review-enrichment/coverage/lcov.info
+          flags: rees
+          disable_search: true
+          override_branch: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref_name }}
+          override_commit: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+          override_pr: ${{ github.event_name == 'pull_request' && github.event.pull_request.number || '' }}
+          fail_ci_if_error: true
+      - name: Upload REES coverage to Codecov (fork PR tokenless)
+        if: ${{ (github.event_name == 'push' || needs.changes.outputs.rees == 'true') && github.event.pull_request.head.repo.fork == true }}
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
+        with:
+          files: ./review-enrichment/coverage/lcov.info
+          flags: rees
+          disable_search: true
+          override_branch: ${{ github.event.pull_request.head.repo.owner.login }}:${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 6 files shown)


JSONbored:
Causing way too much CI churn, please stop force pushing.

PR #6491, opened 2026-07-16
fix(test): stop 3 tests from bleeding real autonomy overrides via self-repo fallback

JSONbored:
## Summary

- Currently on \`main\`, 3 tests fail: \`test/unit/backfill.test.ts\` (x2) and \`test/integration/api.test.ts\` (x1), all in the installation-repair-diagnostics area.
- Root cause: these tests use \`"JSONbored/gittensory"\` (the old pre-rename name) as their fixture repo, set up a DB-only \`autonomy\` override, and assert on the computed required GitHub permissions -- but don't stub \`fetch\`. \`raw.githubusercontent.com\` transparently follows the gittensory->loopover repo rename, so the real, unstubbed \`fetch\` call succeeds and returns the ACTUAL live \`.loopover.yml\` content. That file gained a real \`settings.autonomy\` block in #6468 (fixing the "reviews approved but never merged" production bug), so these 3 tests started silently inheriting real production autonomy config instead of their own deterministic DB-only setup, flipping their expected \`pull_requests\`/\`contents\` permission assertions.
- Fix: stub \`fetch\` to a deterministic 404 in each (removing the live-network dependency), and override \`LOOPOVER_DRIFT_ISSUE_REPO\` so a 404'd fetch doesn't then fall back to the bundled self-repo manifest constant (which carries the same \`autonomy\` block).

## Scope
- [x] Test-only change (\`test/unit/backfill.test.ts\`, \`test/integration/api.test.ts\`, plus a similarly-affected test in \`test/unit/queue-5.test.ts\` found during the same investigation)
- [x] No production code changed

## Validation
- [x] All 3 previously-failing tests pass
- [x] Full \`test/unit\` + \`test/integration\` run is green except for one pre-existing, unrelated migration-number collision (0156, filed separately)

### DIFF
--- a/test/integration/api.test.ts
+++ b/test/integration/api.test.ts
@@ -2146,8 +2146,16 @@ describe("api routes", () => {
   });
 
   it("serves installation repair diagnostics and refreshes installation health", async () => {
+    // No repo file mocked (this test only cares about DB-configured settings): stub fetch to a plain 404 so
+    // resolveRepositorySettings's manifest lookup never makes a REAL network call to raw.githubusercontent.com,
+    // which -- being a live, mutable URL that transparently follows the gittensory->loopover rename -- would
+    // silently pick up whatever real .loopover.yml content the actual repo carries today instead of the
+    // deterministic DB-only settings this test asserts against. LOOPOVER_DRIFT_ISSUE_REPO is also overridden
+    // so a 404'd fetch doesn't fall back to the bundled self-repo manifest (LOOPOVER_REPO_FOCUS_MANIFEST_YAML),
+    // which carries its own settings.autonomy block (#773) that would equally clobber this test's DB-only setup.
+    vi.stubGlobal("fetch", async () => new Response("not found", { status: 404 }));
     const app = createApp();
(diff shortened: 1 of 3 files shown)


JSONbored:
Closing as superseded by #6493, which fixes the same 3 tests via the same technique (stub fetch + explicit self-repo override) but also traces the issue to its actual root cause: `test/helpers/d1.ts`'s `createTestEnv()` default `LOOPOVER_DRIFT_ISSUE_REPO` was `"JSONbored/gittensory"` — the same literal string ~100+ generic test fixtures use as an arbitrary placeholder — so ANY bare `createTestEnv()` call using that fixture name was silently vulnerable to the bundled self-repo manifest leaking in, not just these 3.

Running the full suite after fixing only these 3 surfaces 7 more affected files (ops-wire.test.ts, draft.test.ts, public-stats.test.ts, maintainer-recap-wire.test.ts, upstream-ruleset.test.ts, index.test.ts, queue.test.ts, routes-self-dogfood-registration-pack.test.ts, public-stats-route.test.ts, routes-errors.test.ts) that share the identical root cause and would still be broken after this PR alone. #6493 fixes the default itself plus all of these.

PR #6498, opened 2026-07-16
fix(public-stats): rename the live GitHub-stats allowlist/proxy to the current repo

JSONbored:
## Summary

- `wrangler.jsonc`'s `PUBLIC_REPO_STATS_ALLOWLIST` (gates the live `GET /v1/public/github/repos/:owner/:repo/stats` proxy, `src/github/public.ts`) and `github-stats-chip.tsx`'s own fetch/fallback URL both still targeted `JSONbored/gittensory` — worked only via GitHub's repo-rename redirect. Points both at `JSONbored/loopover` directly, and updates its test + sessionStorage cache key to match.
- `LOOPOVER_DRIFT_ISSUE_REPO` is also updated to `JSONbored/loopover` for the same reason (currently dormant — `LOOPOVER_AUTO_FILE_DRIFT_ISSUES` is off — but would otherwise file drift issues against the wrong repo the moment that flag is enabled).
- Deliberately does **not** touch `LOOPOVER_PUBLIC_STATS_REPOS`, which is a separate, correctly old-name-keyed allowlist for this repo's own HISTORICAL `audit_events` ledger data (see its own doc comment in `wrangler.jsonc`) — renaming it would point the historical stats query at a `repo_full_name` with zero matching rows.

Part of a broader gittensory→loopover residue cleanup (see sibling PRs).

**Note:** `wrangler.jsonc` vars require a `wrangler deploy` after merge to take effect in production.

## Validation
- [x] `npx wrangler deploy --dry-run` (config parses, values confirmed)
- [x] `github-stats-chip.test.tsx` green


JSONbored:
Closing as superseded — every change here (wrangler.jsonc's `LOOPOVER_DRIFT_ISSUE_REPO`/`PUBLIC_REPO_STATS_ALLOWLIST` vars, and the GitHub-stats-chip repo string) already landed on `main` via #6532. Rebasing this branch confirms it: the wrangler.jsonc commit drops as an empty patch, and the only remaining conflict is in the generated `worker-configuration.d.ts`, which regenerates identically to what's already on `main`.

PR #6808, opened 2026-07-17
fix(engine): don't throw rendering a track record for a login containing a blocklisted word (#6772)

### LINKED ISSUE #6772: Track-record summary throws on a legitimate GitHub login that collides with the safety blocklist
## Context
`packages/loopover-engine/src/track-record-summary.ts:115-126` defines `PUBLIC_FIELD_BLOCKLIST` as bare word-boundary regexes (`/\bwallet\b/iu`, `/\bhotkey\b/iu`, `/\bcoldkey\b/iu`, `/\breward\b/iu`, `/\bpayout\b/iu`, `/\branking\b/iu`). `renderTrackRecordSummaryMarkdown:424-448` unconditionally renders `GitHub login: ${markdownSafe(summary.login)}` into the same text block that `assertPublicSummaryText` scans, which throws if any blocklist pattern matches anywhere in the full text — with no exemption for the identity field. A GitHub login containing one of these words as a whole word bounded by a hyphen (a legal username character) — e.g. `my-wallet`, `payout-bot`, `coldkey-labs` — makes rendering that contributor's own summary throw, even though `PUBLIC_FIELD_BLOCKLIST` is designed to catch computed/leakable content, not to double as an identity-field validator.

## Requirements
- `renderTrackRecordSummaryMarkdown` must not throw solely because a genuine GitHub login contains a blocklisted word — exclude the login line from the blocklist scan (or scan it separately with a narrower check).
- Preserve existing fail-closed behavior for a genuinely-injected blocked term in a computed field.

## Deliverables
- [ ] Login field excluded from (or separately/narrowly scanned in) the public-safety blocklist check
- [ ] Regression test with `login: "team-wallet"` (or similar) asserting successful rendering instead of a throw
- [ ] Existing incident-label injection test still passes

## Test Coverage Requirements
99%+ Codecov patch coverage on every changed line and branch, plus a regression test that reproduces the exact failure mode described above and asserts it's fixed.

## Expected Outcome
A contributor whose real GitHub login happens to contain a blocklisted word can still have their track-record summary rendered.

## Links & Resources
`packages/loopover-engine/src/track-record-summary.ts:115-126,236-242,424-448`.

shin-core:
## Summary

- `PUBLIC_FIELD_BLOCKLIST` catches leaked **computed** fields, but `renderTrackRecordSummaryMarkdown` ran `assertPublicSummaryText` over the whole rendered block — including the `- GitHub login:` identity line.
- A hyphen is a legal GitHub username character, so a genuine login like `team-wallet`, `payout-bot`, or `coldkey-labs` matches a word-boundary pattern (`/\bwallet\b/iu`, …) and **threw while rendering that contributor's own summary**.
- Fix: compose the identity line *after* the scan and re-insert it at its original position (index 2, right after the heading's blank line). Ordinary logins render byte-identically; every computed field (merge-rate/tenure/conduct labels, evidence URLs, counts) still goes through `assertPublicSummaryText`, so a genuinely-injected blocked term still fails closed.

Closes #6772

## Test plan

- [x] New root test `test/unit/track-record-summary-login-blocklist.test.ts` (5 cases) — required for Codecov, since this module's only prior coverage was the engine's `node --test` suite, which Codecov never sees:
  - REGRESSION: `team-wallet`, `payout-bot`, `coldkey-labs`, `hotkey-ops`, `reward-guild`, `ranking-labs` all render instead of throwing.
  - The login stays on its original line (no positional drift); an ordinary login still renders lowercased + markdown-escaped exactly as before.
  - **Injection guard intact**: a blocked term in a computed field (evidence URL) still throws.
  - A disabled summary still renders empty.
- [x] `npm run test --workspace @loopover/engine` green — 588 pass / 0 fail (the existing injection test unchanged).
- [x] `npm run typecheck` clean.
- [x] Coverage: every changed line (426-455) is exercised; the file's remaining uncovered lines (348-349, 408-417) are pre-existing and outside this diff.

### DIFF
--- a/packages/loopover-engine/src/track-record-summary.ts
+++ b/packages/loopover-engine/src/track-record-summary.ts
@@ -423,10 +423,15 @@ export function shouldIncludeTrackRecordSummary(
  */
 export function renderTrackRecordSummaryMarkdown(summary: TrackRecordSummary): string {
   if (!summary.enabled) return "";
-  const lines = [
+  // The login is the contributor's own IDENTITY, not computed content, and a hyphen is a legal GitHub username
+  // character — so a real login like `team-wallet`/`payout-bot` trips the word-boundary blocklist and used to
+  // throw on that contributor's own summary (#6772). The blocklist exists to catch leaked COMPUTED fields, not
+  // to validate identity, so the login line is composed after the scan and excluded from it. Everything the
+  // blocklist is actually for (labels, evidence, counts) still goes through assertPublicSummaryText below, so a
+  // genuinely-injected blocked term in a computed field still fails closed exactly as before.
+  const scannedLines = [
     "### Public contributor record",
     "",
-    `- GitHub login: ${markdownSafe(summary.login)}`,
     `- Resolved public PRs: ${summary.outcomes.resolved} (${summary.outcomes.merged} merged, ${summary.outcomes.closedWithoutMerge} closed without merge)`,
     `- Public merge rate: ${summary.mergeRate.label}`,
     `- Public tenure: ${summary.tenure.label}`,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Please find a better implementation: 

```
The splice(2, 0, ...) approach is a bit fragile since it depends on the scannedLines array always having exactly 2 lines before the insertion point
```

Issue #6840, opened 2026-07-17
claude-cli driver's --permission-mode acceptEdits denies every Read/Bash call, silently blocking all real work


JSONbored:
## Problem

`@loopover/miner`'s `claude-cli` coding-agent driver invokes Claude Code with `--permission-mode acceptEdits` (`defaultClaudeCliArgs`, `packages/loopover-engine/src/miner/cli-subprocess-driver.ts:102-104`). In practice, `acceptEdits` only auto-approves file **edit** tool calls — it does **not** grant `Read` or `Bash` access. Every real coding task needs to read the target file(s) and (per the acceptance-criteria contract itself) run the repo's test suite before finishing, so **every** `Read`/`Bash` call gets silently denied in a headless `--print` invocation with no TTY to resolve an interactive permission prompt.

## Reproduction

Found while running AMS's full discover -> plan -> code -> PR loop end-to-end against a real repo (part of the AMS standalone-usage verification work, #6205). Two separate real `loopover-miner attempt <owner/repo> <issue#> --live` runs against the same real, small, well-scoped issue both aborted in under 1.5 seconds with zero driver turns:

```
attempt_log_events: "Self-review could not conclusively determine pass/fail: driver run did not
  complete successfully: claude_code_error_success Downgrading to abandon rather than optimistically
  handing off." (repeated identically on a second, independent retry)
```

`claude_code_error_success` comes from `claudeErrorStatus` (`cli-subprocess-driver.ts:213-218`) reading a self-contradictory `is_error: true` + `subtype: "success"` envelope back from the `claude` CLI subprocess -- itself a symptom of the underlying permission wall, not a separate bug in the error-parsing code.

To isolate the cause, I ran the same real issue's instructions directly against `claude --print --output-format json` (bypassing AMS entirely), from the exact worktree AMS had prepared:

- **`--permission-mode acceptEdits`** (AMS's actual current flag): completed after 31 turns / $0.59, but **19 of 19** tool calls (`Read` on the acceptance-criteria file, `Bash` for `npm test`/`npm run lint`, tried repeatedly with several different invocation shapes) were denied. Zero real work done -- no file was ever read, no fix was ever made, no test was ever run. `is_error` happened to read `false` in this particular run (unlike the two real AMS attempts, which hit `is_error: true`), but the outcome is identical either way: nothing gets accomplished.
[...]

PR #6885, opened 2026-07-17
fix(miner): give the coding-agent CLI subprocess HOME + its own credential so it can authenticate (#6875)

### LINKED ISSUE #6875: claude-cli/codex-cli driver's env allowlist drops HOME and every coding-agent credential
## Problem

The `claude-cli`/`codex-cli` CLI-subprocess coding-agent driver's env allowlist (`CODING_AGENT_ENV_ALLOWLIST`, `packages/loopover-engine/src/miner/cli-subprocess-driver.ts:71-85`) never passes through `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME`/`XDG_STATE_HOME`, nor any coding-agent credential (`CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `CODEX_ACCESS_TOKEN`). This means the spawned `claude`/`codex` subprocess can **never authenticate at all** — not a permission-scope problem (that was #6840, already fixed), a pure "who am I" problem: the subprocess has no `HOME` to locate a persisted credential file, and no credential env var either.

This is separate from #6840 despite surfacing through the identical-looking `claude_code_error_success` symptom — #6840's `--permission-mode`/`--allowedTools` fix was real and necessary but not sufficient; a live attempt still fails after that fix, now with an explicit `"Not logged in · Please run /login"` result once the actual cause is isolated.

## Reproduction

Confirmed on a real machine, with a real, valid `CLAUDE_CODE_OAUTH_TOKEN`, replicating the exact env `buildAllowlistedEnv` constructs for the subprocess today:

```
$ env -i PATH="$PATH" TERM="$TERM" LANG="$LANG" claude --print --output-format json \
    --permission-mode acceptEdits --allowedTools Read Bash -- 'say hello in 3 words'
{"type":"result","subtype":"success","is_error":true, ...,
 "result":"Not logged in · Please run /login", "terminal_reason":"api_error", ...}
```

Adding back exactly `HOME` and `CLAUDE_CODE_OAUTH_TOKEN` (nothing else) succeeds:

```
$ env -i PATH="$PATH" TERM="$TERM" LANG="$LANG" HOME="$HOME" CLAUDE_CODE_OAUTH_TOKEN="$CLAUDE_CODE_OAUTH_TOKEN" \
    claude --print --output-format json --permission-mode acceptEdits --allowedTools Read Bash -- 'say hello in 3 words'
{"type":"result","subtype":"success","is_error":false, "result":"Hey there, friend!", ...}
```

Also confirmed via a full, real `loopover-miner attempt <owner/repo> <issue#> --live` run with #6840's fix already installed — still failed with the identical `claude_code_error_success`/zero-turns symptom, because the env-allowlist gap is a distinct, still-live blocker.

## Root cause and existing precedent for the fix

This package already has a **correct** general-purpose allowlist that includes exactly what's missing —
`SUBPROCESS_CLI_ENV_ALLOWLIST` (`packages/loopover-engine/src/subprocess-env.ts:14-32`) includes `HOME`,
`XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_STATE_HOME` alongside the same proxy/locale/PATH vars
`CODING_AGENT_ENV_ALLOWLIST` already has. `cli-subprocess-driver.ts` defines its own separate, narrower list
instead of using this one, and neither list carries any credential env var name.

## Suggested scope for a fix (not investigated further here — this issue is the bug report, not the fix)

- Add `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME`/`XDG_STATE_HOME` to whatever allowlist `cli-subprocess-driver.ts` ends up using (reusing `SUBPROCESS_CLI_ENV_ALLOWLIST` directly would also retire a duplicate, already-drifted list).
- Explicitly forward only the credential env var that matches the **actually-configured** provider (not a blanket allowlist of every possible credential name regardless of which provider is in use — an operator with both `ANTHROPIC_API_KEY` and `OPENAI_API_KEY` set but running `claude-cli` should not have the unrelated `OPENAI_API_KEY` reach the subprocess). `packages/loopover-miner/lib/coding-agent-construction.js` already knows the resolved provider name at construction time and is a natural place to inject this as the `extra`/`options.env` override `buildAllowlistedEnv` already supports.
- Add a regression test asserting the real, spawned argv's env actually carries `HOME` and the resolved credential — the existing driver tests, per #6840's own bug report, are argv-shape assertions and didn't catch this gap either.

## Impact

This is the primary reason no `claude-cli`/`codex-cli`-driven AMS attempt has ever completed real work in production — more fundamental than #6840's permission-mode gap, which was necessary but not sufficient. Both together are required for a live attempt to succeed.

real-venus:
## Summary

The `claude-cli`/`codex-cli` coding-agent subprocess driver's env allowlist (`CODING_AGENT_ENV_ALLOWLIST`,
`packages/loopover-engine/src/miner/cli-subprocess-driver.ts`) forwarded **no `HOME`/XDG dir and no credential**
to the spawned `claude`/`codex` process. So the subprocess had **nowhere to locate a persisted credential file**
*and* **no credential env var** — every live attempt failed with `"Not logged in · Please run /login"`. This is
the actual cause behind the `claude_code_error_success` symptom that #6840's `--permission-mode` fix was
necessary-but-not-sufficient for (confirmed in the issue against a real machine + a real `--live` run).

Two parts:

1. **HOME/XDG passthrough** — add `HOME`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `XDG_STATE_HOME` to
   `CODING_AGENT_ENV_ALLOWLIST`. These are non-secret *location pointers* so the CLI can find its own on-disk
   credential/config; the credential itself is **not** put in this static list.
2. **Provider-specific credential forwarding** — in `createCliProvider` (`driver-factory.ts`), forward **only the
   selected provider's own** credential env vars as the driver's `env` override (which `buildAllowlistedEnv`
   already merges on top of the allowlisted parent env):
   - `claude` → `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_API_KEY`
   - `codex` → `OPENAI_API_KEY`, `CODEX_ACCESS_TOKEN`

   Deliberately provider-specific, **not** a blanket allowlist of every credential name: an operator with both
   providers' keys set who is running `claude-cli` must not have the unrelated `OPENAI_API_KEY` leak into the
   claude subprocess (per the issue's explicit "decide and document" ask).

Closes #6875

## Scope

- [x] I linked a currently open issue this PR resolves (`Closes #6875`).

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck`
- [x] `npm run test:coverage` locally — the new `providerCredentialEnv` (both the credential-present and unset/empty guard branches, for both `claude` and `codex`) and the injection site are fully covered, verified via `coverage-final.json`. The `HOME`/XDG allowlist additions are exercised by the passthrough assertion.
- [x] engine `build` + `node --test` (588 passing) + `engine-parity:drift-check`
- [x] `npx vitest run test/unit/coding-agent-miner.test.ts` (62 passing) — new regressions assert the **spawned subprocess env actually carries** `HOME`/`XDG_CONFIG_HOME` + the resolved provider credential, that the *other* provider's key is NOT forwarded, and that an unset/empty credential forwards nothing.

- Two engine miner-driver files + their unit test; no UI/OpenAPI/MCP/migration surface, so those checks are N/A to this diff.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, user PATs, private keys, raw trust scores, private rankings, or private maintainer evidence are exposed. — credential **values** are read from `env` at runtime; nothing is hardcoded, and test fixtures use obviously-fake placeholders.
- [x] Auth, cookie, CORS, GitHub App, Cloudflare, or session changes include negative-path tests. — the credential-forwarding negative paths (wrong provider's key NOT forwarded; unset/empty NOT forwarded) are explicitly tested, so a credential can never reach the wrong subprocess.
- [x] API/OpenAPI/MCP behavior is updated and tested where needed. — N/A: no API/MCP surface changed.
- [x] No visible UI changes (miner subprocess env only), so no `UI Evidence` section is required.
- [x] Public docs/changelogs: none needed.

## Notes

- The credential env is merged by the driver's existing `buildAllowlistedEnv(parentEnv, allowlist, extra)` call — HOME/XDG come from the allowlisted parent env, the provider credential comes from the `extra` override — so a deployment with no credential configured is byte-identical to today (it still fails at "not logged in", just now with the actual cause fixed once a credential is set).

### DIFF
--- a/packages/loopover-engine/src/miner/cli-subprocess-driver.ts
+++ b/packages/loopover-engine/src/miner/cli-subprocess-driver.ts
@@ -69,6 +69,10 @@ const MAX_TRANSCRIPT_CHARS = 8000;
 const MAX_ERROR_DETAIL_CHARS = 500;
 
 const CODING_AGENT_ENV_ALLOWLIST = [
+  // #6875: HOME + the XDG base-dir vars let the spawned `claude`/`codex` CLI locate its own persisted
+  // credential/config file. Without HOME the subprocess is "Not logged in" even with a valid token on disk. These
+  // are non-secret location pointers (the credential itself is forwarded provider-specifically, NOT via this list).
+  "HOME",
   "HTTPS_PROXY",
   "HTTP_PROXY",
   "LANG",
@@ -79,6 +83,9 @@ const CODING_AGENT_ENV_ALLOWLIST = [
   "SSL_CERT_DIR",
   "SSL_CERT_FILE",
   "TERM",
+  "XDG_CONFIG_HOME",
+  "XDG_DATA_HOME",
+  "XDG_STATE_HOME",
   "https_proxy",
   "http_proxy",
   "no_proxy",
--- a/packages/loopover-engine/src/miner/driver-factory.ts
+++ b/packages/loopover-engine/src/miner/driver-factory.ts
@@ -144,6 +144,28 @@ function buildCliArgsWithConfiguredModel(
   };
 }
 
+/** The credential env vars each CLI provider authenticates with (#6875). ONLY the selected provider's credentials
+ *  are forwarded to its subprocess: an operator with keys for both providers who is running `claude-cli` must not
+ *  have the unrelated `OPENAI_API_KEY` reach the claude subprocess, so this is deliberately provider-specific
+ *  rather than a blanket allowlist of every credential name. (The subprocess's HOME/XDG dirs — where a persisted
+ *  credential file lives — are handled separately by CODING_AGENT_ENV_ALLOWLIST in cli-subprocess-driver.ts.) */
+const CLI_PROVIDER_CREDENTIAL_ENV_KEYS: Record<"claude" | "codex", readonly string[]> = {
+  claude: ["CLAUDE_CODE_OAUTH_TOKEN", "ANTHROPIC_API_KEY"],
+  codex: ["OPENAI_API_KEY", "CODEX_ACCESS_TOKEN"],
+};
+
+/** The selected provider's credential env vars that are actually set, as an `env` override the driver merges on
+ *  top of the allowlisted parent env. Empty when none are configured — the caller then fails at "not logged in"
+ *  the same as before, but a configured credential now reaches the subprocess. (#6875) */
+function providerCredentialEnv(command: "claude" | "codex", env: Record<string, string | undefined>): Record<string, string> {
+  const credentials: Record<string, string> = {};
+  for (const key of CLI_PROVIDER_CREDENTIAL_ENV_KEYS[command]) {
+    const value = env[key];
+    if (typeof value === "string" && value.length > 0) credentials[key] = value;
+  }
+  return credentials;
+}
+
 function createCliProvider(
   command: "claude" | "codex",
   modelEnvKey: string,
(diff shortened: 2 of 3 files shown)


JSONbored:
Duplicate of another PR, closing.

PR #6978, opened 2026-07-17
fix(ui): SSR-safe homepage install-tab via useLocalStorage

### LINKED ISSUE #6814: Homepage install-tab selection reads localStorage synchronously, risking an SSR hydration mismatch
## Context
`apps/loopover-ui/src/routes/index.tsx:514-517` initializes `useState` by reading `window.localStorage.getItem("gt:install-tab")` directly. This app is server-rendered (`@tanstack/react-start`). The server render always produces `"miners"` (no `window`); the client's first hydration render immediately reads `localStorage`, which can differ for a returning visitor — a client/server markup mismatch on first paint. The codebase has an established, correct fix already used elsewhere: `apps/loopover-ui/src/lib/use-local-storage.ts` ("SSR-safe localStorage hook, reads once on mount via useEffect, exposes a hydrated flag"), correctly adopted in `app.index.tsx`, `app.runs.tsx`, `app.analytics.tsx`, `app.workbench.tsx` — but not here.

## Requirements
- Replace the inline `useState` initializer with `useLocalStorage("gt:install-tab", "miners")` (or equivalent useEffect-deferred read), gating the active-tab render on its `hydrated` flag the way `app.workbench.tsx` does.
- Add/adjust a test asserting the initial render is `"miners"` regardless of a pre-existing `localStorage` value, then updates post-mount.

## Deliverables
- [ ] Homepage install-tab state migrated to `useLocalStorage`
- [ ] Regression test for the SSR-safe initial render

## Test Coverage Requirements
99%+ Codecov patch coverage on every changed line and branch, plus a regression test that reproduces the exact gap described above (note: apps/** UI code may fall outside the strict Codecov patch gate — check coverage.include before assuming it's enforced here, and add a before/after screenshot table per this repo's UI-PR convention regardless).

## Expected Outcome
No client/server hydration mismatch on the homepage for a returning visitor with a saved tab preference.

## Links & Resources
`apps/loopover-ui/src/routes/index.tsx:514-517`, `apps/loopover-ui/src/lib/use-local-storage.ts`.

RealDiligent:
## Summary
- Migrate homepage `ClientSetupTabs` install-tab state to `useLocalStorage`, gating the active tab on `hydrated` so the first paint always matches the SSR default (`miners`).
- Migrate pre-#6814 bare tab ids in `localStorage` to JSON so returning visitors keep their preference.
- Add regression tests for initial SSR-safe render, bare-id migration, and JSON persistence.

Closes #6814.

## Test plan
- [ ] UI unit tests green (`index.test.tsx`)
- [ ] `validate-code` / prettier green
- [ ] Manual: with `gt:install-tab` set, first paint is Miner CLI then hydrates to the stored tab

### DIFF
--- a/apps/loopover-ui/src/routes/index.test.tsx
+++ b/apps/loopover-ui/src/routes/index.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, waitFor, act } from "@testing-library/react";
 import { beforeEach, describe, expect, it } from "vitest";
 
 import { ClientSetupTabs } from "./index";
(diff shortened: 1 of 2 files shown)


RealDiligent:
Closing — #6814 was already fixed on main (overlapping SSR-safe `useLocalStorage` migration). Picking up a different help-wanted next.

Issue #6797, opened 2026-07-17
AMS contribution-profile: local cache store + doctor integration


JSONbored:
Part of #6793. Depends on #6795 (schema) and #6796 (extraction logic this store persists the output of).

## Context

Extraction (#6796) shouldn't run on every `discover` invocation — label taxonomy and docs are cheap to fetch but not free, and a profile is stable enough to cache with a refresh policy, matching the pattern already established by the miner's existing local stores.

## Requirements

- A new local SQLite store (e.g. `contribution-profile-cache.sqlite3` under the standard miner config dir, following the same `LOOPOVER_MINER_CONFIG_DIR`/`LOOPOVER_MINER_<NAME>_DB` override convention every other miner store already uses — see `packages/loopover-miner/DEPLOYMENT.md`'s store table) keyed by `owner/repo`, storing the profile plus a fetch/refresh timestamp.
- A sensible default TTL (informed by how often labels/docs realistically change — err toward the existing `policy-doc-cache.sqlite3`'s ETag-based approach if that pattern fits better than a flat TTL).
- `doctor`'s `store-integrity:*` checks extended to cover this new store, matching every other store's existing coverage.

## Deliverables

- The cache store implementation and its `doctor` integration.
- Regression tests for the cache read/write/expiry behavior.

## Expected Outcome

A repeated `discover` run against the same repo within the cache's freshness window doesn't re-fetch/re-parse the same signals every time.

Issue #6793, opened 2026-07-17
epic: AMS per-repo contribution-profile discovery (labels, docs, agent-facing eligibility signals)


JSONbored:
## Problem

`loopover-miner discover <owner/repo>` ranks and enqueues candidate issues purely on its own generic scoring model — it has no awareness of a target repo's actual PR-eligibility rules. On JSONbored's own gate-enabled repos (loopover, metagraphed), a contributor PR must link-close an existing open issue that carries a points-bearing `gittensor:*` label, is not labeled `maintainer-only`, and is not assigned to the repo owner — an issue that fails any of those conditions still ranks and gets enqueued by `discover` today, and a PR opened against it would get auto-closed by the gate on arrival.

Hardcoding loopover's own `gittensor:*`/`maintainer-only` convention into AMS would only work for JSONbored's own repos — any other repo AMS targets will have a different (or no) eligibility convention, expressed however that maintainer chose to document it (a CONTRIBUTING.md section, a label description, a PR template checklist, an AGENTS.md/.claude skill file, or nothing explicit at all).

## Proposal

AMS should build a generic, per-repo **contribution profile** by reading whatever signals a target repo actually exposes — label taxonomy (via the GitHub API, including label descriptions), `CONTRIBUTING.md`, `.github/*` templates, and any AI-agent-facing docs (`AGENTS.md`, `.claude/skills/**`, similar) — and use that profile to filter/deprioritize ineligible candidates in `discover`, instead of assuming every repo works like loopover's own.

Found during live end-to-end verification of AMS's standalone usage (#6205): running `discover` against a real repo with a mix of contributor-eligible and maintainer-only issues surfaced this gap directly — nothing in AMS today distinguishes the two.

## Deliverables

Tracked via the sub-issues below. Together they should produce:
- A defined `ContributionProfile` shape (what a profile records, how confident it is, how it's cached).
- A concrete extraction path from real repo signals (labels, docs, agent-facing files) to a profile.
- `discover` actually consulting the profile to exclude/deprioritize ineligible candidates.
- Documented, deliberately conservative behavior when no profile can be inferred at all.

## Boundaries

- This epic doesn't require any change to how loopover's own gate enforces eligibility server-side — it's about AMS's own client-side candidate selection getting smarter before it ever opens a PR, so a contributor (human or AMS) doesn't waste effort on a PR the gate would reject anyway.
- Not scoped to build a UI for reviewing/editing a profile — CLI/library-level only for this epic.

[...]

PR #7034, opened 2026-07-17
feat(miner): AMS contribution-eligibility filtering for discover

### LINKED ISSUE #6798: AMS contribution-profile: wire eligibility filtering into discover
Part of #6793. Depends on #6796 (extraction) and #6797 (cache) — this issue wires the finished profile into actual candidate selection.

## Context

This is the issue that closes the loop: today `discover` ranks and enqueues candidates with zero eligibility awareness, exactly the gap that motivated this whole epic (#6793).

## Requirements

- `loopover-miner discover <owner/repo>` fetches (or reads the cached) `ContributionProfile` for the target repo and uses it to exclude or deprioritize candidates that fail the profile's eligibility rules (wrong/missing label, exclusion label present, excluded assignee) before ranking/enqueueing.
- When the profile's confidence is low or empty (no discoverable signals at all — see #6796's requirement to degrade honestly rather than guess), `discover` must default to NOT filtering anything out — a low-confidence profile should never cause AMS to silently skip real, eligible work on a repo whose conventions it simply couldn't read. Filtering only activates when the profile has high enough confidence to trust.
- Surface the applied profile (and any excluded candidates + why) in `discover`'s output, so a human running it can see what AMS inferred and why a candidate was skipped.

## Test Coverage Requirements

Regression tests for: a repo whose profile correctly excludes maintainer-only-equivalent issues, a repo with a low-confidence/empty profile where nothing gets filtered, and a repo where the profile's exclusion rules and inclusion rules both match the same issue (conflicting signals) — document and test whatever conservative resolution is chosen.

## Deliverables

- The `discover` integration and its test coverage.
- A short doc update (`packages/loopover-miner/README.md` or `docs/`) explaining the new eligibility-aware behavior and its safe-default posture.

## Expected Outcome

`loopover-miner discover` no longer surfaces candidates that a target repo's own gate/maintainer would reject, for any repo whose conventions AMS can confidently read — closing the exact gap found during #6205's live verification.

JSONbored:
## Summary
- Closes #6798, the final step of the AMS contribution-profile epic (#6793). #6797 (the cache store this depends on) already merged independently via #7001 — this PR builds on that.
- Wires eligibility filtering into `discover`: resolves each target repo's `ContributionProfile` (cache hit when fresh, extracted live otherwise) and excludes a candidate missing a required eligibility label, carrying an exclusion label, or assigned to the repo's own owner, before ranking/enqueueing.
- Extends `RawCandidateIssue` with `assignees` — already present in the same GitHub list/search payload `discover` already fetches, no extra request.
- New `contribution-profile-eligibility.js` (matcher evaluation + conditional per-repo label-description fetch for the "rust E-easy" description-field case) and `contribution-profile-resolution.js` (cache-or-extract-live per repo).

## Design notes
- **Safe-default posture**: a label rule only excludes when the profile actually resolved a matcher for it. A repo whose conventions couldn't be read (`value: null`) excludes nothing via that rule — `discover` never silently skips real work on an unreadable repo.
- **Conflicting signals**: a candidate that is both eligible and excluded resolves conservatively — exclusion always wins.
- Both `--dry-run` and the real run apply the same filtering; `--dry-run` never opens the cache, matching the existing policy-doc/policy-verdict cache convention.
- Excluded candidates and their reasons are surfaced in both `--json` and the human-readable summary (`excludedByEligibility`).

## Test plan
- [x] `npm run typecheck` — clean.
- [x] `npm run test:coverage` (full, unsharded) — 978 files / 18314 tests passed, 0 failures.
- [x] New unit tests for the eligibility matcher (29 tests), profile resolution (8 tests), and discover-cli integration (4 new tests covering real exclusion, dry-run, default cache open/close, and corrupt-cache degradation).


JSONbored:
Closing as superseded — #7024 independently implemented and merged this exact issue (#6798) while this PR was in progress, along with #6797 (#7001) before it. Both sub-issues are now closed.

One real gap remains: #7024 candidly scoped out assignee-exclusion as a follow-up ("a genuine second shared-file change... rather than half-wiring it"). I'll file a narrow, separate issue/PR for just that piece on top of the now-merged contribution-profile-filter.js, rather than carrying this duplicate diff forward.

Issue #6839, opened 2026-07-17
Miner dashboard chat: governor action-dispatch


JSONbored:
## Context
Part of epic #6504's chat track, blocked on the action-dispatch scaffolding child issue of this same epic. Governor pause/resume already has a real API route (`/api/governor/{pause,resume}`, wired from `ledgers.tsx`'s existing governor-control section) — this issue wires chat to call that same, already-existing route.

## Requirements
- Wire chat's action-dispatch layer (from the scaffolding child issue) to call the existing `/api/governor/{pause,resume}` routes — never a new or parallel route.
- No changes to the existing route or button-triggered flow.

## Deliverables
- [ ] Chat wiring to the existing governor pause/resume routes
- [ ] Test coverage

## Test Coverage Requirements
99%+ Codecov patch coverage on every changed line and branch, plus a regression test for the new/changed behavior (note: apps/** UI code may fall outside the strict Codecov patch gate — check coverage.include — but a before/after screenshot table per this repo's UI-PR convention is still required for any visual change).

## Expected Outcome
Chat can trigger governor pause/resume through the exact same governed endpoint the dashboard buttons already use.

## Links & Resources
Epic: #6504. Blocked by the "action-dispatch scaffolding" issue in this same epic. `apps/loopover-miner-ui/src/routes/ledgers.tsx`.

JSONbored:
Already done — via #6521. `apps/loopover-miner-ui/src/lib/chat-governor-actions.ts` already registers governor pause/resume chat actions bound to the existing `pauseGovernor`/`resumeGovernor` clients — the same functions the Ledgers buttons call, no new route added (`vite-governor-api.ts` is unmodified). Closing as already resolved.

PR #7041, opened 2026-07-17
fix(miner): fail closed when a chat-action handler throws

### LINKED ISSUE #6989: chat-action-dispatch.js doesn't catch handler exceptions (unlike its own paramsValidator call)
## Context
`packages/loopover-miner/lib/chat-action-dispatch.js`'s `dispatchChatAction` wraps its `paramsValidator` call in a try/catch, explicitly commented "fail closed" (a validator that throws is treated as a rejection, not a dispatch error) — but the very next line, `await registered.handler(request)`, is not wrapped at all. Every other "must never crash" write path in this package wraps its fallible call the same way `paramsValidator` does here: `pretooluse-hook.js:32`, `sentry.js:37`/`47`, and `iterate-loop.ts`'s `safeAppendAttemptLogEvent`.

Today this is latent — the dashboard-chat REST endpoint that will actually invoke `dispatchChatAction` end-to-end is still an open epic (#6839, governor action-dispatch, and its siblings). Once wired, a handler throwing (e.g. a network error from `pauseGovernor`/`resumeGovernor`, or any other registered action's network call) becomes an unhandled rejection instead of the module's own typed `{ ok, status }` contract every other failure path already returns.

## Requirements
- Wrap `await registered.handler(request)` in a try/catch, returning the module's existing typed failure shape (`{ ok: false, status: ..., action, error: ... }`) on a thrown error, consistent with the `paramsValidator` catch immediately above it.
- Choose a `status` value distinct from `"invalid_params"`/`"unknown_action"` (e.g. `"handler_error"`) so a caller can tell a validation failure apart from a handler-execution failure.
- Do not change `paramsValidator`'s existing catch behavior or the registry lookup logic — this issue is scoped to the one unguarded `handler(request)` call.

## Deliverables
- [ ] `registered.handler(request)` is wrapped in try/catch with a typed failure result on throw
- [ ] Test asserting a throwing handler produces `{ ok: false, ... }` instead of an unhandled rejection

## Test Coverage Requirements
99%+ Codecov patch coverage on every changed line and branch, plus the regression test above.

## Expected Outcome
A registered chat action's handler throwing (e.g. a network failure) fails closed with the module's own typed result shape, matching every other guarded call in this dispatch path, instead of surfacing as an unhandled rejection once the dashboard-chat epic wires this module to a live REST endpoint.

## Links & Resources
`packages/loopover-miner/lib/chat-action-dispatch.js` — the call site. `packages/loopover-miner/lib/pretooluse-hook.js:32`, `sentry.js:37,47` — sibling fail-closed catch precedent.

shin-core:
## What

`dispatchChatAction` (`packages/loopover-miner/lib/chat-action-dispatch.js`) wraps its `paramsValidator` call in a try/catch — explicitly "fail closed" — but the next line, `await registered.handler(request)`, was unguarded. A handler that throws (e.g. a network error from a governor action) would surface as an unhandled rejection instead of the module's own typed `{ ok, status }` contract that every other failure path already returns.

This is latent today (the dashboard-chat REST endpoint that invokes it is still an open epic, #6839), but wiring it live would turn any handler throw into an unhandled rejection.

## How

Wrap `registered.handler(request)` in try/catch, returning `{ ok: false, status: "handler_error", action, error }` on throw — mirroring the `paramsValidator` catch immediately above. The distinct `"handler_error"` status (vs `"invalid_params"`) lets a caller tell an execution failure apart from a validation failure. This matches the fail-closed catch pattern used by the package's other must-never-crash paths (`pretooluse-hook.js:32`, `sentry.js:37,47`).

## Validation

Tests added to the existing `test/unit/miner-chat-action-dispatch.test.ts`: a handler throwing an `Error` yields `handler_error` with the message, and a non-`Error` throw is stringified — covering both sides of the `instanceof Error` ternary. Confirmed both **fail against the unfixed code** (unhandled rejection) and pass with the fix; 100% branch coverage on the file.

Closes #6989

### DIFF
--- a/packages/loopover-miner/lib/chat-action-dispatch.js
+++ b/packages/loopover-miner/lib/chat-action-dispatch.js
@@ -71,7 +71,15 @@ export async function dispatchChatAction(request, options = {}) {
     return { ok: false, status: "invalid_params", action };
   }
 
-  const result = await registered.handler(request);
+  let result;
+  try {
+    result = await registered.handler(request);
+  } catch (error) {
+    // A handler that throws fails closed with the module's typed result shape (#6989), consistent with the
+    // paramsValidator catch above -- a distinct "handler_error" status lets a caller tell an execution
+    // failure apart from a params-validation failure.
+    return { ok: false, status: "handler_error", action, error: error instanceof Error ? error.message : String(error) };
+  }
   return { ok: true, status: "dispatched", action, result };
 }
 
--- a/test/unit/miner-chat-action-dispatch.test.ts
+++ b/test/unit/miner-chat-action-dispatch.test.ts
@@ -147,4 +147,21 @@ describe("dispatchChatAction (#6519)", () => {
     const result = await dispatchChatAction({ action: "demo", params: {} }, { env: enabledEnv, registry });
     expect(result).toEqual({ ok: false, status: "invalid_params", action: "demo", error: "boom" });
   });
+
+  it("fails closed with handler_error when the handler throws, not an unhandled rejection (#6989)", async () => {
+    const registry = registryWith("demo", () => true, () => {
+      throw new Error("network down");
+    });
+    const result = await dispatchChatAction({ action: "demo", params: {} }, { env: enabledEnv, registry });
+    // Distinct from invalid_params so a caller can tell an execution failure from a validation failure.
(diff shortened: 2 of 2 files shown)


JSONbored (commented):
Outstanding superagent issue, closing.

PR #7068, opened 2026-07-17
fix(review-enrichment): lockfile-drift OSV batch failure falls back to per-item queries

### LINKED ISSUE #7009: lockfile-drift.ts's queryOsvBatch silently drops a chunk on batch-API failure, unlike its dependency-scan.ts sibling
## Context
`review-enrichment/src/analyzers/lockfile-drift.ts`'s `queryOsvBatch` (line 388) silently drops an entire chunk of CVE findings when the OSV batch API call fails: `if (!response.ok) continue;` — no fallback, that chunk's changes simply never get findings.

Its near-identical sibling, `review-enrichment/src/analyzers/dependency-scan.ts`'s own `queryOsvBatch` (line 328), handles the exact same failure differently: on `!response.ok`, it falls back to per-item `fetchOsvDirect` calls for every change in that chunk, so a batch-API hiccup degrades to slower-but-complete results instead of silently losing data.

Both functions have the same signature shape, the same chunking strategy, and query the same `api.osv.dev/v1/querybatch` endpoint — `lockfile-drift.ts`'s version never got the per-item fallback its sibling already has.

## Requirements
- Add the same per-item fallback to `lockfile-drift.ts`'s `queryOsvBatch`: on `!response.ok`, iterate the failed chunk's changes and fetch each individually, following `dependency-scan.ts`'s exact pattern.
- `lockfile-drift.ts` may need its own per-item direct-query helper (mirroring `dependency-scan.ts`'s `fetchOsvDirect`) if one doesn't already exist for lockfile changes — check for an existing equivalent before writing a new one.
- Preserve existing chunking, `AbortSignal`, and rate-limit (`maxCallsPerCategory`) behavior — this issue is scoped to the failure-fallback path only.

## Deliverables
- [ ] `lockfile-drift.ts`'s `queryOsvBatch` falls back to per-item queries on batch failure, matching `dependency-scan.ts`
- [ ] Test asserting a simulated batch-API failure still produces per-item CVE results instead of an empty chunk

## Test Coverage Requirements
99%+ Codecov patch coverage on every changed line and branch, plus the regression test above.

## Expected Outcome
A transient OSV batch-API failure during lockfile-drift analysis degrades to slower per-item queries instead of silently losing an entire chunk's CVE findings — matching the resilience `dependency-scan.ts` already has for the identical failure mode.

## Links & Resources
`review-enrichment/src/analyzers/dependency-scan.ts:328-395` (`queryOsvBatch`'s existing fallback — the precedent to mirror, including `fetchOsvDirect`). `review-enrichment/src/analyzers/lockfile-drift.ts:388` — the function to fix.

lourincedaging0-commits:
Closes #7009

## What

`review-enrichment/src/analyzers/lockfile-drift.ts`'s `queryOsvBatch` did
`if (!response.ok) continue;` when the OSV `/v1/querybatch` call failed — silently
dropping every CVE finding for that entire chunk. Its near-identical sibling
`dependency-scan.ts` handles the same failure by falling back to per-item `/v1/query`
calls, so a transient batch-API hiccup degrades to slower-but-complete results instead
of lost data.

This adds the same resilience to `lockfile-drift.ts`:

- On a non-ok batch response, iterate the failed chunk and query each change
  individually via a new `fetchOsvDirect` helper (mirroring `dependency-scan.ts`'s,
  using the lockfile `toCves` mapping and the OSV single-query endpoint).
- Existing chunking, `AbortSignal`, and per-category rate-limit (`maxCallsPerCategory`)
  behavior are preserved — the change is scoped to the failure-fallback path only.

## Tests

- A successful batch still maps each change's vulns and never issues a per-item call.
- A failed batch falls back to per-item queries, keeping the vulnerable package's CVE
  and recording the clean package as empty (not silently dropped).
- A per-item fallback that also fails records an empty result rather than omitting the
  change.
- The fallback routes through an injected `AnalysisContext` when one is supplied.
- A mid-batch abort short-circuits the per-item fallback to empty results.
- End-to-end: `scanLockfileDrift` still surfaces a per-item CVE finding for a
  vulnerable lockfile bump when the batch endpoint fails.

### DIFF
--- a/review-enrichment/src/analyzers/lockfile-drift.ts
+++ b/review-enrichment/src/analyzers/lockfile-drift.ts
@@ -384,6 +384,44 @@ export function extractLockfileChanges(
   return changes;
 }
 
+/** Per-item OSV.dev query — the fallback the batch path degrades to when `/v1/querybatch` fails, mirroring
+ *  dependency-scan.ts's fetchOsvDirect so a transient batch-API failure yields slower-but-complete results
+ *  instead of silently dropping the whole chunk's findings. */
+async function fetchOsvDirect(
+  ecosystem: string,
+  name: string,
+  version: string,
+  fetchImpl: typeof fetch,
+  signal?: AbortSignal,
+  options: Pick<ScanOptions, "analysis" | "diagnostics" | "limits"> = {},
+): Promise<Cve[]> {
+  if (signal?.aborted) return [];
+  const fetchOptions = {
+    endpointCategory: "osv-query",
+    method: "POST",
+    headers: { "content-type": "application/json" },
+    body: JSON.stringify({ package: { name, ecosystem }, version }),
+    signal,
+    fetchImpl,
+    diagnostics: options.diagnostics,
+    phase: "lockfile-drift",
+    subcall: "osv-query",
+    maxBytes: 1024 * 1024,
+    maxCallsPerCategory: options.limits?.maxOsvQueries ?? MAX_OSV_QUERIES,
+  };
+  const response = options.analysis
+    ? await options.analysis.fetchJson<{ vulns?: OsvVuln[] }>(
+        "https://api.osv.dev/v1/query",
+        fetchOptions,
+      )
+    : await boundedFetchJson<{ vulns?: OsvVuln[] }>(
+        "https://api.osv.dev/v1/query",
+        fetchOptions,
+      );
+  if (!response.ok) return [];
+  return toCves(response.data.vulns);
+}
+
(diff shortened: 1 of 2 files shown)


lourincedaging0-commits:
Withdrawing — the per-item OSV fallback for lockfile-drift's batch queries already landed on main via 75a6e0a0 (fix(rees): add per-item OSV fallback to lockfile-drift batch queries), which implements the same fetchOsvDirect fallback this PR proposed. No longer needed. Thanks!

PR #7260, opened 2026-07-19
fix(engine): block the RFC 6598 CGNAT range in safe-url's SSRF guard

### LINKED ISSUE #7253: safe-url's SSRF IPv4 guard is missing the RFC 6598 CGNAT range (100.64.0.0/10)
## Context
`packages/loopover-engine/src/review/safe-url.ts` is a self-contained SSRF-safe URL guard, ported from reviewbot's `core/source-url.ts`. `ipv4IsPrivateOrLocal` (line 45-52) rejects a set of private/reserved IPv4 ranges by decoding the address to an integer and checking each documented block:

```ts
function ipv4IsPrivateOrLocal(host: string): boolean {
  const n = ipv4ToInt(host);
  if (n === null) return false;
  const a = (n >>> 24) & 0xff;
  const b = (n >>> 16) & 0xff;
  if (a === 0 || a === 10 || a === 127) return true; // 0.0.0.0/8, 10/8, loopback
  if (a === 169 && b === 254) return true; // link-local (incl. cloud metadata 169.254.169.254)
  if (a === 192 && b === 168) return true;
  if (a === 172 && b >= 16 && b <= 31) return true;
  return false;
}
```

This covers 0.0.0.0/8, 10.0.0.0/8, 127.0.0.0/8 (loopback), 169.254.0.0/16 (link-local, explicitly called out for covering the cloud-metadata endpoint), 192.168.0.0/16, and 172.16.0.0/12. It does not check 100.64.0.0/10 (RFC 6598, "Shared Address Space" / Carrier-Grade NAT) — a range reserved specifically for service-provider address sharing, which is exactly the kind of internal, non-publicly-routable range this guard exists to block, in the same category as the five ranges already checked. Real cloud/hosting environments assign 100.64.0.0/10 addresses to internal service endpoints (it's used by several major cloud providers for exactly this purpose), so a URL resolving to a `100.64.x.x`–`100.127.x.x` host currently passes this guard's IPv4 check as if it were a normal public address.

This module is explicitly a "SELF-CONTAINED NATIVE PORT" with "no imports, no I/O" (file header) — the fix is a pure, additive range check following the identical pattern already used for every other private range in the same function; it does not touch the guard's overall design, its call sites, or any other part of the SSRF protection surface (hostname checks, IPv6 handling, encoded-IP decoding are all unaffected).

## Requirements
- Add a check for 100.64.0.0/10 (`a === 100 && b >= 64 && b <= 127`) to `ipv4IsPrivateOrLocal`, following the exact same inline style and range-comment convention already used for the five existing checks in this function.
- Do not change any other range check, the function's signature, or any other function in this file.
- Since this is a security-relevant guard (SSRF protection), the PR must not touch anything beyond this one additive range check — no refactoring of the surrounding decode/parse logic in the same change.

## Deliverables
- [ ] `ipv4IsPrivateOrLocal` in `packages/loopover-engine/src/review/safe-url.ts` rejects 100.64.0.0/10 (RFC 6598 CGNAT/Shared Address Space) the same way it already rejects the other five private ranges
- [ ] Regression tests covering the new range's boundaries: `100.63.255.255` (just below the range, must NOT be flagged private), `100.64.0.0` and `100.127.255.255` (the range's inclusive bounds, must be flagged private), `100.128.0.0` (just above the range, must NOT be flagged private) — matching the boundary-testing style already used for the file's other range checks in its existing test suite

## Test Coverage Requirements
This repo's Codecov patch gate is 99%+ hard (branch-counted) on every changed line/branch in `src/**`/`packages/**`. The four boundary-value regression tests above must all be present and passing; this is a security-relevant module so boundary precision matters — an off-by-one here would either fail to close the gap or wrongly reject legitimate public addresses adjacent to the range.

## Expected Outcome
A URL resolving to a 100.64.0.0/10 host is now correctly rejected by `ipv4IsPrivateOrLocal`, closing an SSRF gap in the same guard that already protects against the other standard private/reserved IPv4 ranges.

## Links & Resources
`packages/loopover-engine/src/review/safe-url.ts:45-52` (`ipv4IsPrivateOrLocal`, the function to extend). RFC 6598 defines 100.64.0.0/10 as Shared Address Space / Carrier-Grade NAT range.

shin-core:
## What

`safe-url.ts`'s `ipv4IsPrivateOrLocal` — the SSRF-safe URL guard's IPv4 check — rejected five private/reserved ranges (0/8, 10/8, 127/8, 169.254/16, 192.168/16, 172.16/12) but **not `100.64.0.0/10`** (RFC 6598 "Shared Address Space" / carrier-grade NAT). That range is non-publicly-routable and is assigned by several major cloud/hosting providers to internal service endpoints — exactly the category this guard exists to block. A URL resolving to a `100.64.x.x`–`100.127.x.x` host currently passes the IPv4 check as if it were a normal public address.

## How

Add one additive check — `if (a === 100 && b >= 64 && b <= 127) return true;` — following the identical inline style and range-comment convention as the five existing checks. No other range, the function signature, or any other part of the SSRF surface (hostname checks, IPv6, encoded-IP decoding) is touched, per the issue's security-scoping requirement.

## Validation

Regression tests added to `safe-url-engine.test.ts` covering the range boundaries: `100.64.0.0` and `100.127.255.255` (inclusive bounds — must be blocked), `100.100.50.1` (mid-range — blocked), and `100.63.255.255` / `100.128.0.0` (just outside — must remain public/safe), matching the file's existing boundary-testing style. Confirmed the tests **fail against the unfixed guard** (the CGNAT addresses pass through) and pass with the fix; the full `safe-url-engine` suite (21 tests) passes.

Closes #7253

### DIFF
--- a/packages/loopover-engine/src/review/safe-url.ts
+++ b/packages/loopover-engine/src/review/safe-url.ts
@@ -51,6 +51,7 @@ function ipv4IsPrivateOrLocal(host: string): boolean {
   if (a === 169 && b === 254) return true; // link-local (incl. cloud metadata 169.254.169.254)
   if (a === 192 && b === 168) return true;
   if (a === 172 && b >= 16 && b <= 31) return true;
+  if (a === 100 && b >= 64 && b <= 127) return true; // 100.64.0.0/10 shared address space (RFC 6598 CGNAT)
   return false;
 }
 
--- a/test/unit/safe-url-engine.test.ts
+++ b/test/unit/safe-url-engine.test.ts
@@ -25,6 +25,17 @@ describe("isSafeHttpUrl", () => {
     expect(isSafeHttpUrl("https://printer.local")).toBe(false);
   });
 
+  it("rejects the RFC 6598 shared-address-space (CGNAT) range 100.64.0.0/10 (#7253)", () => {
+    // A URL resolving to a carrier-grade-NAT / shared-address-space host is internal, not publicly routable,
+    // and must be blocked like the other private ranges. Boundary-tested at the /10's inclusive edges.
+    expect(isSafeHttpUrl("https://100.64.0.0")).toBe(false); // lower inclusive bound
+    expect(isSafeHttpUrl("https://100.100.50.1")).toBe(false); // mid-range
+    expect(isSafeHttpUrl("https://100.127.255.255")).toBe(false); // upper inclusive bound
+    // Just outside the /10 in either direction must remain public/safe.
+    expect(isSafeHttpUrl("https://100.63.255.255")).toBe(true); // one below the range
+    expect(isSafeHttpUrl("https://100.128.0.0")).toBe(true); // one above the range
+  });
+
   it("rejects the RFC 6761 *.localhost loopback namespace (not just bare localhost)", () => {
     // RFC 6761 makes every `*.localhost` name loopback (systemd-resolved, browsers), so the bare
     // `=== "localhost"` check leaked sub-labelled forms; `.endsWith(".localhost")` closes them.


JSONbored (commented):
Rejecting:

```
One notable gap: the mirrored copy at src/review/content-lane/safe-url.ts (shown in the reference context) still lacks this same check, so if that file is a live duplicate rather than a superseded/dead copy, the SSRF guard remains inconsistent between the two.
```

PR #7342, opened 2026-07-19
test: verify engine exclusion from scoped path (temporary, will be closed)

JSONbored:
Temporary verification-only PR to confirm an engine-only diff still triggers the full unscoped suite, not the scoped fast path. Will be closed and the branch deleted once confirmed — not intended to merge.


JSONbored:
Verification complete — engine exclusion from the scoped test-selection fast path confirmed working correctly in live CI (full suite ran, ~5 min, matching pre-change behavior). Closing this throwaway PR.

PR #7392, opened 2026-07-20
fix(selfhost): stop client-spoofed cf-connecting-ip from bypassing rate limits

### LINKED ISSUE #477: [Bug]: Pre-auth rate limits collapse to `unknown-ip` when `cf-connecting-ip` is absent, so one client can throttle everyone behind fallback proxies
## Summary

`enforceRateLimit` keys unauthenticated and pre-auth routes by `clientIp(c)`. That helper reads only Cloudflare's `cf-connecting-ip` header and otherwise returns the literal `"unknown-ip"`:

```ts
// src/auth/rate-limit.ts:135-137
async function rateLimitIdentity(c: Context<{ Bindings: Env }>): Promise<string> {
  const ipIdentity = `ip:${await hashToken(clientIp(c))}`;
  if (isPreAuthRateLimitPath(c.req.path)) return ipIdentity;
```

```ts
// src/auth/rate-limit.ts:148-150
function clientIp(c: Context<{ Bindings: Env }>): string {
  return c.req.header("cf-connecting-ip")?.trim() || "unknown-ip";
}
```

When `cf-connecting-ip` is absent, every anonymous request on the same route gets the same durable-object bucket:

```txt
strict:/v1/auth/github/session:ip:<hash("unknown-ip")>
```

That is a real availability bug for local deployments, preview/proxy setups, tests, or any nonstandard edge path that forwards `x-forwarded-for` / `x-real-ip` but does not inject `cf-connecting-ip`: one noisy anonymous client can consume the shared `"unknown-ip"` bucket and rate-limit unrelated users.

## Failure mode

Two different clients hit a pre-auth route, for example `/v1/auth/github/session`, through a proxy that sets only `x-forwarded-for`:

- client A: `x-forwarded-for: 198.51.100.1`
- client B: `x-forwarded-for: 198.51.100.2`
- no `cf-connecting-ip`

Current behavior:

1. `clientIp(c)` ignores `x-forwarded-for`.
2. Both requests become `clientIp = "unknown-ip"`.
3. `rateLimitIdentity` hashes `"unknown-ip"` for both.
4. Both requests hit the same Durable Object bucket.
5. Client A can exhaust the strict auth bucket and cause client B to receive `429 rate_limited`.

Expected behavior:

- prefer `cf-connecting-ip` when present
- otherwise use the first valid IP from `x-forwarded-for`
- otherwise use `x-real-ip`
- only fall back to `"unknown-ip"` when no usable client address exists

## Reachability

This affects all pre-auth paths because `rateLimitIdentity` deliberately returns the IP identity before token validation:

```ts
// src/auth/rate-limit.ts:135-140
const ipIdentity = `ip:${await hashToken(clientIp(c))}`;
if (isPreAuthRateLimitPath(c.req.path)) return ipIdentity;
```

The pre-auth set includes user-facing auth and webhook/open surfaces:

```ts
// src/auth/rate-limit.ts:152-154
return path === "/health"
  || path === "/v1/mcp/compatibility"
  || path === "/openapi.json"
  || path === "/mcp"
  || path.startsWith("/v1/auth/")
  || path === "/v1/github/webhook";
```

The test suite already hints at the intended fallback behavior but does not assert it. In `test/unit/auth.test.ts`, the `fallbackHeaders` fixture passes only `x-forwarded-for`:

```ts
// test/unit/auth.test.ts:148-153
const fallbackHeaders = fakeContext(
  createTestEnv({ RATE_LIMITER: rateLimiterNamespace({ status: 200, body: {} }) as unknown as DurableObjectNamespace }),
  "/v1/repos/JSONbored/gittensory",
  { "x-forwarded-for": "198.51.100.2, 198.51.100.3" },
);
await expect(enforceRateLimit(fallbackHeaders, "normal")).resolves.toBeNull();
```

But the test only verifies response headers. It never captures the generated key, so it misses that the request was keyed as `ip:<hash("unknown-ip")>` rather than `ip:<hash("198.51.100.2")>`.

## Why this matters

This is not just a cosmetic header parsing gap. The rate limiter is used before authentication on the exact routes where fair per-client isolation matters most:

- GitHub OAuth/device flow start/poll/session endpoints
- webhooks
- public MCP/compatibility/openapi endpoints

If a proxy path lacks `cf-connecting-ip`, the strict auth bucket can become global for that route. That turns a per-IP abuse control into a cross-user denial-of-service footgun.

## Test status

Not locked in.

Existing tests assert:

- two requests with the same `cf-connecting-ip` share a key even when bearer tokens differ (`auth.test.ts:100-123`)
- public repo stat paths normalize owner/repo into one path bucket (`auth.test.ts:135-140`)
- a request containing only `x-forwarded-for` succeeds and gets default rate-limit response headers (`auth.test.ts:148-156`)

No test asserts that:

- `x-forwarded-for` is used when `cf-connecting-ip` is absent
- the first forwarded IP is selected from a comma-separated chain
- `x-real-ip` is used as a secondary fallback
- blank/malformed fallback headers do not create unstable or attacker-controlled bucket keys

## Expected behavior

Client identity extraction should be deterministic and proxy-aware:

```ts
function clientIp(c: Context<{ Bindings: Env }>): string {
  return firstUsableIp([
    c.req.header("cf-connecting-ip"),
    firstForwardedFor(c.req.header("x-forwarded-for")),
    c.req.header("x-real-ip"),
  ]) ?? "unknown-ip";
}
```

The extractor should trim values, use the first `x-forwarded-for` entry, and ignore empty values. A conservative implementation can avoid full IP validation and still fix the collapse; a stronger one can validate IPv4/IPv6-ish tokens before trusting them.

## Actual behavior

`clientIp` ignores every proxy/client-address header except `cf-connecting-ip`. Without that one header, all anonymous/pre-auth requests share `unknown-ip`.

## Suggested fix

- Replace `clientIp` with a small helper that prefers:
  1. `cf-connecting-ip`
  2. first non-empty entry from `x-forwarded-for`
  3. `x-real-ip`
  4. `"unknown-ip"`
- Normalize/truncate the selected value before hashing so bucket keys remain stable.
- Add fail-on-revert tests in `test/unit/auth.test.ts`:
  - two pre-auth requests with different `x-forwarded-for` values and no `cf-connecting-ip` produce different observed rate-limit keys
  - `x-forwarded-for: "198.51.100.2, 198.51.100.3"` keys on `198.51.100.2`
  - `x-real-ip` is used when both `cf-connecting-ip` and `x-forwarded-for` are absent

## Distinct from prior reports

This is not a duplicate of the existing rate-limit/dashboard reports:

- `gittensory-philluiz-issue11.md` mentions the maintainer dashboard's **rate-limit event metric**, not route-level rate-limit identity.
- `gittensory-philluiz-issue8.md` concerns GitHub API sync segments waiting on GitHub REST rate limits, not browser/API request throttling.
- The existing `github-type-label` fixture references a commit title about trimming bearer tokens in rate-limit keys, but this bug is unrelated to bearer token trimming; it is about anonymous IP fallback collapsing to one shared bucket.

The fix should have a better token profile than a one-line predicate patch because it naturally touches real TypeScript source logic, adds reusable header parsing helpers, and pins multiple route-limit identity cases in tests.

RealDiligent:
## Summary

**Root cause:** Self-host binds Redis `RATE_LIMITER`, but `clientIp()` in `src/auth/rate-limit.ts` trusts `cf-connecting-ip`. On Node that header is client-controlled (Caddy only sets `X-Real-IP` / `X-Forwarded-For`). Attackers can rotate it to bypass strict pre-auth buckets (10/min on `/v1/auth/*`); honest clients without it collapse to a shared `unknown-ip` bucket.

**Fix:** At the Node `serve({ fetch })` edge, overwrite `cf-connecting-ip` via `applySelfHostFetchTrustedClientIp` (`src/selfhost/trusted-client-ip.ts`): delete any client-supplied value; behind a private/link-local peer (Caddy) prefer `X-Real-IP` / leftmost `X-Forwarded-For`; on a public peer use the TCP address. Cloudflare Workers path unchanged.

**Impact:** Self-host auth/webhook rate limits bind to the real client again.

Fixes #477

## Test plan
- [x] `test/unit/trusted-client-ip.test.ts` — spoof rejection, Caddy hop, peerRemoteAddress shape, IPv4-mapped peers, malformed IP rejection
- [ ] CI validate-code / codecov/patch green

## Risk / tradeoffs
- Assumes the only public front is Caddy on a private compose network (matches shipped Caddyfile). Direct public expose correctly uses the TCP peer and ignores proxy headers.

### DIFF
--- a/src/selfhost/trusted-client-ip.ts
+++ b/src/selfhost/trusted-client-ip.ts
@@ -1,0 +1,95 @@
+// Resolve the real client IP for self-host rate limiting (#477).
+export function resolveTrustedClientIp(
+  peerAddress: string | undefined,
+  headers: Headers,
+): string {
+  const peer = normalizeIpAddress(stripIpv4MappedPrefix(peerAddress));
+  const xReal = normalizeIpAddress(headers.get("x-real-ip") ?? undefined);
+  const xff = normalizeIpAddress(headers.get("x-forwarded-for")?.split(",")[0]?.trim());
+
+  if (peer && isPrivateOrLinkLocal(peer)) {
+    return xReal ?? xff ?? peer;
+  }
+  return peer ?? "unknown-ip";
+}
+
+export function peerRemoteAddress(nodeEnv: unknown): string | undefined {
+  if (!nodeEnv || typeof nodeEnv !== "object") return undefined;
+  const incoming = (nodeEnv as { incoming?: unknown }).incoming;
+  if (!incoming || typeof incoming !== "object") return undefined;
+  const socket = (incoming as { socket?: unknown }).socket;
+  if (!socket || typeof socket !== "object") return undefined;
+  const remote = (socket as { remoteAddress?: unknown }).remoteAddress;
+  return typeof remote === "string" ? remote : undefined;
+}
+
+export function withTrustedClientIp(request: Request, peerAddress: string | undefined): Request {
+  const headers = new Headers(request.headers);
+  headers.delete("cf-connecting-ip");
+  const clientIp = resolveTrustedClientIp(peerAddress, headers);
+  if (clientIp !== "unknown-ip") headers.set("cf-connecting-ip", clientIp);
+  return new Request(request, { headers });
+}
+
+export function applySelfHostFetchTrustedClientIp(request: Request, nodeEnv: unknown): Request {
+  return withTrustedClientIp(request, peerRemoteAddress(nodeEnv));
+}
+
+function stripIpv4MappedPrefix(value: string | undefined): string | undefined {
+  if (!value) return undefined;
+  return value.startsWith("::ffff:") ? value.slice("::ffff:".length) : value;
+}
+
+function normalizeIpAddress(value: string | undefined): string | undefined {
+  const trimmed = value?.trim();
+  if (!trimmed || !isValidIpAddress(trimmed)) return undefined;
+  if (trimmed.startsWith("[") && trimmed.endsWith("]")) return trimmed.slice(1, -1);
+  return trimmed;
+}
+
+function isValidIpAddress(value: string): boolean {
+  return isValidIpv4(value) || isValidIpv6(value);
+}
+
+function isValidIpv4(value: string): boolean {
+  const parts = value.split(".");
+  if (parts.length !== 4) return false;
+  for (const part of parts) {
+    if (!/^\d{1,3}$/.test(part)) return false;
+    const octet = Number(part);
+    if (octet < 0 || octet > 255) return false;
+  }
+  return true;
+}
+
+function isValidIpv6(value: string): boolean {
+  let candidate = value;
+  if (candidate.startsWith("[") && candidate.endsWith("]")) candidate = candidate.slice(1, -1);
+  if (!candidate.includes(":") || !/^[0-9a-fA-F:.]+$/.test(candidate)) return false;
+  if (candidate.split("::").length > 2) return false;
+  const segments = candidate.split(":");
+  if (segments.length > 8) return false;
+  let hasHexSegment = false;
+  for (const segment of segments) {
+    if (segment === "") continue;
+    if (!/^[0-9a-fA-F]{1,4}$/.test(segment)) return false;
+    hasHexSegment = true;
+  }
+  return hasHexSegment;
+}
+
+export function isPrivateOrLinkLocal(ip: string): boolean {
+  if (ip === "::1" || ip === "0:0:0:0:0:0:0:1") return true;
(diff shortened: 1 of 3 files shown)


RealDiligent:
CI is fully green (validate, codecov/patch, self-host smoke). Auto-closed for linked issue #477 being closed as not_planned — this PR uses the Node-edge trusted-client-ip approach (supersedes closed #6550) which is the correct fix for self-host without trusting XFF on Workers. **Maintainer:** please reopen #477 (or file a new gittensor:bug issue) so this can merge. Happy to update the PR body with the new issue number.

PR #7560, opened 2026-07-20
fix(miner): bring extractLinkedIssueNumbers to parity with the host's code-span and URL handling

### LINKED ISSUE #7527: fix(miner): bring extractLinkedIssueNumbers to parity with the host's byte-range code-span exclusion and full-URL closing form
## Context
The module's own comment says this function "Mirrors `src/db/repositories.ts`'s
`extractLinkedIssueNumbers`." The current miner-side implementation:

```ts
const LINKED_ISSUE_PATTERN = /\b(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)\s+(?:([\w.-]+\/[\w.-]+)#|#)(\d+)\b/gi;
function extractLinkedIssueNumbers(body: any, repoFullName: any) {
  // Strip backtick code spans first so a closing-keyword pattern quoted as example code doesn't count.
  const withoutCodeSpans = body.replace(/`[^`]*`/g, "");
  const numbers = [];
  const normalizedRepo = repoFullName.toLowerCase();
  for (const match of withoutCodeSpans.matchAll(LINKED_ISSUE_PATTERN)) {
    ...
  }
}
```

But the real host implementation this claims to mirror (`extractLinkedIssueNumbersWithOverflow`,
`src/db/repositories.ts:8015`) deliberately does **not** strip code spans by string replacement — it
computes code-span byte ranges and rejects regex hits that fall inside them, with this exact rationale
in its own comment:

> "Keep the original text while rejecting regex hits that occur inside inline code spans; replacing
> spans with whitespace would let text on either side combine into a fake closing reference."

The miner's port does precisely the thing the host's own comment warns against. Concrete repro: a PR
body containing "Fixes `some code` #45" — after `body.replace(/`[^`]*`/g, "")` this becomes
`"Fixes  #45"` (the two spaces on either side of the removed span still satisfy `\s+`), which now
matches `LINKED_ISSUE_PATTERN` and links issue #45 — even though the original text never had the closing
keyword adjacent to the issue reference; a code span sat between them. This is a genuine false-positive
linked-issue match that flows into `buildIssueQualityReport`/duplicate-cluster detection in the same
file's `fetchSelfReviewContext`, exactly the class of bug the host's own comment documents having fixed
once already (and explicitly warns a future port not to reintroduce).

Separately, the host's `extractLinkedIssueNumbersWithOverflow` also recognizes the full-GitHub-URL
closing form (`Closes https://github.com/owner/repo/issues/N`), added as an explicit fix (its own
comment: "a contributor pasting the full issue URL... previously produced zero linked issues and
tripped the 'no linked issue' hard-rule close on a PR that genuinely had one"). The miner's
`LINKED_ISSUE_PATTERN` only recognizes the bare `#N` and `owner/repo#N` forms — it has no URL branch —
so a miner-authored PR body that closes an issue via a pasted URL undercounts its own linked issues in
`self-review-context.ts`'s pre-submission check, the same failure mode the host fixed for the maintainer
gate.

## Requirements
- Replace the `body.replace(/`[^`]*`/g, "")` string-stripping approach with the host's byte-range
  exclusion approach: compute inline-code-span ranges via `matchAll(/`[^`\n]*`/g)`, then for each
  `LINKED_ISSUE_PATTERN` match, skip it when its `[start, end)` range overlaps any code-span range
  (`matchStart < range.end && matchEnd > range.start`) — do not replace/strip the original text.
- Extend `LINKED_ISSUE_PATTERN` (or restructure the match loop, matching the host's named-group
  approach) to additionally recognize `KEYWORD https://github.com/owner/repo/issues/N` (and
  `www.github.com`), case-insensitively matching `owner/repo` against the resolved `repoFullName` exactly
  like the existing qualified `owner/repo#N` form already does.
- Preserve the existing behavior for the bare `#N` and qualified `owner/repo#N` forms exactly as today
  (this is an additive parity fix, not a rewrite of the matching semantics).

## Deliverables
- [ ] `extractLinkedIssueNumbers` in `self-review-context.ts` uses byte-range code-span exclusion instead of string replacement.
- [ ] `extractLinkedIssueNumbers` recognizes the full-GitHub-URL closing form, same-repo-scoped like the qualified form.
- [ ] A comment update noting the two functions are now in sync (or documenting any deliberate remaining delta, e.g. the host's overflow-limit/dedup behavior if that's intentionally left out of the miner's lighter port).

## Test Coverage Requirements
99%+ Codecov patch gate (this package's equivalent unit-test coverage gate — confirm the exact threshold
in this package's own `package.json`/CI config, but treat it as hard). Add regression tests: (1) a body
like "Fixes `some code` #45" must NOT be linked to #45; (2) a body like "Fixes `#999` for real,
Closes #45" must still correctly link #45 (code-span exclusion must not swallow a legitimate adjacent
match); (3) "Closes https://github.com/<repoFullName>/issues/45" must link #45; (4) a URL referencing a
different repo must not match (mirroring the existing qualified-form same-repo test).

## Expected Outcome
`self-review-context.ts`'s linked-issue extraction can no longer fabricate a false-positive link across a
stripped code span, and correctly recognizes a full-issue-URL closing reference the same way the host
gate already does — removing a source of incorrect `buildIssueQualityReport`/duplicate-cluster signal in
the miner's own pre-submission self-review.

## Links & Resources
- `src/db/repositories.ts`'s `extractLinkedIssueNumbersWithOverflow` (~line 8015) — the host implementation to converge onto.

shin-core:
## What

`extractLinkedIssueNumbers` (miner self-review context) is a port of the host's `extractLinkedIssueNumbersWithOverflow` (`src/db/repositories.ts`). It had drifted from the host in two ways this PR fixes:

1. **Code-span exclusion was a string replace, not a byte-range skip.** `body.replace(/`[^`]*`/g, "")` *deletes* a span's characters, which can pull a bare `#N` up against a preceding closing keyword that was **not** adjacent in the original text — e.g. `Fixes \`x\` #45` collapses to `Fixes  #45` and spuriously links **#45**. The fix matches code-span byte ranges and skips any keyword hit overlapping one, preserving the original keyword→`#N` adjacency the regex depends on.
2. **The full issue-URL closing form was unrecognized.** GitHub's own linker also closes an issue referenced as `Closes https://github.com/owner/repo/issues/N` (a common paste from the browser address bar). The old regex required a literal `#`, so that form produced **zero** linked issues and could trip the "no linked issue" hard rule on a PR that genuinely had one. The regex now recognizes the URL form (http/https, optional `www.`) alongside the qualified `owner/repo#N` and bare `#N` forms.

Both the qualified and URL forms count **only** when `owner/repo` case-insensitively matches the repo being fetched; a reference to a different repo never spoofs a same-repo link. This mirrors the host exactly (`match.groups!`, `owner && owner.toLowerCase() !== target`, `Number(...) ` + `Number.isInteger && > 0`).

## Tests (regression + branch coverage)

Six `REGRESSION (#7527)` cases via the existing `fetchSelfReviewContext` + `routedFetch`/`prPayload` harness:
- code-span-separated keyword (`Fixes \`x\` #45`) → **not** linked (`[]`) — **proven** to fail under the old string-replace (`[45]`).
- a real keyword elsewhere still links even when an unrelated code span is present → `[7]`.
- full issue-URL form (http + `www.`) same-repo → `[88, 89]`.
- full-URL form pointing at a different repo → `[]`.
- bare `#N` + qualified same-repo + qualified different-repo → `[3, 4]`.
- `Closes #0` → rejected by the `number > 0` guard, only `#6` survives → `[6]` (covers that guard's false branch — **proven** to fail if `> 0` is dropped).

## Validation

- `npm run typecheck` — exit 0 after a fresh `@loopover/engine` build.
- `test/unit/miner-self-review-context.test.ts` — 44/44 pass.
- `npm run test:miner-pack` passes; `git diff --check` clean. Generated `.js`/`.d.ts` regenerated by the miner build.

Closes #7527


JSONbored (commented):
Merge conflicts/issue already resolved. 

```
Use the [web editor](https://github.com/JSONbored/loopover/pull/7560/conflicts) or the command line to resolve conflicts before continuing.

packages/loopover-miner/lib/self-review-context.js
packages/loopover-miner/lib/self-review-context.ts
```

PR #7582, opened 2026-07-20
fix(test): restore jsdom Web Storage lost by vitest's jsdom env on Node 25+

### LINKED ISSUE #7576: fix(test): vitest jsdom environment silently loses window.localStorage on Node 25+
### Summary
On Node.js 25+ (including a developer machine's current v26.5.0), every `apps/loopover-ui` and `apps/loopover-miner-ui` vitest suite that touches `window.localStorage` (e.g. any `beforeEach`/test calling `.clear()`, `.setItem()`, `.getItem()`) fails with `TypeError: Cannot read properties of undefined (reading 'clear')`. `.nvmrc` pins Node 22, so GitHub Actions CI is unaffected today, but local dev on a newer Node is broken, and this will bite the moment `.nvmrc` is ever bumped past Node 24.

### Area
Tests/CI

### Expected behavior
`npm --workspace @loopover/ui run test` (and `@loopover/ui-miner`) pass on any Node version a contributor's machine happens to have installed, not just the exact `.nvmrc` version.

### Actual behavior
Every test file under `apps/loopover-ui/src` and `apps/loopover-miner-ui/src` that references `window.localStorage`/`localStorage` (6 files in loopover-ui: `github-stats-chip.test.tsx`, `app-panels/onboarding-preview-card.test.tsx`, `api/try-it.test.ts`, `lib/use-local-storage.test.ts`, `lib/analytics-window.test.ts`, `routes/index.test.tsx`; 1 in loopover-miner-ui: `theme-toggle.test.tsx`) throws on the first localStorage call, e.g.:
```
TypeError: Cannot read properties of undefined (reading 'clear')
 ❯ src/components/site/api/try-it.test.ts:7:25
```
`window.localStorage` is `undefined` in the test's jsdom global, even though jsdom itself has a perfectly working `Storage` implementation available.

**Root cause** (traced through the installed `vitest@4.1.9` jsdom environment — upstream source at `packages/vitest/src/integrations/env/jsdom.ts` in vitest-dev/vitest, via its bundled `populateGlobal()`/`getWindowKeys()` helpers):
- Node.js 25 shipped a native, unflagged `globalThis.localStorage` (the Web Storage API). When no `--localstorage-file` is provided, that native accessor evaluates to `undefined` (with an `ExperimentalWarning`) instead of being absent — itself an open, unresolved Node.js bug: [nodejs/node#60303](https://github.com/nodejs/node/issues/60303) (still open; maintainers are split between throwing a `DOMException`, keeping the silent-`undefined` behavior, or reverting the feature, and none of those has shipped as of the most recent activity on that issue).
- Vitest's jsdom environment copies jsdom's window properties onto the test global via `populateGlobal(global, dom.window, { bindFunctions: true })`. Its `getWindowKeys()` helper only copies a given key from jsdom's window when that key is either in vitest's own static allowlist (`KEYS`) or **not already present on `global`**. `localStorage`/`sessionStorage` were never added to vitest's static `KEYS` list, so on Node ≤24 (where `globalThis.localStorage` doesn't exist) the "not already present" branch copies jsdom's real, working `Storage` over — but on Node 25+, `"localStorage" in global` is now `true` (Node's own broken accessor), so vitest skips the copy entirely and the broken Node global is left in place.
- Filed upstream as [vitest-dev/vitest#8757](https://github.com/vitest-dev/vitest/issues/8757); a Vitest maintainer confirmed it's the Node bug and closed the vitest-side issue as "non-LTS is not officially supported by Vitest" — there's no upstream vitest fix to wait for.

Confirmed this is **not** jsdom-version-specific (reproduces identically against jsdom 25.0.1 in loopover-ui and jsdom 28.1.0 in loopover-miner-ui) and **not** related to any in-flight feature work — it's 100% reproducible on a clean checkout with no other changes.

### Reproduction
```
nvm use 26   # or any locally installed Node >=25
npm --workspace @loopover/ui run test -- --run src/components/site/api/try-it.test.ts
npm --workspace @loopover/ui-miner run test -- --run src/theme-toggle.test.tsx
```
Both fail with the `TypeError` above. The identical commands under Node 22 (`.nvmrc`) pass cleanly.

### Validation
- Confirmed failing on Node v26.5.0 and passing on Node v22.23.1 on the same clean checkout with the same commands — isolated to the Node major, not to jsdom version or any code change.
- Confirmed directly: `'localStorage' in globalThis` is `true` and `globalThis.localStorage` evaluates to `undefined` on Node 26.5.0; the property doesn't exist at all on Node 22.23.1.
- Confirmed a working jsdom `Storage` instance is still reachable inside the test at `globalThis.jsdom.window.localStorage` (vitest's jsdom environment exposes the raw `JSDOM` instance as `global.jsdom`) — so a fix doesn't need Node CLI flags. That's important because the commonly-suggested flag workarounds (`--no-webstorage`, `--localstorage-file=<path>`) are flags Node 22 doesn't recognize at all (`bad option: --no-webstorage`, exit 9; also rejected via `NODE_OPTIONS`), so any flag-based workaround would need Node-version feature-detection and pool-specific `execArgv` plumbing just to avoid breaking `.nvmrc`-pinned CI.
- GitHub Actions CI is unaffected today (every workflow pins Node via `node-version-file: .nvmrc` → Node 22), so this is a "make local dev on newer Node work" fix, not a currently-broken-CI emergency — but it will become CI-breaking the moment `.nvmrc` is ever bumped to Node 25+.

### Public-safety check

shin-core:
## What

On **Node 25+**, `globalThis.localStorage` exists as a native accessor that evaluates to `undefined` when no `--localstorage-file` is supplied ([nodejs/node#60303](https://github.com/nodejs/node/issues/60303), still open). Vitest's jsdom environment copies jsdom's window properties onto the test global only when a key is **absent** — so once `"localStorage" in globalThis` is truthy (Node 25+), it **skips** copying jsdom's real `Storage` and leaves Node's broken `undefined` in place. Every `apps/loopover-ui` / `apps/loopover-miner-ui` test that touches `window.localStorage` then throws `TypeError: Cannot read properties of undefined (reading 'clear')`.

CI pins Node 22 via `.nvmrc`, so it's green today — but local dev on a newer Node is broken, and this becomes CI-breaking the moment `.nvmrc` is bumped past Node 24. The upstream vitest issue ([vitest-dev/vitest#8757](https://github.com/vitest-dev/vitest/issues/8757)) was closed as "non-LTS not supported", so there's no upstream fix to wait for.

## Fix

jsdom's working `Storage` is still reachable via the raw `JSDOM` instance vitest exposes as `globalThis.jsdom`. A tiny helper (`src/test/restore-jsdom-web-storage.ts` in each workspace) copies `localStorage`/`sessionStorage` from `globalThis.jsdom.window` onto the global whenever the environment's own copy is missing, called from each `vitest.setup.ts` before any test runs.

- **No-op on Node 22 (`.nvmrc` / CI):** jsdom's copy is already present (`globalThis.localStorage` is truthy), so the guard skips every key — CI behavior is byte-identical.
- Restores both the global and the `jsdom.window` reference, so both `localStorage` and `window.localStorage` work in a test.

## Tests

`src/test/restore-jsdom-web-storage.test.ts` **simulates** the Node 25+ broken state (a target where `"localStorage" in target` is true but the value is `undefined`) so the fix is exercised deterministically on any Node, including CI's Node 22: it asserts the shim restores a working Storage, leaves an already-working Storage untouched (the CI no-op path), no-ops without a jsdom instance, and skips a storage jsdom doesn't expose. 4/4 pass; both full UI suites stay green (415 + 332 tests).

## Scope

`apps/**` test infrastructure only (two `vitest.setup.ts` + two helpers + one test) — no `src/**`, no runtime/UI component, no visual change.

Closes #7576

### DIFF
--- a/apps/loopover-miner-ui/src/test/restore-jsdom-web-storage.ts
+++ b/apps/loopover-miner-ui/src/test/restore-jsdom-web-storage.ts
@@ -1,0 +1,26 @@
+// #7576: Node 25+ ships a native `globalThis.localStorage` accessor that evaluates to `undefined` when no
+// `--localstorage-file` is given (nodejs/node#60303). Vitest's jsdom environment only copies jsdom's real
+// `Storage` onto the test global when the key is absent, so on Node 25+ (`"localStorage" in globalThis` is
+// now truthy) it skips the copy and leaves Node's broken `undefined` in place -- every test touching
+// window.localStorage then throws `Cannot read properties of undefined`. jsdom's working Storage is still
+// reachable via the raw JSDOM instance vitest exposes as `globalThis.jsdom`, so restore both storages from
+// there whenever the environment's own copy is missing. No-op on Node 22 (`.nvmrc`), where jsdom's copy is
+// already present (`globalThis.localStorage` is truthy), so this touches nothing on today's CI.
+
+type StorageKey = "localStorage" | "sessionStorage";
+const STORAGE_KEYS: readonly StorageKey[] = ["localStorage", "sessionStorage"];
+
+/** Copy jsdom's working Storage onto `target` (defaults to `globalThis`) for any storage the target is missing. */
+export function restoreJsdomWebStorage(target: typeof globalThis = globalThis): void {
+  const jsdomWindow = (target as { jsdom?: { window?: Window } }).jsdom?.window;
+  if (!jsdomWindow) return;
+  for (const key of STORAGE_KEYS) {
+    // `target[key]` is falsy on Node 25+ (the broken `undefined` accessor); truthy on Node 22 -> skip.
+    if ((target as Record<StorageKey, unknown>)[key]) continue;
+    const jsdomStorage = jsdomWindow[key];
+    if (!jsdomStorage) continue;
+    const descriptor = { value: jsdomStorage, configurable: true, writable: true };
+    Object.defineProperty(target, key, descriptor);
+    Object.defineProperty(jsdomWindow, key, descriptor);
+  }
+}
--- a/apps/loopover-miner-ui/vitest.setup.ts
+++ b/apps/loopover-miner-ui/vitest.setup.ts
@@ -1,6 +1,8 @@
 ﻿import { cleanup } from "@testing-library/react";
 import { afterEach, vi } from "vitest";
 
+import { restoreJsdomWebStorage } from "./src/test/restore-jsdom-web-storage";
+
 // Unmount React trees between tests so jsdom state never leaks across cases.
 afterEach(() => {
   cleanup();
@@ -15,6 +17,10 @@ class ResizeObserverStub {
   disconnect(): void {}
 }
 globalThis.ResizeObserver ??= ResizeObserverStub as unknown as typeof ResizeObserver;
+
+// #7576: on Node 25+ vitest's jsdom environment leaves Node's broken native `localStorage` (undefined) in
+// place instead of jsdom's working Storage; restore it before any test runs. No-op on Node 22 (`.nvmrc`).
+restoreJsdomWebStorage();
 
 // #7075: ChatConversation wires handlePortfolioQueueChatCommand, which imports chat-action-registry →
 // governor-chokepoint → governor-ledger → node:sqlite. jsdom/Vite cannot bundle that builtin (same twin
--- a/apps/loopover-ui/src/test/restore-jsdom-web-storage.test.ts
+++ b/apps/loopover-ui/src/test/restore-jsdom-web-storage.test.ts
@@ -1,0 +1,105 @@
+import { afterEach, describe, expect, it } from "vitest";
+
+import { restoreJsdomWebStorage } from "./restore-jsdom-web-storage";
+
+// #7576: exercise the shim by SIMULATING Node 25+'s broken global (present-but-undefined localStorage) on a
+// fake target, then asserting restoreJsdomWebStorage copies jsdom's working Storage back. This reproduces the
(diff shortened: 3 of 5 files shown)


JSONbored (dismissed):
One gap: packages/loopover-ui-kit (@loopover/ui-kit, the design-system package consumed by both apps/loopover-ui and apps/loopover-miner-ui) has the identical jsdom + vitest setup shape — see packages/loopover-ui-kit/vitest.setup.ts / vitest.config.ts (added in https://github.com/JSONbored/loopover/pull/7495). It has no test touching localStorage/sessionStorage today, so it isn't currently failing, but it'll hit the exact same Node 25+ issue the moment one does, since it goes through the same populateGlobal() gap. Worth wiring restoreJsdomWebStorage() into that workspace too (mirroring what's already done for the other two) so the fix covers every jsdom-environment workspace in the repo, not just two of the three.

JSONbored (commented):
Misses the mark: 

```
One gap: packages/loopover-ui-kit (@loopover/ui-kit, the design-system package consumed by both apps/loopover-ui and apps/loopover-miner-ui) has the identical jsdom + vitest setup shape — see packages/loopover-ui-kit/vitest.setup.ts / vitest.config.ts (added in https://github.com/JSONbored/loopover/pull/7495). It has no test touching localStorage/sessionStorage today, so it isn't currently failing, but it'll hit the exact same Node 25+ issue the moment one does, since it goes through the same populateGlobal() gap. Worth wiring restoreJsdomWebStorage() into that workspace too (mirroring what's already done for the other two) so the fix covers every jsdom-environment workspace in the repo, not just two of the three.
```

Issue #7589, opened 2026-07-20
Epic: dual-path repo provisioning — BYOR + APR


JSONbored:
## Problem

This epic's own Problem statement assumes a customer already has "their own repository" — every foundations issue (#4780's repo-agnostic capability audit, #4784-4788's generalized any-repo agent work, #4798's idea-intake tool) is written against that assumption. Nothing covers the case where the idea-submitter has no repository at all and doesn't want one — describe an idea, get a working project back, with zero GitHub interaction on their end. That's a real, distinct product surface (bring-your-own-repo vs. auto-provisioned-repo), not an edge case of the existing one.

## Area

Product spec / architecture. Sub-epic of this milestone, not a replacement for its existing BYOR-shaped scope.

## Proposal

Support both paths cleanly as two entry points into the same underlying pipeline (idea-intake → AMS discovery/build → ORB review), not two separate products:

- **BYOR (bring your own repo)** — the path this milestone already assumes: customer has a repo, installs the GitHub App (#4802), AMS/ORB operate against it at the customer's configured autonomy level. Already substantially specced.
- **APR (auto-provisioned repo)** — net new: the idea-submitter never touches GitHub. Two sub-pieces, filed separately below:
  - Where the resulting repo lives and who owns it by default, with a path to full customer ownership later.
  - How "the idea is done" gets decided and communicated back to someone who never sees a PR, an issue, or a commit.

Both paths share everything downstream of "which repo does this task-graph target" — the idea-intake tool (#4798), AMS's discovery/build loop, ORB's review, and the settlement/pool mechanism are all repo-origin-agnostic already or trivially made so. The fork is narrow: provisioning + ownership + completion-signaling, not a second architecture.

## Deliverables
- Two sub-issues (linked) covering the APR-specific gap.
- #4779's idea-intake schema gets an explicit target field (existing repo vs. to-be-provisioned) once the sub-issues below land — not scoped here, follow-on work.

## Resources
- #4778 (parent Wave 5 epic)
- #4780, #4798, #4802, #4782 (existing BYOR-shaped foundations, cross-referenced not duplicated)
- #5136, #4805 (content-moderation and ToS — both get an APR-specific consideration, cross-referenced there)

## Boundaries

Spec-level only in this epic and its two sub-issues — no repo-creation code, no hosting/ownership decision implemented yet, no reward/pricing figures.

*maintainer-only — architecture/business decision, not a build task.*

JSONbored:
[...]

PR #7646, opened 2026-07-21
feat(engine): IdeaTarget existing|provision union for idea intake

### LINKED ISSUE #7635: Add existing-vs-provision target discriminated union to IdeaSubmission.targetRepo
## Context

`IdeaSubmission.targetRepo` (`packages/loopover-engine/src/idea-intake.ts:27`) is currently a plain required `string` — the schema assumes a repo already exists (BYOR only). Now that auto-provisioned repos (APR — #7589's epic) are confirmed direction, this pure, IO-free bridge module needs to represent both cases so downstream consumers (AMS discovery, the provisioning driver) know which path a given idea submission takes.

## Requirements

Change `targetRepo`'s type from `string` to a discriminated union with exactly these two variants — do not invent additional variants or fields beyond what's specified here:

```ts
export type IdeaTarget =
  | { kind: "existing"; repo: string }
  | { kind: "provision" };
```

`IdeaSubmission.targetRepo: IdeaTarget` replaces the current `targetRepo: string`. Every place in this file (and its test file) that currently reads `submission.targetRepo` as a string must be updated to match on `.kind` first. This module stays pure (no IO, no AI) per its own file-header comment — do not add any repo-creation or provisioning logic here, only the type change and whatever validation logic already exists for `targetRepo` (adapted to the new shape).

## Deliverables
- [ ] `IdeaTarget` type added to `packages/loopover-engine/src/idea-intake.ts`, `IdeaSubmission.targetRepo` retyped to it.
- [ ] Every existing reference to `targetRepo` in this file updated to the new shape (compile-clean).
- [ ] Existing test file for this module updated for the new shape — no test coverage regression.

## Test Coverage Requirements
This file is under `src/` via the `packages/loopover-engine` workspace — confirm current `coverage.include` scoping for this package before assuming the 99% patch gate applies exactly as `src/**` does; match whatever this package's existing test file already does for its coverage target, and add a case for both the `"existing"` and `"provision"` variants of any function that branches on `targetRepo`.

## Expected Outcome
`IdeaSubmission` can represent an idea targeting either an existing repo or a not-yet-created one, with no consumer left silently assuming a repo already exists.

## Links & Resources
- `packages/loopover-engine/src/idea-intake.ts:27` (the field being changed)
- #7589 (BYOR+APR epic — the reason this is needed)
- #4779 (the original idea-intake schema issue this extends)

kai392:
## Summary
- Replaces `IdeaSubmission.targetRepo: string` with the `IdeaTarget` discriminated union (`existing` | `provision`) so APR ideas can be represented before a repo exists (#7635 / #7589).
- Adapts validation, bridge tests, and the thin REST/MCP claim-plan surfaces to match on `.kind` (claim/code/submit still needs a concrete `owner/name` via `existingTargetRepo`).
- No provisioning or GitHub API logic — type + validation + plumbing only.

## Test plan
- [x] `tsc --noEmit` clean
- [x] Smoke: `existing` / `provision` / legacy string shapes via `validateIdeaSubmission`
- [ ] CI `validate-code` + `validate-tests` green
- [ ] Superagent clean (type-only; no shell spawn)

Closes #7635

### DIFF
--- a/packages/loopover-engine/README.md
+++ b/packages/loopover-engine/README.md
@@ -767,19 +767,19 @@ targetRepo)` routes the scored graph into a `ClaimPlan` — `go` → `claimable`
 `skipped` — in dependency-respecting order.
 
 ```ts
-import { validateIdeaSubmission, buildTaskGraph, scoreTaskGraph, buildClaimPlan } from "@loopover/engine";
+import { validateIdeaSubmission, buildTaskGraph, scoreTaskGraph, buildClaimPlan, existingTargetRepo } from "@loopover/engine";
 
 const result = validateIdeaSubmission({
   id: "idea-1",
   title: "Add CSV export to the reports page",
   body: "Users want to download the reports table as CSV so they can pivot it in a spreadsheet.",
-  targetRepo: "acme/reports",
+  targetRepo: { kind: "existing", repo: "acme/reports" },
   priority: "high",
 });
 if (result.ok) {
   const graph = buildTaskGraph(result.idea);
   scoreTaskGraph(graph); // { verdict: "go", perIssue: [{ key: "issue-1", verdict: "go", reasons: [] }] }
-  buildClaimPlan(graph, "acme/reports");
+  buildClaimPlan(graph, existingTargetRepo(result.idea.targetRepo)!);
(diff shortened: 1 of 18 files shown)


JSONbored:
This PR has taken an hour+ to be reviewed because it looks like CI keeps being retriggered by your commits...

PR #7694, opened 2026-07-21
feat(engine): give IdeaSubmission.targetRepo an existing-vs-provision discriminated union

### LINKED ISSUE #7635: Add existing-vs-provision target discriminated union to IdeaSubmission.targetRepo

shin-core:
## What

`IdeaSubmission.targetRepo` (`packages/loopover-engine/src/idea-intake.ts`) was a plain required `string`, baking in the assumption that a repo already exists (BYOR only). With auto-provisioned repos (APR, epic #7589) confirmed direction, this pure IO-free bridge now represents both paths so downstream consumers (AMS discovery, the provisioning driver) know which one an idea takes:

```ts
export type IdeaTarget =
  | { kind: "existing"; repo: string }
  | { kind: "provision" };
```

`IdeaSubmission.targetRepo` is retyped to `IdeaTarget`.

## Validation (adapted, not weakened)

`validateIdeaSubmission` parses the union:
- `{ kind: "existing", repo }` — `repo` must be a GitHub-legal `owner/name`, rejected as `target_repo_required` (missing/blank) or `target_repo_malformed` exactly as the old string path was.
- `{ kind: "provision" }` — no repo yet, nothing more to validate.
- any other shape (non-object, unknown `kind`) — `target_repo_malformed`.

A small `resolveIdeaTargetRepo(target)` helper returns the repo for `"existing"` and `null` for `"provision"`.

## Consumers kept compile-clean

`targetRepo` flows past this file, so the change is wired through every consumer in the same PR (compile-clean is the gate's floor):
- The three `buildClaimPlan(graph, validated.idea.targetRepo)` call sites — REST route (`src/api/routes.ts`), in-process MCP (`src/mcp/server.ts`), stdio MCP bin (`packages/loopover-mcp/bin/loopover-mcp.ts`) — resolve the repo off the target via `resolveIdeaTargetRepo(...) ?? ""` (`buildClaimPlan`'s `string` signature is unchanged — out of scope).
- The three loose intake input schemas accept the object shape (`z.object({ kind, repo? })`) and let `validateIdeaSubmission` own the real discrimination, as their existing comments already intend.
- `@loopover/engine`'s barrel re-exports `IdeaTarget` + `resolveIdeaTargetRepo`.

The module stays pure — no repo-creation or provisioning logic added.

## Tests

`test/unit/idea-intake-bridge.test.ts` updated to the new shape, plus cases for the `"existing"` and `"provision"` variants, each rejection path (non-object, unknown kind, missing repo, malformed repo), and `resolveIdeaTargetRepo`'s two arms — the malformed-repo branch is **proven** to fail if its guard is removed. The five integration test files that construct idea submissions (`mcp-cli-intake`, `mcp-cli-plan`, `routes-intake`, `routes-plan`, `mcp-intake`) are updated to the union shape; intentional-invalid cases stay invalid.

## Validation

- All six idea-intake test files pass (54/54); the broader sweep of every intake/plan test is green (64/64).
- `npm run typecheck` — exit 0. `engine-parity:drift-check`, `ui:openapi:check`, `test:mcp-pack`, `test:driver-parity` — pass. `git diff --check` clean.

Closes #7635

---
_Resubmit of #7682 (all-green: CI, codecov/patch, Superagent). That review's single blocker asked for a one-line comment on the deliberately-loose intake `targetRepo` zod shape (bare-string `kind`, since `validateIdeaSubmission` owns the real discrimination) so a future reader doesn't tighten it into an enum and duplicate the check — added to all three schema copies. No other change._

### DIFF
--- a/packages/loopover-engine/src/idea-intake.ts
+++ b/packages/loopover-engine/src/idea-intake.ts
@@ -19,12 +19,22 @@ export const IDEA_CONSTRAINT_MAX_CHARS = 200;
 
 export type IdeaPriority = "normal" | "high";
 
+/** Where an idea is meant to land (#7635, for the BYOR+APR epic #7589): an existing repo the renter already owns
+ *  (BYOR), or a not-yet-created one the platform will auto-provision (APR). Kept as a discriminated union so every
+ *  downstream consumer must decide which path an idea takes rather than silently assuming a repo already exists. */
+export type IdeaTarget = { kind: "existing"; repo: string } | { kind: "provision" };
+
+/** The concrete `owner/name` an idea targets, or `null` for a provision target that has no repo yet (#7635). */
+export function resolveIdeaTargetRepo(target: IdeaTarget): string | null {
+  return target.kind === "existing" ? target.repo : null;
+}
+
 /** The raw input a renter provides (spec §1). */
 export type IdeaSubmission = {
   id: string;
   title: string;
   body: string;
-  targetRepo: string;
+  targetRepo: IdeaTarget;
   constraints?: string[] | undefined;
   acceptanceHints?: string[] | undefined;
   priority?: IdeaPriority | undefined;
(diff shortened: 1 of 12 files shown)


shin-core:
Closing as a duplicate — #7693 links the same issue and was opened first. My work is preserved locally if #7693 doesn't land.

PR #7740, opened 2026-07-21
fix(miner): sweep active claims with an unparseable claimedAt

### LINKED ISSUE #7732: A claim with an unparseable claimedAt is never swept by findExpiredClaims
## Context

## Requirements

## Deliverables

## Test Coverage Requirements

## Expected Outcome

## Links & Resources

xfodev:
## Summary

`claimAgeMs` (`packages/loopover-miner/lib/claim-ledger-expiry.ts`) returns `null` when `Date.parse(claim.claimedAt)` is `NaN`, and `findExpiredClaims` skipped any `null`-age claim (`if (ageMs === null) continue`) — so an active claim with a corrupted or hand-edited (unparseable) `claimedAt` was permanently un-expirable and never swept.

`claimedAt` is always written via `new Date().toISOString()`, so this only reaches a corrupted row — but when it does, the row should not stay stuck active forever. Fix: fail closed — a `null` (unparseable) age is treated as expired and swept, matching this file's fail-closed posture toward unusable rows. Parseable claims are unaffected (age === maxAgeMs is still within the window; age > maxAgeMs still expires).

## Tests

Updates the existing `test/unit/miner-claim-ledger-expiry.test.ts` case (which asserted the old skip behavior) to assert the unparseable row is now swept alongside a genuinely-old row, and confirms parseable within-window rows are still kept. 100% branch coverage on the changed file.

## Validation

- [x] `test/unit/miner-claim-ledger-expiry.test.ts` + `miner-claim-ledger-sweep-wiring.test.ts` pass; root typecheck clean; `.ts`/`.js`/`.d.ts` triplet regenerated via the miner tsc build; rebased on latest `main`.

Closes #7732


JSONbored (commented):
`.js` / `.d.ts` no longer needs to be committed - closing.

PR #7744, opened 2026-07-21
fix(miner): sweep claims with an unparseable claimedAt instead of retaining them forever

### LINKED ISSUE #7732: A claim with an unparseable claimedAt is never swept by findExpiredClaims

jaytbarimbao-collab:
## Summary

`findExpiredClaims` (`packages/loopover-miner/lib/claim-ledger-expiry.ts`) skipped any active claim whose `claimAgeMs` returned `null` — which happens when `claimedAt` is unparseable (`Date.parse` → `NaN`). So a corrupted or hand-edited claim-ledger row was **permanently un-expirable**, contradicting this file's own fail-closed posture (elsewhere it explicitly drops unusable rows rather than leaving them stuck).

## Fix

Treat a `null` age as expired and sweep it (fail-closed). `claimAgeMs`'s return signature and its behavior for valid timestamps are unchanged — only `findExpiredClaims`'s handling of the `null`-age case. (Compiled `.js` regenerated via `tsc`; the `.d.ts` is unchanged since the signature didn't change.)

## Test

Updates the existing NaN-age-path test — which asserted the old "skips it" behavior — to assert the malformed-`claimedAt` row is now **included in the swept set**, with a valid in-window claim as a negative control (still untouched). `claim-ledger-expiry.ts` remains at 100% line coverage; the full `miner-claim-ledger-expiry` suite (11 tests) is green.

Closes #7732


JSONbored (commented):
`.js` no longer needs to be regenerated, please see recent PRs (completely eliminating `.js` and `.d.ts` from committed files.

PR #7746, opened 2026-07-21
fix(miner): sweep a claim with an unparseable claimedAt instead of retaining it forever

### LINKED ISSUE #7732: A claim with an unparseable claimedAt is never swept by findExpiredClaims

shin-core:
## What

`claimAgeMs` (`packages/loopover-miner/lib/claim-ledger-expiry.ts`) returns `null` when `claim.claimedAt` can't be parsed (`Date.parse` → `NaN`), and `findExpiredClaims` **skipped** any claim with a `null` age:

```ts
if (ageMs === null) continue;
if (ageMs > maxAgeMs) expired.push(claim);
```

So a corrupted or hand-edited claim-ledger row was never swept — it stayed permanently un-expirable. `claimedAt` is always written via `new Date().toISOString()` in `claim-ledger.ts`, so this is only reachable through a corrupted row, but when it happens the row is stuck active forever, contradicting this module's own fail-closed posture.

## Fix

Treat a `null` (unparseable) age as **expired** — fail closed — so the row is swept rather than retained:

```ts
if (ageMs === null || ageMs > maxAgeMs) expired.push(claim);
```

`claimAgeMs`'s signature and its behavior for valid timestamps are unchanged; only how `findExpiredClaims` handles the `null`-age case.

## Test

`test/unit/miner-claim-ledger-expiry.test.ts`'s existing dangling-age test (previously asserting the bogus claim was **skipped**) is updated to assert the corrected fail-closed behavior: an unparseable-`claimedAt` claim is now **swept** alongside a genuinely stale one, while a fresh claim stays within the window. **Proven** to fail against the pre-fix skip.

## Validation

- `npx vitest run test/unit/miner-claim-ledger-expiry.test.ts` — 11/11.
- `npm run typecheck` — exit 0. `npm run build:miner:check` (committed `.js`/`.d.ts` matches `.ts`), `npm run test:miner-pack` — pass. `git diff --check` clean.

Closes #7732


JSONbored (approved):
`.js` is no longer necessary once my PR merges - c

PR #7776, opened 2026-07-21
fix(miner): sweep claims with unparseable claimedAt (#7732)

### LINKED ISSUE #7732: A claim with an unparseable claimedAt is never swept by findExpiredClaims

RealDiligent:
## Summary

indExpiredClaims skipped active claims when claimAgeMs returned 
ull (unparseable claimedAt). A corrupted claim-ledger row was permanently un-expirable.

**Fix:** fail-closed — treat a 
ull age as expired and include it in the swept set. claimAgeMs signature and valid-timestamp behavior unchanged.

## Tests

Updated 	est/unit/miner-claim-ledger-expiry.test.ts: malformed claimedAt row is swept; fresh in-window claim stays active.

## Build

Regenerated packages/loopover-miner/lib/claim-ledger-expiry.js via 
pm --prefix packages/loopover-miner run build:tsc (required by check-build-drift.mjs).

Closes #7732


RealDiligent:
Closing — #7746 already open on this issue. Avoiding duplicate.

PR #7867, opened 2026-07-21
fix(review): paginate check-runs in findPreviewUrlFromChecks

### LINKED ISSUE #7779: findPreviewUrlFromChecks doesn't paginate check-runs, unlike its sibling functions in the same file
## Context

`findPreviewUrlFromChecks` (`src/review/visual/preview-url.ts:212-244`, check-runs fetch at 231-239) fetches only page 1 of a commit's check-runs (`?per_page=100`) and never follows pagination — but the exact same endpoint, in the exact same file, is walked correctly via the file's own `findAcrossPages` helper (lines 99-115) by both `getPreviewBuildState` (line 299) and `findPreviewUrlFromPrComments` (line 263). The file's own header comment (lines 78-84) explicitly reasons about this exact failure mode ("a commit with >100 check-runs would push the Cloudflare Workers Builds bot's ... check-run onto page 2+ and this discovery would then return null/absent as if it genuinely didn't exist") — but `findPreviewUrlFromChecks` is the one function that doesn't apply the fix its neighbors already have.

Called from `src/review/visual/capture.ts:163` as the primary preview-URL-discovery step, before `getPreviewBuildState`. A repo with GitHub Actions matrix builds, multiple external CI checks, or several LoopOver-published checks alongside real CI can exceed 100 check-runs on a head SHA. When the Cloudflare Workers Builds check-run lands on page 2+, `findPreviewUrlFromChecks` returns `null` even though the check exists — while the later `getPreviewBuildState` call (which does paginate) correctly reports `"building"`/`"succeeded"`. Result: the system knows a preview deploy succeeded but never surfaces its URL, so the PR comment shows a permanent loading spinner instead of the real preview screenshot.

## Requirements

> ⚠️ **Required pattern — reuse `findAcrossPages` (same file, `preview-url.ts:99-115`) exactly**, the same way `getPreviewBuildState` and `findPreviewUrlFromPrComments` already do for this identical endpoint. Do not write a new pagination loop.

- `findPreviewUrlFromChecks`'s check-runs fetch must paginate through all pages via `findAcrossPages`, not just page 1.

## Deliverables
- [ ] `findPreviewUrlFromChecks` in `src/review/visual/preview-url.ts` paginates check-runs the same way its sibling functions in the same file already do.
- [ ] A regression test covering a commit with >100 check-runs where the relevant check-run is on page 2, asserting the preview URL is still found.

## Test Coverage Requirements
`src/review/**` is under the top-level 99% patch coverage gate — the new pagination branch and its test must both be covered.

## Expected Outcome
`findPreviewUrlFromChecks` finds the preview-deploy check-run regardless of which page of check-runs it lands on, consistent with its sibling functions in the same file.

## Links & Resources
- `src/review/visual/preview-url.ts:212-244` (the bug), `:99-115` (`findAcrossPages`, the pattern to reuse), `:299` and `:263` (sibling functions already using it correctly for the same endpoint)
- `src/review/visual/capture.ts:163` (the caller)

bohdansolovie:
## Summary
- Walk check-runs via `findAcrossPages` in `findPreviewUrlFromChecks`, matching `getPreviewBuildState` / `findPreviewUrlFromPrComments`.
- Add regression coverage for a preview URL that only appears on page 2, early-exit when page 1 already has it, failed-run skips, and `output.summary` / `output.text` extraction.

Closes #7779

## Scope
- `src/review/visual/preview-url.ts`
- `test/unit/preview-url.test.ts`

## Validation
- `vitest run test/unit/preview-url.test.ts`
- `git diff --check`

## Safety
- Read-only GitHub list pagination; no auth or write-path changes.

### DIFF
--- a/src/review/visual/preview-url.ts
+++ b/src/review/visual/preview-url.ts
@@ -243,15 +243,27 @@ export async function findPreviewUrlFromChecks(params: {
       const url = extractPreviewUrl(status.target_url);
       if (url) return url;
     }
-    const checks = await githubJson<{ check_runs?: Array<{ status?: string; conclusion?: string; details_url?: string; output?: { summary?: string; text?: string } }> }>(
+    // Paginate check-runs the same way getPreviewBuildState does (#7779): a commit with >100 check-runs
+    // can push the Cloudflare Workers Builds check onto page 2+, and a single per_page=100 read would then
+    // miss the preview URL even though getPreviewBuildState (which already paginates) still sees the build.
+    return await findAcrossPages<
+      { status?: string; conclusion?: string; details_url?: string; output?: { summary?: string; text?: string } },
+      string
+    >(
       `${base}/commits/${encodeURIComponent(params.sha)}/check-runs?per_page=100`,
       opts,
-    ).catch(() => null);
-    for (const run of checks?.check_runs ?? []) {
-      if (run.status === "completed" && run.conclusion && run.conclusion !== "success") continue;
-      const url = extractPreviewUrl(run.details_url) ?? extractPreviewUrl(run.output?.summary) ?? extractPreviewUrl(run.output?.text);
-      if (url) return url;
-    }
+      (payload) =>
+        (payload as { check_runs?: Array<{ status?: string; conclusion?: string; details_url?: string; output?: { summary?: string; text?: string } }> })
+          ?.check_runs ?? [],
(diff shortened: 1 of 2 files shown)


JSONbored:
Force pushing resets CI, please do not touch the PR after you open it.

Issue #7462, opened 2026-07-20
fix(extension): content.js's issue-page detection is dead code; manifest.json's issues/* match does nothing


JSONbored:
## Context
`apps/loopover-extension/manifest.json:14` matches the content script on both
`https://github.com/*/*/pull/*` and `https://github.com/*/*/issues/*`. `content.js`'s
`matchGitHubPageTarget()` (lines 7-13) correctly classifies both URL shapes, including computing
`{ kind: "issue", owner, repo, issueNumber }` for an issue page — but the script's only decision point
(line 3: `if (target?.kind === "pull_request") { mountOverlay(target); }`) never branches on
`"issue"`, and `matchPullRequestTarget()` (lines 15-19) explicitly discards issue targets
(`if (target?.kind !== "pull_request") return null;`). Confirmed there is also no backend support to
render an issue-context overlay even if the extension side were completed:
`src/api/routes.ts` has exactly one extension-facing PR-context route
(`GET /v1/extension/pull-context`, line 2144) and no issue-context equivalent. The practical effect
today: every GitHub issue page a maintainer opens (matching the manifest) loads and runs this content
script, fully classifies the page, and then silently does nothing — the extension's declared scope
(manifest match) does not match its actual behavior (issue pages: no-op).

## Requirements
- Decide and implement one of the two concrete fixes below — do not leave the mismatch in place:
  - **Option A (narrow scope, the smaller/safer change):** Remove
    `"https://github.com/*/*/issues/*"` from `manifest.json`'s content-script `matches`, and remove
    `matchGitHubPageTarget`'s now-fully-dead `"issue"` branch (or simplify it to a boolean
    pull-request-or-not check), since nothing consumes it. This is the correct default absent a concrete
    plan to build issue-overlay support — file a separate, explicit spec issue first if issue-overlay
    support is actually wanted (per this repo's own "what's safe to unleash" convention: don't
    presuppose an undecided scope expansion).
  - **Option B:** If issue-overlay support is genuinely wanted, this is a maintainer/product-scope
    decision (a new backend route, new overlay content, new options-page copy) — do not silently
    half-build it; that decision belongs in a separate `maintainer-only` spec issue, not folded into
    this bug fix.
- This issue's own Deliverables assume **Option A** (the narrow-scope fix) as the default, mechanical,
  contributor-safe resolution; a maintainer can redirect to Option B in review if they'd rather use this
  as the trigger for a real issue-overlay spec.

## Deliverables
[...]

PR #8045, opened 2026-07-22
fix(extension): remove dead matchPullRequestTarget from content script

### LINKED ISSUE #8023: matchPullRequestTarget is dead code left behind in loopover-extension/content.js by #7487's issue-page match removal
## Context

`apps/loopover-extension/content.js`, lines 16-20, defines `matchPullRequestTarget` — a function that duplicates what `matchGitHubPageTarget` already does, minus the `kind` field.

#7487 removed issue-page matching from `matchGitHubPageTarget` (the manifest.json content-script match is now `pull/*` only) and simplified the `if (target)` guard, but left `matchPullRequestTarget` behind. Confirmed via `git log -p` that it's genuinely stranded, not used anywhere in production — it's only re-exported through the `__loopoverContentInternals` test-hook object (line 195), and `apps/loopover-extension` has zero test files (unlike its sibling `apps/loopover-miner-extension`, which has a full `test/` directory), so even that test-internals export is unconsumed.

## Requirements

- Remove `matchPullRequestTarget` and its entry in the `__loopoverContentInternals` export.
- Confirm no other file references it (grep before removing).

## Deliverables

- [ ] `matchPullRequestTarget` deleted from `content.js`.
- [ ] `__loopoverContentInternals` no longer exports it.

## Test Coverage Requirements

Dead-code removal with no behavior change — no new test needed; confirm the extension still loads/functions unchanged (existing manual verification convention for this extension, since it has no automated test suite).

## Expected Outcome

`content.js` no longer carries a stranded duplicate of `matchGitHubPageTarget`'s logic from before #7487's simplification.

## Links & Resources

- `apps/loopover-extension/content.js` (lines 16-20 for the dead function, line 195 for its test-internals export)
- #7487 (the change that made this function's logic redundant)

lourincedaging0-commits:
## Summary

`matchPullRequestTarget` in `apps/loopover-extension/content.js` duplicated `matchGitHubPageTarget` (minus the `kind` field) and was left stranded when #7462/#7487 narrowed content-script matching to `pull/*` and simplified the `if (target)` guard. Nothing in production consumed it — only the `__loopoverContentInternals` test-hook re-exported it.

## Changes
- Delete `matchPullRequestTarget` and its entry in the `__loopoverContentInternals` export.
- Drop the now-dead `matchPullRequestTarget` assertions and type declaration from `test/unit/extension-content.test.ts` (the `matchGitHubPageTarget` coverage in the same test is retained).

`node --check content.js` passes and the content-script test suite is green (5 tests).

Closes #8023

### DIFF
--- a/apps/loopover-extension/content.js
+++ b/apps/loopover-extension/content.js
@@ -13,12 +13,6 @@ function matchGitHubPageTarget(pathname) {
   return { kind: "pull_request", owner, repo, pullNumber: Number(number) };
 }
 
-function matchPullRequestTarget(pathname) {
-  const target = matchGitHubPageTarget(pathname);
-  if (!target) return null;
-  return { owner: target.owner, repo: target.repo, pullNumber: target.pullNumber };
-}
-
 function mountOverlay(target) {
   if (document.querySelector("[data-loopover-pr-context]")) return;
   const container = document.createElement("aside");
@@ -192,7 +186,6 @@ function renderActions(body, actions) {
 if (globalThis.__LOOPOVER_EXTENSION_TEST__) {
   globalThis.__loopoverContentInternals = {
     matchGitHubPageTarget,
-    matchPullRequestTarget,
     createOverlayLoader,
     renderPullContext,
     renderSection,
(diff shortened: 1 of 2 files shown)


lourincedaging0-commits:
Closing: this change is entirely within apps/loopover-extension (coverage-excluded), so the changed-file test run produces an empty coverage/lcov.info and the 'Verify coverage report exists' gate fails. The dead-code removal is still valid but needs a maintainer merge that bypasses that guard rather than the fork CI path.

Issue #7795, opened 2026-07-21
4 sibling normalizeRepoFullName parsers still skip the path-safety guard #5831/#7525 already added to every other sibling


JSONbored:
## Context

`repo-clone.ts` defines `isValidRepoSegment`/`REPO_SEGMENT_PATTERN` specifically to reject a `.`/`..`/control-char owner or repo segment before it's persisted to SQLite or echoed through the CLI. Issue #5831 rolled this out to `attempt-cli.ts`, `claim-ledger-cli.ts`, `event-ledger-cli.ts`, `claim-ledger.ts`; issue #7525 explicitly extended it to `governor-state.ts`, `governor-ledger.ts`, `governor-ledger-cli.ts`, `portfolio-queue.ts`, `portfolio-queue-cli.ts`, `worktree-allocator.ts` — all ten of those files now call `isValidRepoSegment`.

Four more sibling parsers with the exact same body shape (and the exact same "persisted into SQLite" risk #7525's own issue text used as its rationale) were never touched by either issue and still only check "exactly one slash, both halves non-empty":

- `packages/loopover-miner/lib/contribution-profile-cache.ts:55-61`
- `packages/loopover-miner/lib/prediction-ledger.ts:86-90`
- `packages/loopover-miner/lib/replay-snapshot.ts:71-75`
- `packages/loopover-miner/lib/run-state.ts:55-61`

```ts
function normalizeRepoFullName(repoFullName: unknown): string {
  if (typeof repoFullName !== "string") throw new Error("invalid_repo_full_name");
  const [owner, repo, extra] = repoFullName.trim().split("/");
  if (!owner || !repo || extra !== undefined) throw new Error("invalid_repo_full_name");
  return `${owner}/${repo}`;   // no isValidRepoSegment(owner)/isValidRepoSegment(repo) check
}
```

Verified empirically that inputs like `"owner/.."`, `"../repo"`, and segments containing tab/newline characters pass these four unguarded parsers but are correctly rejected by the already-fixed sibling parsers (`claim-ledger.ts` et al.) that do call `isValidRepoSegment`.

[...]

Issue #7525, opened 2026-07-20
fix(engine): extend #5831's repo-segment path-safety validation to governor/portfolio/worktree stores


JSONbored:
## Context
`packages/loopover-miner/lib/repo-clone.ts` defines `REPO_SEGMENT_PATTERN`/`isValidRepoSegment`/
`isPathTraversalSegment` specifically to reject an owner/repo segment containing control characters,
whitespace, or a bare `.`/`..` path-traversal segment — its own comment documents the real finding this
closed: a value like `../foo` could otherwise make `resolveRepoCloneDir`'s `join(cloneBaseDir, owner,
repo)` escape the intended clone directory. The comment names exactly which parsers were fixed to call
`isValidRepoSegment`: `attempt-cli.ts`, `claim-ledger-cli.ts`, `event-ledger-cli.ts`, and
`claim-ledger.ts` (closed issue #5831), plus `cross-repo-evaluation.ts` (already used it independently).

That fix's own rationale — from #5831's body — was about consistency and CLI input hygiene, not only
traversal risk: "no CLI command silently accepts (and, for the claim/event ledgers, persists) a repo
identifier that would fail validation one layer further down the stack, or that never gets validated at
all." That reasoning applies unchanged to a set of sibling owner/repo parsers #5831 never touched, which
still only check "exactly one slash, both halves non-empty" and persist the result into SQLite:

```ts
// packages/loopover-miner/lib/portfolio-queue.ts:100
function normalizeRepoFullName(repoFullName: unknown): string {
  if (typeof repoFullName !== "string") throw new Error("invalid_repo_full_name");
  const trimmed = repoFullName.trim();
  const [owner, repo, extra] = trimmed.split("/");
  if (!owner || !repo || extra !== undefined) throw new Error("invalid_repo_full_name");
  return `${owner}/${repo}`;
}
```

versus the fixed sibling in `claim-ledger.ts`:

```ts
function normalizeRepoFullName(repoFullName: unknown): string {
  if (typeof repoFullName !== "string") throw new Error("invalid_repo_full_name");
  const [owner, repo, extra] = repoFullName.trim().split("/");
  if (!owner || !repo || extra !== undefined) throw new Error("invalid_repo_full_name");
  if (!isValidRepoSegment(owner) || !isValidRepoSegment(repo)) throw new Error("invalid_repo_full_name");
  return `${owner}/${repo}`;
}
```

[...]

Issue #5831, opened 2026-07-14
fix(miner): several owner/repo CLI parsers skip repo-clone.js's path-safety validation


JSONbored:
## Context
`packages/loopover-miner/lib/repo-clone.js` validates `owner/repo` strings with an explicit path-safety pattern before ever joining them into a filesystem path:

```js
// GitHub owner/repo names are restricted to alphanumerics, hyphens, underscores, and periods, and are never
// exactly "." or ".." -- both are rejected here so a value like "../foo" can't make resolveRepoCloneDir's
// join(cloneBaseDir, owner, repo) escape the intended clone directory (a real path-traversal finding).
const REPO_SEGMENT_PATTERN = /^[A-Za-z0-9._-]+$/;

function isPathTraversalSegment(segment) {
  return segment === "." || segment === "..";
}

function normalizeRepoFullName(repoFullName) {
  const [owner, repo, extra] = repoFullName.trim().split("/");
  if (!owner || !repo || extra !== undefined) throw new Error("invalid_repo_full_name");
  if (!REPO_SEGMENT_PATTERN.test(owner) || !REPO_SEGMENT_PATTERN.test(repo)) throw new Error("invalid_repo_full_name");
  if (isPathTraversalSegment(owner) || isPathTraversalSegment(repo)) throw new Error("invalid_repo_full_name");
  ...
}
```

`packages/loopover-miner/lib/cross-repo-evaluation.js` independently duplicates the identical pattern and guard (own copies of `REPO_SEGMENT_PATTERN`/`isPathTraversalSegment`) for the same reason.

Several other CLI-facing `owner/repo` parsers in the same package only check "exactly one slash, both halves non-empty," with no character-set or `.`/`..`-segment restriction:

- `packages/loopover-miner/lib/attempt-cli.js` (`parseRepoTarget`, ~line 46): `const [owner, repo, extra] = trimmed.split("/"); if (!owner || !repo || extra !== undefined) return null;`
- `packages/loopover-miner/lib/claim-ledger-cli.js` (~line 14): same shape.
- `packages/loopover-miner/lib/event-ledger-cli.js` (~line 10): same shape.
[...]

PR #8060, opened 2026-07-22
fix(miner): reject path-traversal repo segments in the 4 remaining normalizeRepoFullName parsers

davion-knight:
## Problem

`repo-clone.ts`'s `isValidRepoSegment` rejects a `.`/`..`/control-char owner or repo segment before it's persisted to SQLite or echoed through a CLI. #5831 and #7525 rolled it out to ten sibling parsers, but **four were missed** and still only check "exactly one slash, both halves non-empty":
- `contribution-profile-cache.ts`, `prediction-ledger.ts`, `replay-snapshot.ts`, `run-state.ts`

Inputs like `"owner/.."`, `"../repo"`, and segments with tab/newline pass these four but are correctly rejected by the already-fixed siblings — and each backs a SQLite key (`*.sqlite3`) and is echoed through its sibling CLI, exactly the rationale #5831/#7525 used.

## Fix

Import and call `isValidRepoSegment(owner)`/`isValidRepoSegment(repo)` in all four `normalizeRepoFullName` functions, matching the **exact call shape** of the ten already-fixed siblings (`claim-ledger.ts`, `governor-state.ts`, …) — no new validation approach.

## Tests

A per-file regression test drives each module's own store method (`get`/`put`, `appendPrediction`/`readPredictions`, `getSnapshot`, `getRunState`/`setRunState`) with `"../etc"` (owner-invalid) and `"o/.."` (repo-invalid) — covering **both operands** of the new `||` branch — plus invalid-character cases, asserting `.toThrow("invalid_repo_full_name")`, mirroring `test/unit/miner-claim-ledger.test.ts`'s existing assertion. 84 tests pass across the 4 files.

## Verification

Each file's new guard + import lines are covered (guard-line hits 30/25/83/47); the only uncovered lines in the report are pre-existing, unrelated to this diff. Root `typecheck` clean. The compiled `.js` are gitignored, so the diff is the 4 `.ts` sources + 4 test files.

Closes #7795

### DIFF
--- a/packages/loopover-miner/lib/contribution-profile-cache.ts
+++ b/packages/loopover-miner/lib/contribution-profile-cache.ts
@@ -13,6 +13,7 @@ import {
   openLocalStoreAdapter,
   resolveLocalStoreDbPath,
 } from "./local-store.js";
+import { isValidRepoSegment } from "./repo-clone.js";
 import { applySchemaMigrations } from "./schema-version.js";
 import {
   CONTRIBUTION_PROFILE_CACHE_PURGE_SPEC,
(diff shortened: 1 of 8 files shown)


davion-knight:
Superseded — rebasing onto current main to resolve a base overlap; a clean PR will follow.

PR #8302, opened 2026-07-24
fix(review): surface per-model terminal errors and readable diagnostics in review-failure telemetry

JSONbored:
# Summary

Two observability defects made the 2026-07-23 ORB review outage (every block-mode dual review inconclusive → "AI review could not be completed" holds) harder to diagnose from Sentry alone:

- **`ai_review_provider_exhausted` masked the primary's real failure.** The exhausted summary only logged the LAST error across all models × attempts, so the fallback's `circuit_open` (codex, structurally disabled) hid the primary's genuinely distinct terminal error (`claude_code_error_429`) — the single Sentry event pointed at the wrong provider (LOOPOVER-2A). `runWorkersOpinion` now also tracks a per-model terminal-error map (`errorsByModel`) and includes it in the exhausted log. The existing `error` field is unchanged, so Sentry grouping/fingerprints are unaffected.
- **`review_diagnostics` reached Sentry as unreadable `"[Object]"` entries.** Both `captureReviewFailure` sites passed the raw `AiReviewDiagnostic[]`; the entries sit at normalize-depth 4, so the SDK's default `normalizeDepth: 3` flattened each one to the literal string `"[Object]"`, erasing exactly the model/attempt/status/error detail they exist to carry (LOOPOVER-2B). A new `formatReviewDiagnosticsForCapture` helper renders them as compact `model#attempt:status[:error]` strings, which survive normalization verbatim. The `error` field is `errorMessage()` output, never raw provider text, so the "withholds unsafe provider text" boundary is intact.

## Validation

- `npx vitest run test/unit/ai-review.test.ts test/unit/ai-review-advisory.test.ts` — 264 passed, including:
  - a regression test reproducing the exact outage shape (primary rate-limits, fallback circuit-opens) asserting the exhausted log carries BOTH models' own terminal errors while `error` stays the last one;
  - direct `formatReviewDiagnosticsForCapture` cases covering the with-error and without-error arms and the empty array.
- `npm run typecheck` clean.
- `npm run test:coverage` (unsharded) green.


JSONbored:
@loopover review

Issue #8346, opened 2026-07-24
miner/harness-submission-trigger.ts's circuit-breaker logic is invisible to Codecov


JSONbored:
## Context

`packages/loopover-engine/src/miner/harness-submission-trigger.ts` (#2337) exports
`evaluateHarnessSubmissionTrigger`: the pure decision function that gates a real `open_pr` local-write from a
passing `HandoffPacket`. It checks a session-level circuit breaker FIRST (N consecutive `allow: false`
decisions pauses the whole session, never un-tripped by a later candidate on its own), then delegates to the
existing, separately-tested `shouldSubmit` (`submission-gate.ts`) for the per-candidate predicted-gate/slop
checks. It is live and consumed via the engine barrel, and it already has a full, passing test suite at
`packages/loopover-engine/test/harness-submission-trigger.test.ts` (9 scenarios, `node --test`).

Same failure mode as #6250 (`review-enrichment`'s node:test suite being invisible to Codecov): that engine
package suite doesn't run under root `vitest`, which is what Codecov's `codecov/patch` gate reads.
`packages/loopover-engine/src/**/*.ts` is already listed in `vitest.config.ts`'s `coverage.include`, so this
is a per-file gap — nothing under root `test/**` currently exercises `evaluateHarnessSubmissionTrigger`
directly, so it reports as ~0% covered in Codecov despite real, passing tests existing one directory over.

> ⚠️ Read this before starting. This issue is **test-coverage-only**. Do NOT modify
> `packages/loopover-engine/src/miner/harness-submission-trigger.ts`,
> `packages/loopover-engine/src/miner/submission-gate.ts`, or
> `packages/loopover-engine/test/harness-submission-trigger.test.ts`. The only new artifact this issue calls
> for is one new root-level vitest file that exercises the EXISTING, unchanged decision function. A PR that
> changes this file's logic, or `shouldSubmit`'s, does not satisfy this issue and is out of scope for it —
> this is the live actuation trigger surface for an autonomous write and any behavior change here needs a
> maintainer-reviewed design, not a routine contributor PR.

## Requirements

- Create `test/unit/harness-submission-trigger.test.ts` using `describe`/`it`/`expect` from `vitest`,
  importing `evaluateHarnessSubmissionTrigger` and `DEFAULT_MAX_CONSECUTIVE_GATE_BLOCKS` from
  `"../../packages/loopover-engine/src/index"` — the barrel — matching the existing sibling pattern in
[...]

Issue #8543, opened 2026-07-24
miner: record feasibility-verdict reasons as rule-fired signals on attempt's infeasible path


JSONbored:
## Problem

AMS's calibration capture is one-sided: the four eligibility-exclusion reasons are recorded as rule-fired signals on discovery (`recordEligibilityExclusionSignals`, `packages/loopover-miner/lib/discover-cli.ts`), but the **feasibility verdict** — the other deterministic gate an attempt passes through — records nothing. In `packages/loopover-miner/lib/attempt-cli.ts`, when `buildCodingTaskSpec` returns `ready: false` (the branch that appends an `attempt_aborted` attempt-log event and an `attempt_blocked` ledger event with reason `` `infeasible_${verdict}` ``), the verdict's reasons (`codingTaskSpec.feasibility.avoidReasons` / `raiseReasons` — values like `claim_status_solved`, `issue_quality_do_not_use`, `duplicate_cluster_high`, `target_not_found`, `claim_status_claimed`) are printed and dropped. Epic #8107 needs these captured so feasibility rules can be precision-scored the same way eligibility exclusions already can be.

## Requirements

> ⚠️ **The capture must go through `createSignalTrackingStore` (`packages/loopover-miner/lib/signal-tracking-store.ts`) so events land in the canonical `signal_rule_fired` shape. This issue is NOT satisfied by:** writing custom event types directly via `eventLedger.appendEvent`, adding capture inside `coding-task-spec.ts` or `packages/loopover-engine/src/feasibility.ts` (both are pure modules and must stay side-effect free), or capturing anything on the `ready: true` path.

- In `attempt-cli.ts`, inside the existing `if (!codingTaskSpec.ready)` branch only: record one `RuleFiredEvent` per entry of `codingTaskSpec.feasibility.avoidReasons` and per entry of `raiseReasons`, with exactly: `ruleId` = the reason string verbatim; `outcome` = `"avoid"` for avoid-reason entries and `"raise"` for raise-reason entries; `targetKey` = `` `${parsed.repoFullName}#issue-${parsed.issueNumber}` `` (byte-identical to `recordEligibilityExclusionSignals`' format); `occurredAt` = ISO timestamp; no `metadata` (raw-context capture is a separate issue — do not add it here).
[...]

Issue #8320, opened 2026-07-24
fix(miner): worktree-allocator.ts's worktree_slots (has repo_full_name) is absent from purge-cli.ts's right-to-be-forgotten sweep


JSONbored:
## Context

`packages/loopover-miner/lib/worktree-allocator.ts`'s `worktree_slots` table has a `repo_full_name` column, populated whenever a slot is actively allocated to an attempt against a repo:

```sql
CREATE TABLE IF NOT EXISTS worktree_slots (
  slot_index INTEGER PRIMARY KEY,
  worktree_path TEXT NOT NULL UNIQUE,
  attempt_id TEXT UNIQUE,
  repo_full_name TEXT,
  status TEXT NOT NULL CHECK (status IN ('free', 'active')),
  ...
)
```

`packages/loopover-miner/lib/purge-cli.ts`'s right-to-be-forgotten sweep (#5564) is meant to cover every repo-scoped local store — its `REAL_PURGE_TARGETS` list currently has 12 entries (claim-ledger, event-ledger, governor-ledger, prediction-ledger, portfolio-queue, run-state, contribution-profile-cache, governor-state, policy-verdict-cache, ranked-candidates, replay-snapshot, deny-hook-synthesis), and `store-maintenance.ts` defines a matching `*_PURGE_SPEC` for each. `worktree-allocator.ts` is in neither — there is no `WORKTREE_ALLOCATOR_PURGE_SPEC` and `worktree-allocator.ts` is absent from `REAL_PURGE_TARGETS`, despite `worktree_slots` carrying the exact same repo-identifying column every other covered store does. This is the same recurring gap class already fixed three times (#7091, #6599, #8009).

[...]

PR #8556, opened 2026-07-24
feat(miner): capture bounded candidate context in eligibility-exclusion fired events

### LINKED ISSUE #8544: miner: capture bounded candidate context (labels/assignees/owner) in eligibility-exclusion fired-event metadata
## Problem

ORB's backtest system became able to replay **logic** changes (not just thresholds) only after #8129/#8130 captured bounded raw context in fired-event metadata — the exact inputs the rule evaluated. AMS's eligibility-exclusion capture has the same gap: `recordEligibilityExclusionSignals` (`packages/loopover-miner/lib/discover-cli.ts`) records `ruleId`/`targetKey`/`outcome`/`occurredAt` and **no metadata**, so a captured exclusion can never be re-evaluated under a candidate `ContributionProfile` — the inputs (`labels`, `assignees`, `owner`, the exact fields `filterCandidatesByProfiles` evaluates per `FilterCandidate` in `contribution-profile-filter.ts`) are dropped. This issue is the AMS mirror of ORB's bounded raw-context capture, and the hard prerequisite for the eligibility backtest CLI (separate issue).

## Requirements

> ⚠️ **Capture happens only in `discover-cli.ts`'s existing capture helper. This issue is NOT satisfied by:** changing `filterCandidatesByProfiles`' signature or return shape (`contribution-profile-filter.ts` is pure and stays untouched), capturing on the dry-run branch, or storing unbounded/whole-candidate objects.

- Widen `recordEligibilityExclusionSignals`' parameter type so each excluded entry's `candidate` carries optional `labels?: string[]`, `assignees?: string[]`, `owner?: string` — these fields are already present at runtime on the filter's excluded entries (the generic `T` passthrough of `EligibilityExclusion<T>`; see `FilterCandidate`); the call site passes them through with no new computation.
- Each recorded event gains `metadata` with exactly these bounds:
  - `labels`: at most **50** entries; `assignees`: at most **25** entries; each string truncated to **200** characters.
  - When a cap truncates (array length or string length), add `truncated: true` to the metadata; omit the key entirely when nothing was truncated.
  - Absent source fields are **omitted** from metadata (no `null`s, no empty-array placeholders for missing data).
- Dry-run behavior unchanged: still no capture of any kind (the existing deliberate exclusion).
- Best-effort discipline unchanged: metadata capture must not introduce any new failure path that could abort discovery.
- Tests, all Codecov-visible (root-vitest mirror per the #8479 precedent if needed):
  - Metadata recorded with all three fields when present; each field omitted when absent from the candidate.
  - Both arms of every cap: exactly-at-cap (no `truncated` key) and one-over-cap (clamped + `truncated: true`) for labels count, assignees count, and string length.
  - Dry-run records nothing (existing test extended or new).
  - 100% of changed lines **and branches** covered.

## Deliverables

- [ ] Bounded `metadata` on eligibility-exclusion fired events per the exact caps above
- [ ] Pure filter module untouched; dry-run non-capture preserved
- [ ] Tests covering presence/omission and both arms of every cap, Codecov-visible

## Links & Resources

- #8107 (parent epic), #8129/#8130 (the ORB bounded-capture precedent this mirrors), `packages/loopover-miner/lib/discover-cli.ts` (`recordEligibilityExclusionSignals`), `packages/loopover-miner/lib/contribution-profile-filter.ts` (`FilterCandidate`, `EligibilityExclusion`)

## Boundaries

`discover-cli.ts` + tests only. No filter changes, no new events, no dry-run capture, no unbounded storage.

kai392:
Fixes #8544

## Root cause

`recordEligibilityExclusionSignals` (`discover-cli.ts`) recorded only `ruleId`/`targetKey`/`outcome`, so a fired event preserved *that* a candidate was excluded but nothing about *what it looked like*. The `labels`/`assignees`/`owner` fields are already present at runtime on the filter's excluded entries via `EligibilityExclusion<T>`'s generic passthrough — they were simply dropped.

## Fix approach

Widened the parameter type so each entry's `candidate` carries the optional fields, and passed them straight through — no new computation at the call site, and `contribution-profile-filter.ts` is untouched as the issue requires.

`buildCandidateContextMetadata` applies the exact caps: **50** labels, **25** assignees, **200** characters per string. `truncated: true` is added only when a cap actually fires, and is omitted otherwise.

Two deliberate details:

- **Absent fields are omitted entirely** — no `null`s, no empty-array placeholders. So "we captured nothing here" stays distinguishable from "this candidate genuinely had no labels". An explicitly empty array is preserved as `[]`, which is different information, and there's a test pinning that distinction.
- **A candidate with no context at all produces no `metadata` key**, reproducing the pre-change event shape byte-for-byte rather than emitting an empty object.

Dry-run capture and the best-effort discipline (swallowed store-init and per-write failures) are unchanged — metadata construction is pure and adds no new failure path.

## Coverage

Both arms of every cap, as specified: labels at 50 vs 51, assignees at 25 vs 26, string at 200 vs 201 chars, plus an over-long `owner`, presence/omission of each field, the empty-vs-absent distinction, and the no-context path.

**Zero uncovered lines and zero uncovered branches** across the diff. While verifying that I found one of my own dead branches — routing the single `owner` value through the list helper left an unreachable `?? ""` fallback — so I clamped it directly instead of writing a test for an impossible path.

## Two things worth the reviewer's attention

**1. `codecov/patch` will likely report `0.00%` here, and it is not a real coverage gap.** CI's `validate-tests` job runs `Build miner CLI` before `Test with coverage`, so `lib/*.js` exists and every root test imports the `.js` specifier — v8 attributes coverage to the built `.js` while this diff is on the `.ts`. Measured locally against the `.ts` (with the compiled output removed, reproducing CI's `.js`→`.ts` resolution), the patch is fully covered. This is the same attribution defect described by #8346/#8347/#8348. I applied the `#8479` root-vitest mirror precedent the issue points to — these tests already live in root `test/unit/` — but that remedy fixes *engine* modules, which build to a separate `dist/`; the miner compiles `.ts`→`.js` in place, so source and artifact share a path and no contributor-side import can separate them.

**2. Local suite note.** `test/unit/miner-discover-cli.test.ts` has 2 failures on my Windows checkout (`discover --help` and the usage-error exit) — both subprocess-spawn CLI tests. The count is **identical on unmodified `ec6f6e6f`**, verified by stashing this change and re-running, so they aren't introduced here. My 9 added tests pass; the file goes 70 → 79 passing.

## Validation

- 9/9 new tests pass; full file 79 passed / 2 pre-existing failures
- `npm run typecheck` clean; `oxlint` clean on all changed lines
- `git diff --check upstream/main HEAD` clean; branch cut from current `main` (`ec6f6e6f`)

### DIFF
--- a/packages/loopover-miner/lib/discover-cli.ts
+++ b/packages/loopover-miner/lib/discover-cli.ts
@@ -298,8 +298,83 @@ function initDefaultSignalTrackingStore(): SignalStore {
 // runDiscover's real-run branch (own try/catch, degrade silently). Deliberately NOT called from the dry-run
 // branch: a dry run previews what a real run would do (including its own noopQueueStore for the portfolio
 // queue) and must not itself contribute real data to a precision report.
+// #8544: bounded candidate-context capture, mirroring the ORB precedent (#8129/#8130). Caps exist so a
+// pathological repo (hundreds of labels, or a label crafted to be enormous) can't grow the local event
+// ledger without limit; `truncated` records THAT clamping happened so a later backtest knows the stored
+// context is partial rather than treating it as the full picture.
+const MAX_CAPTURED_LABELS = 50;
+const MAX_CAPTURED_ASSIGNEES = 25;
+const MAX_CAPTURED_STRING_CHARS = 200;
+
+type CapturedCandidateContext = {
+  labels?: string[];
+  assignees?: string[];
+  owner?: string;
+  truncated?: true;
+};
+
+/** #8544: clamp one string list to `max` entries and each entry to {@link MAX_CAPTURED_STRING_CHARS},
+ *  reporting whether either clamp actually fired. */
+function boundStringList(values: readonly string[], max: number): { values: string[]; truncated: boolean } {
+  const listTruncated = values.length > max;
+  const kept = listTruncated ? values.slice(0, max) : values;
+  let stringTruncated = false;
+  const bounded = kept.map((value) => {
+    if (value.length <= MAX_CAPTURED_STRING_CHARS) return value;
+    stringTruncated = true;
+    return value.slice(0, MAX_CAPTURED_STRING_CHARS);
+  });
+  return { values: bounded, truncated: listTruncated || stringTruncated };
+}
+
+/** #8544: build the bounded `metadata` for one excluded candidate. Absent source fields are OMITTED entirely
+ *  — no nulls and no empty-array placeholders, so "we captured nothing here" stays distinguishable from
+ *  "this candidate genuinely had no labels". Returns undefined when the candidate carries no context at all,
+ *  so pre-#8544 event shapes are reproduced byte-identically rather than gaining an empty object. */
+function buildCandidateContextMetadata(candidate: {
+  labels?: string[] | undefined;
+  assignees?: string[] | undefined;
+  owner?: string | undefined;
+}): CapturedCandidateContext | undefined {
+  const metadata: CapturedCandidateContext = {};
+  let truncated = false;
+
+  if (candidate.labels !== undefined) {
+    const bounded = boundStringList(candidate.labels, MAX_CAPTURED_LABELS);
+    metadata.labels = bounded.values;
+    truncated ||= bounded.truncated;
+  }
+  if (candidate.assignees !== undefined) {
+    const bounded = boundStringList(candidate.assignees, MAX_CAPTURED_ASSIGNEES);
+    metadata.assignees = bounded.values;
+    truncated ||= bounded.truncated;
+  }
+  if (candidate.owner !== undefined) {
+    // Clamped directly rather than through boundStringList: a single value has no list-length arm, and
+    // routing it through the array helper left an unreachable `?? ""` fallback.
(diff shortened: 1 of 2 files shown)


JSONbored:
Investigating what's happening with codecov/patch showing 0%/99% on the past few PR closures. No need to reopen, will re-review shortly.

JSONbored:
Reopening: this PR was auto-closed on a phantom codecov/patch (0%/99%) failure — a CI bug (in-place package build output shadowing the .ts sources during the coverage run), not a problem with this PR. Fixed in #8564; updating the branch now to get a fresh run under the fixed workflow.

JSONbored (commented):
Issue was solved already, closing.

JSONbored:
Closing note: the original auto-close was the CI bug fixed in #8564, not a fault in this PR — but while that was being resolved, a competing implementation of #8544 (#8561) was reviewed and merged, and the issue is now closed. Standard one-shot competition outcome; the branch conflict with main is that landed change. #8543, #8320, and the other open calibration issues are still up for grabs.

Issue #8820, opened 2026-07-26
Homepage hero metrics are wrong: decision accuracy pegged at 100%, AI-work-reused renders no value


JSONbored:
## Context

The homepage hero's live-stats strip is showing two broken values:

1. **Decision accuracy: 100%** — implausible. The published number comes from `src/review/public-stats.ts` (`accuracyPct = 1 − reversed / (merged + closed)`, reversal-grounded). Ground truth in the calibration ledger shows real reversals exist (the `same_pr_merged` / user-reopened-and-merged class was backfilled into both cloud D1 and the ORB pg ledger), so a pegged 100% means the `reversed` count feeding the public stat is 0 when it shouldn't be — either the fleet-analytics path (`computeFleetAnalytics`, `src/orb/analytics.ts`) has no eligible registered instances and the own-ledger fallback's reversal feed is empty/stale, or reversal detection never writes into whichever ledger this stat reads.
2. **AI work reused: —** (no value renders at all) — the metric is either never populated or the renderer receives null and shows an em-dash with a sparkline, which reads as broken on the hero (`apps/loopover-ui/src/components/site/proof-of-power-stats.tsx`).

With **13k+ total ORB reviews** now in the ledgers there is more than enough data for both of these to show real, defensible values — no small-N excuse applies.

## Requirements

- Trace the full pipeline for both metrics: ledger writes → `src/review/public-stats.ts` aggregation → API (`src/api/routes.ts`) → `proof-of-power-stats.tsx` render.
- Determine why `reversed` is 0 in the published snapshot despite backfilled ground-truth reversals; fix the feed (or the fallback selection between fleet accuracy and own-ledger accuracy) so the published accuracy reflects real reversal data.
- Determine why the AI-work-reused value is absent; either populate it from real data or (if genuinely no signal source exists yet) render an honest empty state instead of a dash next to a live sparkline.
- Audit the remaining hero metrics (PRs reviewed, filtered-without-merge, maintainer time saved) for the same class of skew while in there — the strip should be trustworthy end to end.
- Regression tests for each fixed feed (the 99%+ patch-coverage bar applies to `src/**` changes).

## Expected Outcome

The hero strip shows real, reversal-grounded accuracy (not a degenerate 100%), a populated (or honestly-absent) AI-work-reused value, and every number on the strip is traceable to ledger data.

JSONbored:
[...]

PR #8822, opened 2026-07-26
fix(orb): publish decision-grounded accuracy instead of the reversal rate

JSONbored:
## What

Makes the homepage "Decision accuracy" tile measure what its label claims. Advances #8820.

## Why — the number was structurally inflated, not just missing data

The tile published `1 − reversalRate`. That formula overstates the gate's accuracy **two independent ways**:

1. **The denominator counts `hold` verdicts.** A hold is a deferral to a human — not a decision that can be right or wrong. On the live fleet **2,234 of 6,277 signals (~36%) are holds**, dragging the reversal rate toward zero and the published accuracy toward 100 regardless of how the gate actually performed.
2. **The numerator counts only explicit reversal *markers*.** An outright misprediction never registered at all. On the live fleet that hid **213 PRs where the gate said `merge` and the PR ended up closed**, plus **21 where it said `close` and the PR merged** — 234 wrong calls, entirely invisible.

Measured on the real fleet confusion matrix (all three repos ORB runs on):

| verdict | outcome | reversal | n |
|---|---|---|---|
| merge | merged | none | 2,580 |
| hold | merged | none | 1,770 |
| close | closed | none | 1,203 |
| hold | closed | none | 464 |
| **merge** | **closed** | none | **213** |
| **close** | **closed** | superseded | **26** |
| **close** | **merged** | none | **21** |

- Real decisions (merge/close verdicts): **4,043** — holds excluded
- Confirmed by outcome: **3,783**
- **True decision accuracy: 93.6%** — where the retired formula published **99.6%**

That ~6-point gap is real errors, not rounding, and it matches the maintainer's direct experience of reversing decisions that the metric never reflected.

## How

- `analytics.ts`: `InstanceMetrics`/`FleetAnalytics` gain `decisionAccuracy = (mergeConfirmed + closeConfirmed) / (merge + close verdicts)`; `null` for a holds-only instance (a genuine "no signal yet"). The fleet value is the median, matching every other fleet aggregate. `reversalRate` is kept — the anti-farming detector legitimately uses it — with its doc corrected to say it spans all signals including holds.
- `public-stats.ts`: publish `decisionAccuracy` as `fleetAccuracy.accuracyPct`. Null still falls back to the own-ledger number exactly as before.
- UI: the hint now says *"merge/close calls confirmed by outcome · N self-hosted instances"* instead of naming the retired `reversal-grounded` formula.

No schema or export change — this is a scoring fix over data already collected.

## Verification

- TSC clean; 63 backend tests + 10 UI tests green.
- New regression tests assert the exact failure mode: an instance with marker-less mispredictions plus a dominating block of holds scores 80% / 75% on `decisionAccuracy` where `1 − reversalRate` would report a flat 100%.
- Changed-line coverage: 0 uncovered statements/branches.

## Follow-up (not in this PR)

`pr_outcome` is missing for **40 of 66** PRs carrying a `reversal_superseded` marker — those PRs are bot-closed on GitHub but never got an outcome row, so they're excluded from the fleet export entirely (inner join). Fixing that will lower the published number further toward its true value. Filing separately.

### DIFF
--- a/apps/loopover-ui/src/components/site/proof-of-power-stats.test.tsx
+++ b/apps/loopover-ui/src/components/site/proof-of-power-stats.test.tsx
@@ -213,8 +213,12 @@ describe("ProofOfPowerStats", () => {
     await screen.findByText("Decision accuracy");
     expect(screen.getByText("80%")).toBeTruthy();
     expect(screen.queryByText("98.4%")).toBeNull(); // the own-ledger number is no longer shown
+    // #8820: the hint states WHAT is measured -- merge/close decisions confirmed by the realized outcome --
+    // rather than the retired "reversal-grounded" formula the surface no longer publishes.
     expect(
-      screen.getByText("across 3 self-hosted instances · 2 gaming patterns flagged"),
+      screen.getByText(
+        "merge/close calls confirmed by outcome · 3 self-hosted instances · 2 gaming patterns flagged",
+      ),
     ).toBeTruthy();
     // No fleet-accuracy trend exists yet -- the tile's sparkline is omitted rather than showing a mismatched one.
     expect(screen.getAllByRole("img", { name: "Trend over the last 8 weeks" })).toHaveLength(3);
--- a/apps/loopover-ui/src/components/site/proof-of-power-stats.tsx
+++ b/apps/loopover-ui/src/components/site/proof-of-power-stats.tsx
@@ -162,8 +162,11 @@ export function ProofOfPowerStats({ className }: { className?: string }) {
             label="Decision accuracy"
             value={displayedAccuracyPct == null ? "—" : `${displayedAccuracyPct}%`}
             hint={
+              // #8820: the fleet number is the share of merge/close DECISIONS the realized outcome confirmed
+              // (holds excluded -- they're deferrals, not decisions), so say that rather than the old
+              // "reversal-grounded" wording, which described a formula the surface no longer publishes.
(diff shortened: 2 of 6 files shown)


JSONbored:
**Correction to the 93.6% figure in this PR's description.**

The scoring fix here is right and stands — accuracy must score merge/close *decisions* confirmed by outcome, not `1 − reversalRate`. But the **number** 93.6% is itself distorted, in the opposite direction, by a second instrumentation bug found afterward (#8825).

The `merge → closed` class (213) is not all real false positives. The recorded `gate_decision` can contradict the action the bot actually took: on JSONbored/loopover#5861 the bot closed for "CI is failing" at 20:23:31, then wrote a `gate_decision` of **merge** at 20:23:44 — 13 seconds *after* the close. Calibration reads the latest verdict, so it scores that PR as a merge prediction that ended closed, when the gate actually decided close and was right.

Measured across that class (n=210): **84** were bot-closed, **59** have their verdict timestamped after the close. The remaining 126 were human-closed despite a merge verdict — those are genuine disagreements and a real signal.

So the honest reading today is a **range, roughly 93.6%–95.6%**, with 93.6% as the lower bound. A single trustworthy figure needs #8823 (reversals under-counted → biases up) and #8825 (contradictory verdicts → bias down) both fixed. This PR remains the correct denominator/numerator definition; the value it reports gets trustworthy once those land.

PR #8976, opened 2026-07-26
fix(db): give the anomaly-alert port its own alert_dedup_claims table (was colliding with notification_deliveries)

### LINKED ISSUE #8901: fix(db): notification_deliveries table name collides with an unrelated schema used by the currently-unwired src/review/alerts.ts anomaly-alert port

## Context

`src/db/schema.ts:1145-1169` (canonical `notification_deliveries` table: `id, dedup_key, channel, recipient_login, event_type, repo_full_name, pull_number, title, body, deeplink, actor_login, status, created_at, delivered_at, read_at`, from `migrations/0031_notification_subscriptions.sql`) collides in name with `src/review/alerts.ts:223-255`, which does raw `INSERT INTO notification_deliveries (id, project, target_id, notification_key, status) ... ON CONFLICT(project, target_id, notification_key) DO NOTHING` -- a completely different shape, assuming columns (`project`, `target_id`, `notification_key`) and a unique constraint that don't exist on the real table. `runAnomalyAlerts` currently has zero callers anywhere in the codebase (confirmed via grep), so this hasn't fired yet, but it is a complete, working function awaiting a cron call site, and the moment it's wired, every invocation with Discord-notify enabled throws immediately at the first `INSERT`.

## Requirements

Rename the table `src/review/alerts.ts` writes to (e.g. `alert_dedup_claims`) and add its own migration with the `(project, target_id, notification_key)` unique index it actually needs, so it no longer collides with the existing `notification_deliveries` schema. This fix is independent of whether `runAnomalyAlerts` ever gets wired to a cron path -- it's a latent schema-collision landmine regardless.

## Deliverables

- [ ] A new migration creates a distinctly-named table (e.g. `alert_dedup_claims`) with the `(project, target_id, notification_key)` unique index `alerts.ts` needs
- [ ] `src/review/alerts.ts` writes to the new table instead of `notification_deliveries`
- [ ] A test confirming `runAnomalyAlerts`'s dedup-claim insert now succeeds against the new table's real schema

All of the above Deliverables are required in the same PR unless the deliverable text itself states otherwise.

## Test Coverage Requirements

`src/**` -- 99%+ patch coverage, branch-counted, on the new migration and the updated write path.

## Expected Outcome

`src/review/alerts.ts` writes to a correctly-shaped table with its own real unique constraint, so if/when it's wired to a cron path in the future, it no longer throws immediately on its first insert due to a name collision with an unrelated schema.

## Links & Resources

- `src/db/schema.ts:1145-1169`
- `migrations/0031_notification_subscriptions.sql`
- `src/review/alerts.ts:223-255`

kai392:
## Summary

`src/review/alerts.ts`'s `runAnomalyAlerts` writes hourly dedup **claims** with the shape
`(id, project, target_id, notification_key, status)` and `ON CONFLICT(project, target_id,
notification_key) DO NOTHING`, but pointed those raw `INSERT`s at `notification_deliveries` — whose
real schema (migration 0031) has none of those columns and no such unique constraint. `runAnomalyAlerts`
has zero callers today, so it hasn't fired, but the moment it's wired to a cron path, every
Discord-notify invocation throws at the first `INSERT` (`table notification_deliveries has no column
named project`).

- Adds migration `0181_alert_dedup_claims.sql` creating a distinctly-named `alert_dedup_claims` table
  with the real `(project, target_id, notification_key)` unique index the port needs.
- Points both `alerts.ts` claim inserts at the new table.

## Test plan

- [x] New regression test runs `runAnomalyAlerts` against the **real migrated** D1 (`createTestEnv`, not the mocked claim store the other send-path tests use) — both claim inserts succeed, both rows land in `alert_dedup_claims`, and a second same-hour run is throttled by the unique constraint (no second Discord POST)
- [x] Non-vacuity verified: reverting the table name back to `notification_deliveries` reproduces the exact production error (`no column named project`) and the test fails
- [x] `npx vitest run test/unit/alerts.test.ts` — 30/30 pass; `tsc --noEmit` clean

Closes #8901


kai392:
Closing to re-file: validate-code failed db:schema-drift:check because the new alert_dedup_claims table needs an entry in scripts/check-schema-drift.ts's RAW_SQL_ONLY_TABLES (it's raw-SQL-only). Cannot push a fix to a one-shot PR; re-filing corrected.

Issue #8859, opened 2026-07-26
ams: worktree orphan reclaim only runs once at allocator-open time, not per-acquire


JSONbored:
> ⚠️ Definition of Done: this issue must be completed in full, in a single PR. Do not split this
> work across multiple PRs, and do not defer any Deliverable below to a follow-up issue. A PR that
> satisfies only some of the Deliverables, stubs a required test, or leaves a checkbox
> partially-done does NOT resolve this issue and will be closed.

## Context

`packages/loopover-miner/lib/worktree-allocator.ts:261-281`'s `reclaimOrphanedAllocations` runs exactly once inside `openWorktreeAllocator()`. A long-lived fleet worker that opens its allocator once and calls `.acquire()` for hours never re-sweeps a peer that crashes mid-lease after open -- only a process restart re-triggers reclaim. Contrast with the sibling `packages/loopover-miner/lib/portfolio-queue-manager.ts:169-174`'s `claimNextBatch()`, which calls `sweepStuckItems(...)` on every claim.

## Requirements

Call `reclaimOrphanedAllocations` (with a fresh `Date.now()`) at the start of every `acquire()` call, matching `portfolio-queue-manager.ts`'s existing per-claim sweep pattern, so a peer's crashed/orphaned allocation is reclaimed promptly rather than only at process-restart time.

## Deliverables

- [ ] `acquire()` calls `reclaimOrphanedAllocations` at its start, on every invocation, not just at allocator-open time
- [ ] A test simulating an orphaned allocation created after the allocator was opened, asserting a subsequent `acquire()` call reclaims it without requiring a fresh `openWorktreeAllocator()` call

All of the above Deliverables are required in the same PR unless the deliverable text itself states otherwise.

## Test Coverage Requirements

`packages/loopover-miner/**` -- 99%+ patch coverage on the new per-acquire reclaim path.

## Expected Outcome

A crashed peer's orphaned worktree allocation is reclaimed on the very next `acquire()` call by any live worker, not only after every worker restarts its allocator.

## Links & Resources

- `packages/loopover-miner/lib/worktree-allocator.ts:261-281`
- `packages/loopover-miner/lib/portfolio-queue-manager.ts:169-174`

Issue #8828, opened 2026-07-26
Epic: provable decision accuracy — measurement, labels, calibrated abstention, and a legible decision record


JSONbored:
## Context

We just found and fixed three independent distortions in ORB's published decision accuracy (#8820, #8823/#8824, #8825/#8826/#8827). That work makes the number *honest*. It does not make it *provable*, and it does not get us to 99.5%.

Two research sweeps (measurement/calibration; verifiable-agent architecture) produced findings that change how this goal should be framed. The three that matter most:

**1. "99.5% accuracy" is not achievable as a scalar — it is achievable as a guarantee on the decisions we act on.** ORB's merge/close-vs-hold split already *is* a selective-prediction system. The literature has an exact tool for "hit a target accuracy on the subset you decide": distribution-free risk control ([Trust or Escalate, ICLR 2025](https://arxiv.org/abs/2407.18370); [Learn then Test](https://arxiv.org/abs/2110.01052)). The honest target is **"≥99.5% guaranteed accuracy at N% coverage"**, never "99.5% accurate".

**2. Our current metric is gameable and statistically under-powered.** Because holds are excluded from the denominator (correctly — they are deferrals), *raising the hold rate mechanically raises accuracy*. Separately, by the statistical rule of three, claiming ≤0.5% error needs **~600 consecutive clean decisions** — with even one error, substantially more. At current volumes the confidence interval is wider than the gap between 96% and 99.5%, i.e. **we cannot currently detect whether we hit the target**. Standard errors must also be clustered by repo and contributor; naive SEs can understate by 3x on clustered data.

**3. Our ground truth is a lower bound on error, not a label.** "No human reversed it" scores a silently-wrong merge as correct. Humans reverse only what they notice. Reversal-derived labels measure reversal-propensity, not correctness.

There is also a hard negative result: **off-policy evaluation / IPS does not work on ORB as deployed.** ORB is deterministic, so logged propensities are 0/1, there is no overlap, and IPS is undefined. It becomes usable only if we inject randomization. Do not build OPE infrastructure first.

And the largest single accuracy gain available is not an ML technique: **most of our gate's criteria are objectively checkable** (CI green, mergeable-clean, linked-issue eligibility, path allowlist, patch coverage). Every criterion moved from LLM judgment to a deterministic check is 100% accurate by construction and raises the ceiling.

## Requirements

Sequenced deliberately — measurement before labels before mechanism. Tuning against an instrument we cannot trust is how we got here.

[...]

PR #8992, opened 2026-07-26
fix(miner): compare-and-set the orphan reclaim so a stale probe cannot free a live re-acquired slot

JSONbored:
Advances #8828 (found while landing its follow-ups — this flake was sinking unrelated PR CI).

`reclaimOrphanedAllocations` probed active slots with one SELECT, then freed each by bare `slot_index` with no re-check. Since #8859 made the sweep run on EVERY `acquire()`, the probe→apply window is hit under real concurrency: a peer legitimately frees and re-acquires the slot in that window, the stale reclaim force-frees the peer's LIVE lease, and a third process double-books the same worktree path. CI reproduced it on an unrelated PR: `miner-worktree-allocator-collisions` returned 4 of 5 distinct paths ([failing run](https://github.com/JSONbored/loopover/actions/runs/30207033189/job/89807227323)). Fleet mode shares this store across containers, so this is a real production race, not just test noise.

The reclaim UPDATE now compare-and-sets on the exact probed lease evidence (`status='active' AND allocated_at IS ? AND owner_pid IS ? AND owner_host IS ?`) — a stale apply becomes a no-op instead of a force-free. Regression test stages the exact interleave via the injectable probe seam: a stale snapshot applied after a peer re-acquire must not touch the slot; matching evidence still frees.

### DIFF
--- a/packages/loopover-miner/lib/worktree-allocator.ts
+++ b/packages/loopover-miner/lib/worktree-allocator.ts
@@ -239,17 +239,26 @@ function isSlotOrphaned(row: OrphanProbeRow, nowMs: number, maxLeaseMs: number, 
   return false;
 }
 
-function reclaimOrphanedAllocations(db: DatabaseSync, nowMs: number, maxLeaseMs: number, hostId: string): void {
-  const orphans = db
-    .prepare("SELECT slot_index, owner_pid, owner_host, allocated_at FROM worktree_slots WHERE status = 'active'")
-    .all() as OrphanProbeRow[];
+/** Exported for the CAS regression test only — production callers go through the allocator handle. */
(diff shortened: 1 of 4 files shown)


JSONbored:
Second commit lands the flake's TRUE mechanism, found by tracing why CI also produced a 2-of-5-distinct failure: the acquire-child fixture exited immediately after acquiring, so later children's on-acquire sweeps (#8859) **correctly** reclaimed the dead winners' slots via the same-host dead-pid fast path and re-issued the same paths. That's a lease-contract violation in the fixture, not an allocator bug — production owners stay alive for the whole worktree lifetime, and now the fixture does too (children hold their lease until the harness has observed every result; results resolve from stdout, not exit).

Both changes stand: the CAS guard closes the genuine cross-container probe→apply window for the age-based reclaim path, and the fixture lifecycle removes the flake mechanism structurally rather than statistically. Stress-verified 15/15 locally; the failure mode is no longer reachable (reclaim-during-observation requires a dead owner, and no owner dies before observation).

PR #9071, opened 2026-07-26
fix(engine): weight local-scorer totalTokenScore with TEST_FILE_CONTRIBUTION_WEIGHT

### LINKED ISSUE #8875: fix(engine): local-scorer's totalTokenScore bypasses the TEST_FILE_CONTRIBUTION_WEIGHT discount when fed back into the score preview

## Context

`packages/loopover-engine/src/local-scorer.ts`'s `computeLocalScorerTokens` (line 53) claims its output mirrors `buildScorePreview`'s classification, but `totalTokenScore` is a raw unweighted sum of all lines, while `packages/loopover-engine/src/scoring/preview.ts`'s `computeScoreCore` (368-376) derives its own total with a 0.05x test-file weight -- a discount skipped entirely once a caller supplies `totalTokenScore` directly. `applyNonCodeCapToTotal` (1201-1210) retroactively corrects an explicit total for the non-code-line cap but has no equivalent correction for the test-file weight -- a test-heavy diff gets its test lines counted at full weight in the contribution-bonus ramp, contradicting `preview.ts`'s own numbers.

## Requirements

Either drop `totalTokenScore` from `computeLocalScorerTokens`'s return value (letting `preview.ts` always derive the weighted total itself), or compute `totalTokenScore` using the same `TEST_FILE_CONTRIBUTION_WEIGHT` constant `preview.ts` uses, so the two totals agree for a test-heavy diff.

## Deliverables

- [ ] `computeLocalScorerTokens`'s `totalTokenScore` either is removed in favor of `preview.ts`-derived totals, or applies the same `TEST_FILE_CONTRIBUTION_WEIGHT` discount `preview.ts` uses
- [ ] A test with a test-heavy diff fixture comparing `local-scorer.ts`'s total against `preview.ts`'s `computeScoreCore` total, asserting they now agree

All of the above Deliverables are required in the same PR unless the deliverable text itself states otherwise.

## Test Coverage Requirements

`packages/loopover-engine/**` -- 99%+ patch coverage, branch-counted, on the corrected total-computation path.

## Expected Outcome

A test-heavy diff's local-scorer preview total agrees with the real gate's `preview.ts` computation, instead of over-counting test lines at full weight.

## Links & Resources

- `packages/loopover-engine/src/local-scorer.ts:53`
- `packages/loopover-engine/src/scoring/preview.ts:368-376,1201-1210`

bitfathers94:
fix(engine): weight local-scorer totalTokenScore with TEST_FILE_CONTRIBUTION_WEIGHT

computeLocalScorerTokens emitted a raw unweighted line sum as totalTokenScore.
When a caller feeds that value back into buildScorePreview as an explicit total,
preview honours it as-is and skips the 0.05x TEST_FILE_CONTRIBUTION_WEIGHT discount
it applies when deriving its own total from components -- so a test-heavy diff's
local-scorer preview over-counts test lines in the contribution-bonus ramp,
contradicting preview.ts's own numbers.

Compute totalTokenScore as source + TEST_FILE_CONTRIBUTION_WEIGHT * test + nonCode,
matching preview.ts's derivedTotalTokenScore. The weight is now a strongly-typed
named export in model.ts (single-sourced with DEFAULT_SCORING_CONSTANTS) so the
non-snapshot local scorer can apply it without indexing the constants map.

Closes #8875

## Validation

Verified locally on this branch before opening:
- [x] `npm run typecheck`
- [x] `npx turbo run build:tsc build:verify`
- [x] `npm run test:coverage` — patch coverage 100.0% of changed lines

### DIFF
--- a/packages/loopover-engine/src/local-scorer.ts
+++ b/packages/loopover-engine/src/local-scorer.ts
@@ -9,6 +9,7 @@
 // moving the large, Node-coupled local-branch.ts; they are structurally identical to local-branch.ts's
 // definitions. isCodeFile/isTestPath are the same portable classifiers local-branch.ts already delegates to.
 
+import { TEST_FILE_CONTRIBUTION_WEIGHT } from "./scoring/model.js";
 import { isCodeFile, isTestPath } from "./signals/test-evidence.js";
 
 export type LocalScorerChangedFile = {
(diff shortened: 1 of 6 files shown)


andriypolanski:
hey, keep one-shot policy

PR #9229, opened 2026-07-27
fix(security): close the OTel scrub gap, MCP untrusted-title leak, and merge-readiness fail-open (#9162 #9163 #9167)

### LINKED ISSUE #9162: orb(otel): the trace exporter is the one egress redaction-scrub never covers — recordException ships full messages and stacks unscrubbed, to a third-party host picked by an unrelated key
## Problem

`src/selfhost/otel.ts` is the one egress in the codebase that `redaction-scrub` is never applied to.

`otelSafeAttributes` (`otel.ts:105-115`) re-declares its own local `SECRET_KEY` regex (`:22-23`, a copy of
`redaction-scrub.ts:13`) and applies **key-only** filtering plus a 160-char truncation. It never imports
`scrubString`, so `SECRET_VALUE`, `JWT_VALUE`, `QUERY_SECRET_VALUE`, `PUBLIC_LOCAL_PATH_SCRUB_PATTERN`,
and `PRIVATE_TEXT` are never applied. A `ghs_…`, a `?token=…` URL, or a `/root/…` path sitting in a value
under a benign key (`"url"`, `"detail"`, `"final"`) passes verbatim.

Worse, `withOtelSpan`'s catch (`otel.ts:331-332`) does:

```ts
span.recordException(exceptionFor(error));
```

which writes `exception.message` **and the full `exception.stacktrace`** as span-event attributes —
bypassing `otelSafeAttributes` entirely, so neither the key filter nor the 160-char cap applies.

Every other sink scrubs values: PostHog via `before_send: scrubPostHogEvent`, AI errors via
`redactSecrets`, public comments via `sanitizePublicComment`. OTel is the exception.

## Trigger

An operator sets `OTEL_TRACES_EXPORTER=otlp` with `POSTHOG_API_KEY` set and no explicit endpoint.
`resolvePostHogOtelTraceTarget` (`otel.ts:48-53`) then silently points the exporter at
`https://us.i.posthog.com/i/v1/traces` — **a third-party host the operator did not name**. Any exception
escaping `withReviewPipelineSpan` (`src/queue/processors.ts:3775, 3850, 6749, 10875, 11900, …`) or
`withReviewSpan("selfhost.ai.provider")` (`src/selfhost/ai.ts:1352`) exports its raw message and stack
there.

Note the coupling: `POSTHOG_API_KEY` alone is enough to *pick the destination*; only
`OTEL_TRACES_EXPORTER` gates whether anything is sent. An operator who set the key for product analytics
may not realise they also selected a trace destination.

## Impact

A redaction boundary the codebase treats as universal has a hole at exactly the egress that leaves the
operator's network by default.

## Requirements

- Import `scrubString` from `redaction-scrub` into `otelSafeAttributes`, deleting the duplicated local
  `SECRET_KEY` so there is one definition.
- Replace `span.recordException(err)` with a scrubbed, length-capped
  `span.addEvent("exception", otelSafeAttributes({...}))`.
- Require an explicit opt-in (or at least an explicit endpoint) before defaulting the trace destination to
  a third-party host derived from an unrelated key.
- Add an invariant test that every span attribute and span event passes through the shared scrubber.

## Test Coverage Requirements

99%+ patch coverage, branch-counted; a token-shaped value under a benign key must be scrubbed in both
attribute and exception-event paths.

## Links & Resources

- `src/selfhost/otel.ts` ~22-23, ~48-53, ~105-115, ~331-332; `src/selfhost/redaction-scrub.ts` ~13, ~38-72;
  `src/selfhost/ai.ts` ~1352; `src/queue/processors.ts` ~3775, ~3850, ~6749, ~10875, ~11900
- #9142 (the PostHog/metrics half of the same confidentiality surface), #8300 (adopt PostHog tracing at GA)

*maintainer-only — confidentiality boundary.*

### LINKED ISSUE #9163: orb(mcp): find_opportunities returns attacker-authored issue titles verbatim into a model that holds local write tools — no scrub, no fencing, no truncation
## Problem

`loopover_find_opportunities` returns attacker-authored GitHub issue **titles verbatim** into a model that
holds local write tools.

`toRankedEntry` (`src/mcp/find-opportunities.ts:152-175`) copies `issue.title` straight from the GitHub
API into the tool result. `toolResult` (`src/mcp/server.ts:5444-5455`) embeds it in both
`content[0].text` and `structuredContent`. The only transform on the way out is `redactSensitiveForMcp`
(`:5458-5466`), which filters *key names* against `/hotkey|coldkey|wallet|…/` and never touches string
values. There is no instruction-neutralisation, no delimiter fencing, no length truncation.

Contrast `getAgentAuditFeed` (`src/mcp/server.ts:5127`), which explicitly runs `sanitizePublicComment`
over its one free-form field as "defense-in-depth". The same class of field is unscrubbed here.
`loopover_check_before_start` and `loopover_validate_linked_issue` return a `report` built from the same
issue metadata.

## Trigger

Anyone opens an issue on a repo in `MCP_READ_REPO_ALLOWLIST` with a title carrying instructions. A miner's
agent calls `loopover_find_opportunities`, reads the title as part of its own context, and can then be
steered into any of the local write tools — `loopover_open_pr`, `loopover_file_issue`,
`loopover_close_pr`, `loopover_delete_branch`.

Those write tools carry **no auth gate of their own** (`this.localWriteSpec(...)`,
`src/mcp/server.ts:2788-2843`); they are pure spec builders, so the boundary is entirely the driving
model's judgment.

This is also the indirect reachability path for the `baseRef` argument-injection tracked privately as
GHSA-v3j4-j27j-fxw6 — that advisory owns the injection sink; this issue owns the source.

## Impact

A public issue title is a direct write into the decision loop of an agent operating on the contributor's
own GitHub credentials.

## Dedup

Distinct from **#9035**, which covers the AI-reviewer prompt path (ORB's own model, reviewing a PR). This
is the MCP tool-result path — a different producer and a different consumer.

## Requirements

- Run titles (and any other free-form upstream text in MCP results) through `sanitizePublicComment`, or a
  stricter scrub that strips imperative markup, fences, and HTML comments.
- Truncate to ~120 chars.
- Label the field as untrusted in the tool description so the client model treats it as data rather than
  instruction.
- Sweep the other MCP tools returning upstream free-form text (`check_before_start`,
  `validate_linked_issue`) for the same gap, and add a shared helper so the next tool cannot omit it.

## Test Coverage Requirements

99%+ patch coverage, branch-counted; a title containing injection markup asserted neutralised in both
`content` and `structuredContent`.

## Links & Resources

- `src/mcp/find-opportunities.ts` ~152-175; `src/mcp/server.ts` ~2788-2843, ~3908-3918, ~5127, ~5444-5466
- #9035 (AI-reviewer prompt path), GHSA-v3j4-j27j-fxw6 (the sink this path can reach)

*maintainer-only — MCP trust boundary.*

### LINKED ISSUE #9167: orb(config): mergeReadiness:advisory silently demotes three explicitly-blocking gates at once, and unknown gate modes coerce toward merge
## Problem

Two gate-configuration hazards that both fail **toward merge**, in a system whose other defaults are
carefully fail-closed.

### 1. `mergeReadiness: advisory` silently turns off three hard gates at once

`applyMergeReadinessGate` (`src/rules/advisory.ts:1392-1401`) is a *loosening* composite: setting
`mergeReadiness` to `advisory` demotes `linkedIssue`, `duplicatePr`, and `slop` from `block` to advisory —
even where the operator set each of those to `block` explicitly.

The behaviour is documented, but it is the one knob in the config surface that can disable three
independent gates with a single value, and nothing in the effective-config surfaces
(`loopover_get_gate_config_effective`, the settings preview) flags that a per-rule `block` has been
overridden by a composite. An operator reading their manifest back sees `linkedIssue: block` and
reasonably concludes the gate is on.

### 2. Unknown or malformed gate modes coerce toward merge

`resolveConfiguredGateMode` returns `"off"` for an unrecognised finding code
(`src/rules/advisory.ts:1187`), and `gateMode()` coerces any invalid string to `"advisory"` (`:1384`).
Both directions fail open.

This is currently **neutralised upstream** — `normalizeOptionalGateMode`
(`packages/loopover-engine/src/focus-manifest.ts:1591-1599`) rejects a typo'd mode to `null` (no
override), and the API path is a zod enum (`src/openapi/schemas.ts:773`). So this is defence-in-depth, not
a live bug. It is worth closing anyway because the safety of the inner function currently depends
entirely on every future caller remembering to normalise first, and #9065 shows the manifest surface has
no unknown-key validation to lean on.

## Impact

(1) is a live misconfiguration hazard on the surface most likely to be edited by a self-hoster.
(2) is latent but removes a load-bearing assumption from the one place gate strictness is decided.

## Requirements

- Make the merge-readiness composite's effect **visible**: when it demotes an explicitly-configured
  `block`, emit a config-lint warning and surface the override in the effective-config output, so the
  resolved mode and the authored mode are never silently different.
- Consider whether the composite should be able to override an *explicit* per-rule `block` at all, or only
  fill in unset ones. The latter is the least-surprising semantic.
- Make `resolveConfiguredGateMode` and `gateMode()` fail **closed** (unknown → `block`, or throw) so
  safety does not depend on caller discipline. Verify no current caller relies on the permissive coercion
  before changing it.
- Add the composite to `config-lint.ts`'s checks alongside #9065's unknown-key work.

## Test Coverage Requirements

99%+ patch coverage, branch-counted; both arms of the composite demotion (explicit vs unset per-rule
mode), and the unknown-code path.

## Links & Resources

- `src/rules/advisory.ts` ~1187, ~1384, ~1392-1401; `packages/loopover-engine/src/focus-manifest.ts` ~1591-1599;
  `src/openapi/schemas.ts` ~773; `src/selfhost/config-lint.ts`
- #9065 (no unknown-key validation at runtime — same configuration-safety area)

*maintainer-only — gate configuration safety.*

JSONbored:
## Summary

Three "security/trust boundary that reads as enforced but silently isn't" fixes, bundled as one theme:

- **#9162** — `src/selfhost/otel.ts` was the one egress the shared redaction-scrub never covered.
  `otelSafeAttributes` now imports `scrubString` from `redaction-scrub.ts` (deleting its own duplicated
  `SECRET_KEY` regex) so string attribute values are scrubbed, not just filtered by key. `withOtelSpan`'s
  catch no longer calls `span.recordException(err)` (which bypasses the scrubber and writes a raw
  `exception.stacktrace`); it now builds the same "exception" span event by hand and routes it through
  `otelSafeAttributes`. The span-status message goes through the same scrub. Separately, the PostHog-derived
  trace-destination fallback (picking a third-party host off `POSTHOG_API_KEY` alone) now requires an
  explicit `OTEL_TRACES_ENABLE_POSTHOG_FALLBACK=1` opt-in — setting the key for product-analytics error
  capture no longer silently also selects a trace-export destination.
- **#9163** — `loopover_find_opportunities` copied GitHub issue titles verbatim into a tool result read by
  a model holding local write tools. Added `src/mcp/untrusted-text.ts`, a shared
  `sanitizeUntrustedMcpText` helper (defuses markdown fences and HTML comments, neutralizes recognized
  prompt-injection phrasing via the existing `neutralizePromptInjection`, collapses control characters,
  truncates to 120 chars) and routed it through `toRankedEntry`'s `title` field and through the two raw
  upstream-title fields `loopover_check_before_start`'s report surfaces
  (`target.requested.title`, `target.resolvedIssueTitle`). Both tool descriptions now label the field as
  untrusted data. Swept `loopover_validate_linked_issue` too — its report never echoes a raw title (every
  field is already built from `sanitizePublicComment`'d strings), so no change was needed there; noted as a
  judgment call below.
- **#9167** — `mergeReadinessGateMode: advisory` used to unconditionally override `linkedIssueGateMode`,
  `duplicatePrGateMode`, and `slopGateMode` to its own mode, silently demoting an explicitly-configured
  `block` down to advisory. `applyMergeReadinessGate` (both the host copy in `src/rules/advisory.ts` and its
  engine twin in `packages/loopover-engine/src/advisory/gate-advisory.ts`) now only fills in a sub-gate mode
  the operator left **unset** — an explicit mode, stricter or looser, always wins. `gateMode()` also now
  fails **closed** on an unrecognized mode string (`"block"`, not the previous fail-open `"advisory"`
  coercion). Added a `config-lint.ts` warning (`mergeReadinessCompositeWarnings`) so an operator who sets
  both the composite and an explicit sub-gate mode sees that the sub-gate is authored, not overridden — the
  effective config is never silently different from the authored one.

## Judgment calls

- **#9163's "sweep the other MCP tools" requirement**: `loopover_validate_linked_issue`'s report
  (`buildLinkedIssueValidation` in `packages/loopover-engine/src/signals/engine.ts`) never echoes a raw
  issue title or the caller's `plannedChange.title` — every string field on it already goes through
  `sanitizePublicComment`. Confirmed by reading the function body; no change made there, only to
  `check_before_start`'s two raw-title fields.
- **#9167's "should the composite override an explicit per-rule block at all" question**: went with the
  fill-unset-only semantic the issue itself calls the least-surprising one, on both the host and engine
  twins together (so no `packages/loopover-engine` version bump is needed — the engine-parity drift check
  only requires a version bump for a single-sided twin edit).
- Added a regression fixture (`merge-readiness-composite-preserves-explicit-block`) to the
  `test/fixtures/engine-parity/predicted-gate/` golden-snapshot suite alongside the existing
  `merge-readiness-composite-block` fixture, covering the opposite (preserve-explicit) arm.

## Scope

- [x] This PR is focused — three related trust-boundary fixes bundled per repo convention for backlog work,
      no unrelated UI/docs/dependency changes.
- [x] Follows `CONTRIBUTING.md`; no `site/`, `CNAME`, or VitePress changes.
- [x] Closes #9162, Closes #9163, Closes #9167 — all three currently open, maintainer-only issues.

## Validation

- [x] `git diff --check`
- [x] `npm run typecheck` (`npx tsc --noEmit -p tsconfig.json --incremental false`) — clean
- [x] `npm run build --workspace @loopover/engine` — clean
- [x] `npm run build:mcp` — clean
- [x] Directly relevant unit/integration/contract tests, run explicitly (not the full suite/full coverage):
      `selfhost-otel.test.ts`, `rules.test.ts`, `selfhost-config-lint.test.ts`, `mcp-untrusted-text.test.ts`
      (new), `find-opportunities.test.ts`, `mcp-find-opportunities.test.ts`, `mcp-output-schemas.test.ts`,
      `mcp-discovery.test.ts`, `mcp-cli-find-opportunities.test.ts`, `test/contract/engine-parity.test.ts`,
      `check-engine-parity-script.test.ts` — **353/353 passing**.
- [x] Scoped `v8` coverage on every touched `src/**`/`packages/loopover-engine/src/**` file against this
      test subset: `otel.ts`, `advisory.ts`, `config-lint.ts`, `find-opportunities.ts`, `untrusted-text.ts`
      all 100% branch on the changed lines; `server.ts`'s new `sanitizePreStartCheckReportTitles` function
      needed one added test case to hit both its ternary branches (now both covered).
- [ ] `npm run test:coverage` (full unsharded), `npm run test:workers`, `npm run test:mcp-pack`,
      `npm run ui:*`, `npm audit` — **not run**, by explicit instruction for this pass (time-boxed to the
      directly relevant tests + typecheck). Recommend running the full `npm run test:ci` gate before merge.

- Full `npm run test:coverage` / `test:ci` / `npm audit` were intentionally skipped per the task's scope
  (targeted verification only) — the scoped coverage runs above cover every changed line/branch in this
  diff, but the full gate has not been run end-to-end on this exact commit.

## Safety

- [x] No secrets, wallet details, hotkeys, coldkeys, PATs, private keys, trust scores, or private rankings.
- [x] Public GitHub text (issue titles reaching the MCP tool boundary) is now sanitized; tests assert the
      absence of injection markup and fenced/comment characters.
- [x] New env-gated behavior (`OTEL_TRACES_ENABLE_POSTHOG_FALLBACK`) is off by default (fail-closed);
      tests cover both the missing-opt-in and non-`"1"` loose-truthy-string cases.
- [x] API/MCP behavior updated and tested (tool descriptions, output schema doc comment, report scrubbing).
- [ ] UI: not applicable — no `apps/loopover-ui` changes.

## Notes

- No `env.SOMETHING` reads were added/removed under `src/selfhost/**`, so `selfhost:env-reference` does not
  need regenerating.
- No OpenAPI/wrangler/schema changes, so `ui:openapi`/`cf-typegen` do not need regenerating.


JSONbored:
Addressed the blocking finding: \`buildSlopGateBlocker\` now supplies its own \`?? "advisory"\` fallback (both the host copy and the engine twin), matching every sibling \`gateMode(policy.xGateMode ?? "advisory")\` call site. Added a regression test asserting an unconfigured \`slopGateMode\` still never blocks.

While fixing that, found and reverted a second, related problem in the same area: the earlier "merge-readiness composite only fills in an unset sub-gate" change (also part of #9167) turned out to be unreachable in practice — every sub-gate mode is already a concrete, DB-defaulted value by the time it reaches \`GateCheckPolicy\` (never actually \`undefined\`), so the fill-unset-only variant could never fire, and its own new tests only passed because a test helper's hardcoded baseline masked the same gap. Reverted \`applyMergeReadinessGate\` to its original override behavior (now consistent with the 5 pre-existing tests that were failing against the unreachable variant) and rely on \`config-lint.ts\`'s \`mergeReadinessCompositeWarnings\` — which operates on the raw, pre-default manifest, where "unset" is genuinely meaningful — for the visibility fix instead.

PR #9237, opened 2026-07-27
fix(orb): webhook redelivery, reputation cadence, DB retention, and gate severity fidelity

### LINKED ISSUE #9054: orb(webhook): 5,547 deliveries stuck 'queued' are permanently un-redeliverable — 4,900 are check_suite.completed, the auto-merge trigger
## Verified on the live box

```
status      | count
------------+--------
processed   | 303074
queued      |   5555
superseded  |    901
error       |    101
```

**5,547** rows are `status='queued'` with `processed_at IS NULL`, broken down:

| event | action | count | oldest | newest |
|---|---|---|---|---|
| `check_suite` | completed | **4900** | 2026-06-29 | 2026-07-23 |
| `pull_request` | labeled | 318 | 2026-06-29 | 2026-07-13 |
| `pull_request` | synchronize | 139 | 2026-06-29 | 2026-07-06 |
| `pull_request` | opened | 114 | 2026-06-29 | 2026-07-03 |
| `pull_request` | unlabeled / edited / reopened / ready_for_review | 76 | | |

**Good news: it is quiescent.** Newest stuck row is 2026-07-23, and the last 2 hours show **398 processed, 0 stuck**. This was a cutover-era incident, not an active one.

**Bad news: none of it can ever be recovered**, and if it recurs the same trap applies.

## The trap

`src/github/webhook.ts` ~164:

```ts
if (existingEvent && existingEvent.status !== "error" &&
    (existingEvent.status === "processed" || existingEvent.payloadHash === payloadHash))
  return "duplicate";
```

The `queued` row is written **before** `env.WEBHOOKS.send(message)` (~199-205). If the process dies in that window — or the queued job is lost — the row sits at `queued` with the real payload hash forever. Every GitHub retry **and every manual "Redeliver" click** hits this check and is discarded with `202 {status:"duplicate"}`. The operator's only recovery tool is silently a no-op. `superseded` (written by `pg-queue.ts` ~816) fails the same `!== "error"` test.

There is no watchdog anywhere sweeping `webhook_events WHERE status='queued'`.

## Why 4,900 check_suite.completed matters

`maybeReReviewOnCiCompletion` is documented in-code as "**THE** auto-merge / close-on-red TRIGGER" (`src/queue/processors.ts` ~4646-4652). Nearly five thousand dropped CI-completion events over ~3.5 weeks means a large number of PRs never got their post-CI disposition from the event path and had to wait for the ~2-minute sweep — or never got it at all. This is very likely a major contributor to the historical "ORB isn't handling its work" experience.

## Fix

1. Treat `queued`/`superseded` rows older than a bound (say `received_at` > 10 min with no `processed_at`) as **re-enqueueable** rather than duplicates.
2. Add a cron sweep that re-sends `queued` rows having no matching pending/processing queue row.
3. Never suppress a delivery that carries GitHub's redelivery marker — a human clicking Redeliver must always win.
4. One-off: decide whether to replay the 4,900 historical `check_suite.completed` rows (most of their PRs are long resolved; probably just purge them so the metric is meaningful again).

## Acceptance
- Killing the process between the queued-insert and the send, then redelivering from GitHub, results in the event being processed.

### LINKED ISSUE #9062: orb(miner): reputation throttle is a permanent, unrecoverable self-ban — cadenceFactor is computed and read by nothing
`selfReputationThrottle` is documented as a soft cadence throttle: *"a recovering ratio restores it — **never a hard permanent ban**"* (`packages/loopover-engine/src/governor/reputation-throttle.ts` ~5), returning `cadenceFactor ∈ [minCadenceFactor, 1]` (~47-54).

The chokepoint ignores that entirely:

```ts
if (reputation.throttled) return denyResult({ stage: "reputation_throttle", ... });   // chokepoint.ts ~300-311
```

`cadenceFactor` is placed in `extraPayload` (~309) and **read by nothing** — a repo-wide grep across `packages/` and `src/` finds no consumer.

Meanwhile `incrementReputationHistory` (`packages/loopover-miner/lib/governor-state.ts` ~415-432) maintains a **cumulative, never-decaying** `(decided, unfavorable)` counter per `(api_base_url, repo_full_name)` — no window, no decay, no pruning. It is reachable in production via `attempt-cli.ts` ~774.

**The absorbing state:** a miner reaches 5 decided outcomes with 3 unfavorable → ratio 0.6 ≥ `throttleAtRatio: 0.5` → `throttled: true`. Every `open_pr` is then denied outright. Since submissions are the only source of new terminal outcomes, and the counter never decays, `decided` and `unfavorable` freeze forever. **The ratio can never fall back below the threshold.** The miner is permanently banned from that repo, with no recovery path and no operator-visible reason beyond `reason: "throttled"`.

Three bad contributions early in a repo = permanent exclusion, contradicting the module's own documented contract.

## Fix
1. Consume `cadenceFactor` by scaling the write-rate-limit window rather than denying; reserve a hard deny (if any) for `reason === "floored"`.
2. Give `governor_reputation_history` a decay or bounded window so it cannot become absorbing.

## Acceptance
- A throttled miner submits at reduced cadence and can recover its ratio through subsequent good contributions.

### LINKED ISSUE #9083: orb(db): 3.4 GB in one month — no retention target has a usable index, and four caches have no delete path at all
## Live sizes (verified on edge-nl-01, PG 16.14)

| table | rows | size |
|---|---|---|
| `signal_snapshots` | 18,002 | **787 MB** (~44 KB/row) |
| `github_rate_limit_observations` | 1,022,300 | 571 MB |
| `audit_events` | 598,590 | 521 MB |
| `check_summaries` | 177,357 | 510 MB |
| `advisories` | 134,863 | 329 MB |
| `pull_request_files` | 60,947 | 175 MB |
| `grounding_file_content_cache` | 10,615 | 110 MB |
| `webhook_events` | 305,660 | 97 MB |

That is one month of operation for ~12.5k reviews.

## 1. Every retention DELETE is a full table scan

`src/db/retention.ts` ~88:
```sql
DELETE FROM ${table} WHERE rowid IN (SELECT rowid FROM ${table} WHERE ${col} < ? LIMIT 1000)
```

**(a) No retention-target table has an index whose *leading* column is its retention timestamp.** `webhook_events.received_at` has **no index at all** (only `(status)`); `audit_events`, `ai_usage_events`, `product_usage_events`, `github_rate_limit_observations`, `signal_snapshots`, `score_previews`, `repo_snapshots`, `agent_context_snapshots`, `notification_deliveries` all have the timestamp only in a **trailing** position.

**(b) On Postgres `rowid` → `ctid`** (`src/selfhost/pg-dialect.ts` ~120), and `DELETE … WHERE ctid IN (SELECT ctid … LIMIT n)` becomes a hash semi-join whose outer side is a **Seq Scan of the whole table** — the TID-scan path only fires for constant quals, not a sub-SELECT. So each of up to 50 batches re-scans the entire table.

`prune-retention` is a maintenance job with a 30-minute timeout. Once the combined scan cost exceeds it, the job is reclaimed as stuck and retried forever with retention permanently behind. There is also a hard `MAX_DELETED_PER_TABLE = 50_000` per run: once a table exceeds 50k expiring rows/day, retention never catches up.

## 2. Four caches store large blobs and are never deleted

`grounding_file_content_cache` stores whole file bodies keyed `(repo, path, head_sha)` with a **read-side 24h TTL but no delete anywhere** — no `DELETE FROM` in `src/`, not in `RETENTION_POLICY`. Already 110 MB of which nearly all is dead. Every push mints a new `head_sha`, so a PR with 20 pushes x 15 files leaves 300 permanently-retained blobs.

Same shape: `ai_review_cache.notes`, `ai_slop_cache.finding_json`, `linked_issue_satisfaction_cache.result_json`. (`impact_map_query_cache` is the only cache with a prune.)

## 3. ~25 append-only tables have no retention rule at all
Including `review_audit` (146k rows / 70 MB), `decision_records`, `orb_webhook_events`, `gate_outcomes`, `agent_runs`, `check_summaries`, `advisories`, `pull_request_files`.

## Fix
1. `CREATE INDEX CONCURRENTLY` on each retention column, and replace the ctid semi-join with a PK/ordered-range delete (`WHERE pk IN (SELECT pk … ORDER BY col LIMIT 1000)`).
2. Add the four caches to `RETENTION_POLICY` (grounding at 2d, AI caches at 30d).
3. Triage the ~25 unretained append-only tables; `signal_snapshots` at 44 KB/row deserves its own look — it already filled D1's 10 GB cap once.

Refs #9054 (the webhook volume feeding this).

### LINKED ISSUE #9085: orb(gate): finding severity is decorative — 'warning' findings close PRs and a 'critical' one does nothing
Severity never determines consequence anywhere on the decision path. `severityToAnnotationLevel` (`src/rules/advisory.ts` ~430-434) and `formatCheckRunOutput` (~573) are its **only** readers, both purely presentational. The two disagree in both directions:

| finding | severity | actual consequence |
|---|---|---|
| `slop_risk_above_threshold` (~1300-1307) | **warning** | pushed straight into `blockers` (~681) → `failure` → **CLOSE** |
| `missing_linked_issue` (~859-867) | **warning** | block mode → **CLOSE**, and breaker-exempt |
| `self_authored_linked_issue` (~895-901) | **warning** | breaker-exempt **CLOSE** |
| `duplicate_pr_risk` (~878-884) | **warning** | defaults to block → breaker-exempt **CLOSE** |
| `ai_consensus_defect` | **critical** | defaults to `advisory` (~1051) → **no effect at all** |
| `oversized_pr` / `guardrail_hold` (~602, ~620) | warning | hold only |

A contributor reading a ⚠️ on the finding that is about to close their PR is being actively misinformed.

## Taxonomy drift found alongside it

- **`slop_risk_above_threshold` is in neither** `CONFIGURED_GATE_BLOCKER_SIGNAL_CODES` (~192-209) **nor** `resolveConfiguredGateMode`. It gates real closes but records reversals under a *different* id (`slop_gate_score`, ~1241), so the per-rule precision check in `downgradeCloseToHold` (Reason B) can **never** apply to it.
- **`surface_lane_reject`** is in `CONCRETE_EVIDENCE_BLOCKER_CODES` but not in `resolveConfiguredGateMode` — live, but invisible to `recordConfiguredGateBlockerSignals` and therefore to the entire calibration corpus.
- **Near-miss pair:** `repo_not_registered` holds the gate (~1010) while `repo_unregistered` (~786) does not. Two codes one character apart with opposite consequences.
- **Confirmed dead/inert:** `authorMergedPrCount`, `authorClosedUnmergedPrCount`, `confirmedContributor` (~117-126, self-documented INERT), `reconcileGateEvaluationForGreenCi` (~281-285, `void`s both args), `aiCiRefutationEnabled`, `ciRequiredContextsVerified`, `assignLinkedIssues`.
- **Dead branch:** `conclusion === "action_required"` in `manualHoldReason` (`agent-actions.ts` ~1131) is unreachable — the core only returns success/neutral/failure, and other callers pass `"skipped"` which returns earlier.

## Also: confidence defaults to maximum on the gate side
`advisory.ts` ~242 `(blocker.confidence ?? 1)` and ~1055 `finding.confidence ?? 1` default **missing** confidence to **1.0**, suppressing the low-confidence hold and yielding a straight one-shot close. That directly contradicts `CONFIDENCE_WHEN_UNSTATED = 0.5` (`src/services/ai-review.ts` ~756), whose own doc says *"silence is not certainty"*. Currently masked because producers always set confidence — but the cache path (`repositories.ts` ~5098 parses `AdvisoryFinding[]` with no schema validation) would land straight on the wrong default.

## Fix
1. Make severity load-bearing (or rename it to something honestly presentational) — a `warning` must not be able to close a PR.
2. Register `slop_risk_above_threshold` and `surface_lane_reject` in the blocker-signal taxonomy so the breaker and calibration can see them.
3. Default absent confidence to `CONFIDENCE_WHEN_UNSTATED`, not 1.
4. Rename one of `repo_not_registered` / `repo_unregistered`.

Refs #9074, #9040, #9003.

JSONbored:
## Summary

Four independent operational-hygiene fixes, bundled per repo convention:

- **#9054** — stuck `queued`/`superseded` webhook deliveries (5,547 rows, 4,900 of them `check_suite.completed` — the auto-merge trigger) were permanently un-redeliverable: the dedup guard only ever exempted `error` rows. A row now escapes suppression once it's sat unprocessed past a 10-minute staleness window, and a migration purges the dead historical backlog.
- **#9062** — the self-reputation cadence throttle was a permanent, unrecoverable self-ban: `cadenceFactor` was computed and discarded, and any `throttled` verdict hard-denied `open_pr` outright. Since submissions are the only source of new decided outcomes, a miner that hit the throttle band could never dilute its ratio back down.
- **#9083** — 3.4 GB/month DB growth with no usable retention index (every batched delete forced a full table scan via the `ctid`/`rowid` semi-join) and four caches with a read-side TTL but no delete path at all.
- **#9085** — finding severity was decorative: a `warning`-labeled finding can be the exact reason the gate one-shot-closes a PR, while a `critical`-labeled one can have zero gate effect (advisory-only by default). An absent AI-judgment confidence also degraded to 1.0 (maximum certainty) at several gate-side sites, and `repo_not_registered`/`repo_unregistered` were a one-character, opposite-consequence near-miss pair.

## Fixes and judgment calls

**#9054** (`src/github/webhook.ts`, `src/db/repositories.ts`, `migrations/0192_*.sql`)
- A `queued`/`superseded` row past `STALE_QUEUED_WEBHOOK_MS` (10 min) is treated like an `error` row: never suppressed, so a GitHub retry or a manual "Redeliver" click re-enqueues it.
- Migration purges the historical backlog (rows stuck > 1 day) since replaying is not useful — the code fix means no future delivery can get stuck this way again.
- **Deferred** (documented in the commit): an active cron sweep that proactively re-triggers GitHub's own redelivery API. `webhook_events` doesn't persist the raw payload by design (to avoid growing the exact blob-retention problem #9083 addresses), and the staleness-window fix already makes any future stuck delivery redeliverable well inside GitHub's redelivery window without an active poke.

**#9083** (`src/db/retention.ts`, `src/db/schema.ts`, `migrations/0191_*.sql`)
- Rewrote the batched delete to range over a real, indexable primary key ordered by the retention column instead of `rowid`/`ctid`, and added the matching leading index for every policy table.
- Added retention for the four never-pruned caches (`grounding_file_content_cache` at 2 days — its own read-side TTL is 24h — the other three at 30 days) and three previously-untracked append-only logs (`review_audit`, `decision_records` at 180 days for dispute evidence, `orb_webhook_events`).
- **Deliberately deferred** (documented in the commit): `pull_request_files`, `check_summaries`, `gate_outcomes`, `agent_runs`, `advisories` are current-state tables (upserted per natural key, read as live state), not append-only logs — pruning them needs a read-pattern review this PR doesn't do.

**#9062** (`packages/loopover-engine/src/governor/chokepoint.ts`, `packages/loopover-miner/lib/governor-{chokepoint,state}.ts`)
- A non-floored `cadenceFactor` now scales the per-repo write-rate-limit window instead of denying; the hard deny is reserved for the extreme `reason === "floored"` case only. The miner-lib wrapper advances its bucket against the same scaled policy the decision was evaluated with, so recorded state never disagrees with the verdict it produced.
- `governor_reputation_history` gains a 14-day half-life decay (applied at both read and increment time), so even a floored ratio ages back below the sample-size floor purely from elapsed calendar time, with zero new submissions required.

**#9085** (`src/rules/advisory.ts`, `packages/loopover-engine/src/advisory/gate-advisory.ts`, `src/github/app.ts`, `src/queue/processors.ts`, `src/review/{parity-wire,unified-comment-bridge}.ts`)
- `formatCheckRunOutput`/`buildCheckRunAnnotations` now accept the real blocker codes from the same advisory's `GateCheckEvaluation` (optional, purely additive — every existing caller that omits it keeps byte-identical rendering). When provided, an actual blocker always renders at the most alarming level regardless of its authored severity, and a non-blocking finding is capped at `warning` even if authored `critical`. Wired the one caller that already has the evaluation in scope (the check-run publish in `processors.ts`).
  - **Judgment call**: I did not attempt to make severity load-bearing in the *actual* merge/close decision (`evaluateGateCheckCore`) — that logic is already correctly driven by `isConfiguredGateBlocker`/policy modes, completely independent of the `severity` field, and rewiring severity itself to drive blocking would be a much larger, riskier behavioral change to a production auto-merge/close path than this PR should attempt. The issue's own complaint — "a contributor reading a warning icon on the finding that is about to close their PR is being actively misinformed" — is specifically about the *displayed* icon not matching the *real* consequence, which this fixes without touching the decision path at all.
  - Taxonomy registration for `slop_risk_above_threshold`/`surface_lane_reject` (item 2 of the issue) was already fixed on `main` by #9114 prior to this branch.
- An absent `AdvisoryFinding.confidence` now degrades to `CONFIDENCE_WHEN_UNSTATED` (0.5) instead of 1.0 at the three gate-side consumption sites in `advisory.ts` (and the mirrored site in the engine twin) — matching the "silence is not certainty" fix `CONFIDENCE_WHEN_UNSTATED` already applied at the model-parsing layer.
- Renamed `repo_not_registered` → `repo_not_cached` (mirrors `pr_not_cached`/`issue_not_cached`'s naming for the identical "not yet synced" shape) to resolve the near-miss with `repo_unregistered`, which fires when a repo record exists but isn't in the Gittensor registry snapshot — a plain non-blocking advisory warning, the opposite of `repo_not_cached`'s app-state hold. Renamed consistently across both `src/rules/advisory.ts` and the engine twin (`gate-advisory.ts`), plus their consumers (`parity-wire.ts`, `unified-comment-bridge.ts`) and tests.

## Validation

- `npx tsc --noEmit` clean on the host, `@loopover/engine`, and `@loopover/miner` packages.
- `npm run build --workspace @loopover/engine` clean.
- `npx tsx scripts/check-engine-parity.ts` — ok (7 hand-duplicated pairs agree).
- Full `@loopover/engine` test suite (`npm run test --workspace @loopover/engine`): 739/739 pass.
- `npx vitest run test/contract/engine-parity.test.ts`: 14/14 pass.
- Directly relevant vitest files (rules, unified-comment-bridge, parity-wire, predicted-gate-engine, slop, scenario-summary, webhook, retention, miner-attempt-input-builder, miner-governor-chokepoint, miner-governor-state, github-app): 622/622 pass, including new regression tests for each fix.
- Did not run the full `npm run test:coverage`/`test:ci` gate per the scope of this pass — happy to run it on request.

Closes #9054
Closes #9062
Closes #9083
Closes #9085

### DIFF
--- a/migrations/0191_retention_column_indexes.sql
+++ b/migrations/0191_retention_column_indexes.sql
@@ -1,0 +1,33 @@
+-- #9083: no RETENTION_POLICY (src/db/retention.ts) table had an index whose LEADING column is its
+-- retention timestamp -- every existing index on these tables either doesn't exist at all
+-- (webhook_events, signal_snapshots, score_previews, repo_snapshots) or has the timestamp only in a
+-- trailing position (audit_events, ai_usage_events, product_usage_events,
+-- github_rate_limit_observations, agent_context_snapshots, notification_deliveries). Paired with the
+-- pruneExpiredRecords rewrite (PK-ordered range delete instead of a rowid/ctid semi-join), the inner
+-- SELECT in each batched delete is now an index range scan instead of a full table scan.
+CREATE INDEX IF NOT EXISTS idx_webhook_events_received_at ON webhook_events(received_at);
+CREATE INDEX IF NOT EXISTS idx_audit_events_created_at ON audit_events(created_at);
+CREATE INDEX IF NOT EXISTS idx_ai_usage_events_created_at ON ai_usage_events(created_at);
+CREATE INDEX IF NOT EXISTS idx_product_usage_events_occurred_at ON product_usage_events(occurred_at);
+CREATE INDEX IF NOT EXISTS idx_github_rate_limit_observations_observed_at ON github_rate_limit_observations(observed_at);
+CREATE INDEX IF NOT EXISTS idx_signal_snapshots_generated_at ON signal_snapshots(generated_at);
+CREATE INDEX IF NOT EXISTS idx_score_previews_generated_at ON score_previews(generated_at);
+CREATE INDEX IF NOT EXISTS idx_repo_snapshots_fetched_at ON repo_snapshots(fetched_at);
+CREATE INDEX IF NOT EXISTS idx_agent_context_snapshots_created_at ON agent_context_snapshots(created_at);
+CREATE INDEX IF NOT EXISTS idx_notification_deliveries_created_at ON notification_deliveries(created_at);
+CREATE INDEX IF NOT EXISTS idx_predicted_gate_calls_created_at ON predicted_gate_calls(created_at);
+
+-- #9083: the four never-pruned caches (grounding_file_content_cache, ai_review_cache, ai_slop_cache,
+-- linked_issue_satisfaction_cache) newly gained a retention rule in the same change -- index their
+-- timestamp column too so the new prune sweep stays cheap as they grow.
+CREATE INDEX IF NOT EXISTS idx_grounding_file_content_cache_fetched_at ON grounding_file_content_cache(fetched_at);
+CREATE INDEX IF NOT EXISTS idx_ai_review_cache_created_at ON ai_review_cache(created_at);
+CREATE INDEX IF NOT EXISTS idx_ai_slop_cache_created_at ON ai_slop_cache(created_at);
+CREATE INDEX IF NOT EXISTS idx_linked_issue_satisfaction_cache_created_at ON linked_issue_satisfaction_cache(created_at);
+
+-- #9083: the three newly-retained append-only tables (review_audit, decision_records, orb_webhook_events)
+-- each already had an index touching their timestamp column, but only in a trailing/combined position --
+-- add the leading single-column index the new PK-ordered delete needs.
+CREATE INDEX IF NOT EXISTS idx_review_audit_created_at ON review_audit(created_at);
+CREATE INDEX IF NOT EXISTS idx_decision_records_created_at ON decision_records(created_at);
+CREATE INDEX IF NOT EXISTS idx_orb_webhook_events_received_at ON orb_webhook_events(received_at);
--- a/migrations/0192_purge_stuck_queued_webhook_events.sql
+++ b/migrations/0192_purge_stuck_queued_webhook_events.sql
@@ -1,0 +1,24 @@
+-- #9054: one-off purge of webhook_events rows PERMANENTLY stuck at 'queued'/'superseded' by the dedup-guard
+-- bug fixed alongside this migration (src/github/webhook.ts's enqueueWebhookByEnv). Before that fix, a row
+-- left at 'queued' (the insert happened but WEBHOOKS.send() was lost, or the enqueued job vanished) or
+-- 'superseded' (overwritten by a later coalesced delivery before either was claimed) could NEVER be
+-- redelivered: every GitHub retry and every operator "Redeliver" click carries the identical delivery_id and
+-- (usually) the identical payload hash, hit the dedup guard, and was silently discarded as a no-op
+-- duplicate, forever.
+--
+-- On the live box this incident left 5,547 such rows (4,900 of them check_suite.completed -- the
+-- maybeReReviewOnCiCompletion auto-merge trigger), all from a ~3.5-week cutover-era window that was already
+-- quiescent by the time this was diagnosed. Replaying them is not useful: the code-level fix means any
+-- FUTURE stuck delivery becomes redeliverable after STALE_QUEUED_WEBHOOK_MS (10 minutes) rather than never,
+-- so there is no ongoing gap to backfill, and the PRs behind this historical backlog are long since
+-- resolved one way or another. Simply deleting them (rather than replaying) is what the issue itself called
+-- for, so the stuck-delivery metric is meaningful again going forward.
+--
+-- Applies once, wherever/whenever this migration runs: any 'queued'/'superseded' row still unprocessed a
+-- full day after receipt is dead by definition (real processing completes in well under a second), so this
+-- is safe as a general one-time data-hygiene sweep on any environment, and a no-op on a fresh database with
+-- no history to purge.
+DELETE FROM webhook_events
+ WHERE status IN ('queued', 'superseded')
+   AND processed_at IS NULL
+   AND received_at < datetime('now', '-1 day');
--- a/packages/loopover-engine/src/advisory/gate-advisory.ts
+++ b/packages/loopover-engine/src/advisory/gate-advisory.ts
@@ -39,6 +39,11 @@ function sanitizeForCheckRun(text: string): string {
 }
 
 const DEFAULT_AI_REVIEW_CLOSE_CONFIDENCE = 0.93;
+// #9085 (host-parity): mirrors CONFIDENCE_WHEN_UNSTATED (src/services/ai-review.ts) -- kept as a local constant
+// rather than an import, matching this file's own divergence design (it deliberately never pulls in the host's
+// signals/services subsystem). An ABSENT `finding.confidence` must never read as maximum certainty (1.0):
+// "silence is not certainty" (#8833).
+const CONFIDENCE_WHEN_UNSTATED = 0.5;
 /** Exported to mirror the src twin (#8224): loopover's LOOSENABLE_KNOBS registry anchors the slop knob's
  *  shipped value on this constant (divided by 100 onto the corpus's confidence scale). Value unchanged. */
 export const DEFAULT_SLOP_BLOCK_THRESHOLD = 60;
@@ -196,12 +201,15 @@ export function buildPullRequestAdvisory(
   const targetKey = pr ? `${repoFullName}#${pr.number}` : `${repoFullName}#unknown`;
   const findings: AdvisoryFinding[] = [];
   if (!repo) {
+    // #9085 (host-parity): renamed from `repo_not_registered` -- one character away from, and with the
+    // OPPOSITE gate consequence of, `repo_unregistered` below (addRepoFindings). See the host copy
+    // (src/rules/advisory.ts) for the full rename rationale.
     findings.push({
-      code: "repo_not_registered",
+      code: "repo_not_cached",
       severity: "warning",
-      title: "Repository registration is unknown",
-      detail: "LoopOver cannot evaluate repo-specific rules until registry data is available.",
-      action: "Refresh the Gittensor registry snapshot.",
+      title: "Repository is not yet cached",
+      detail: "LoopOver has not synced this repository yet, so repo-specific rules cannot be evaluated.",
+      action: "Wait for the next repo sync, or re-deliver the installation webhook.",
     });
   } else {
     addRepoFindings(repo, findings);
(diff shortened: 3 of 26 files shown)


JSONbored:
This finding is a false positive: \`webhook_events_status_idx\` isn't new to this PR — it was created back in \`migrations/0001_initial.sql:143\` (\`CREATE INDEX IF NOT EXISTS webhook_events_status_idx ON webhook_events (status);\`). This PR only adds the \`received\` index (\`idx_webhook_events_received_at\`, migration 0191), which the comment right above it in \`schema.ts\` says explicitly ("only the status index above existed before"). Drizzle requires declaring all of a table's indexes in one block, so the reviewer saw both listed together in the same schema.ts hunk and understandably assumed both were new. \`npm run db:schema-drift:check\` confirms schema.ts and migrations/ already agree with no drift.

PR #9332, opened 2026-07-27
fix(engine): compare maxIterations and percentComplete in progressChanged

### LINKED ISSUE #9323: engine: progressChanged ignores maxIterations/percentComplete changes in the loop-progress snapshot diff

## Context

`progressChanged` in `packages/loopover-engine/src/loop-progress.ts` (~line 79) decides whether a
new `ProgressSnapshot` differs from the previous one "in a way worth pushing to the customer" —
its own doc comment says it "Compares the displayed axes: phase, status, iteration, and the
activity tail's contents," per #4800's on-change-not-polling design.

`ProgressSnapshot` (defined ~line 28) has two more displayed fields the function never compares:
`maxIterations` and `percentComplete` (computed from `iteration / maxIterations`, ~lines 45-47).

Concretely: if a caller's `maxIterations` changes between two snapshots while `iteration`,
`phase`, `status`, and `recentActivity` all stay the same (e.g. an operator raises the iteration
budget mid-run), `percentComplete` silently changes value but `progressChanged` still returns
`false` — so the customer-facing streaming surface that consumes this (`buildProgressSnapshot`'s
caller in `src/api/routes.ts`) never pushes the update, and the progress bar shown to the customer
goes stale even though a real, displayed number changed.

`test/unit/loop-progress.test.ts` has no test exercising a `maxIterations`-only change between two
snapshots.

## Requirements

- `progressChanged` must also return `true` when `prev.maxIterations !== next.maxIterations` or
  `prev.percentComplete !== next.percentComplete`, added to its existing OR-chain of comparisons
  alongside `phase`/`status`/`iteration`/`activityChanged(...)`.
- Do not change `activityChanged` or the other existing comparisons.
- Do not change how `percentComplete`/`maxIterations` are computed in `buildProgressSnapshot` —
  only the diffing function.

## Deliverables

- [ ] `progressChanged` in `packages/loopover-engine/src/loop-progress.ts` compares
      `maxIterations` and `percentComplete` in addition to its existing four axes.
- [ ] A new test case in `test/unit/loop-progress.test.ts` asserting `progressChanged` returns
      `true` when only `maxIterations` (and therefore `percentComplete`) differs between two
      snapshots, with `phase`/`status`/`iteration`/`recentActivity` held identical.
- [ ] A second new test case asserting `progressChanged` still returns `false` when all six axes
      (including the two new ones) are identical between two snapshots — a regression guard
      against over-triggering.

All three deliverables are required in this single PR.

## Test Coverage Requirements

This repo enforces 99%+ Codecov patch coverage, branch-counted, on every changed line/branch in
`src/**`/`workers/**`. (Note: `packages/loopover-engine/src/**` is graded under this same gate —
confirm via the repo's Codecov config if adding a new branch here; the two new test cases above
must exercise both the true and false paths of the added condition.)

## Expected Outcome

A `maxIterations` change between two progress snapshots (and the resulting `percentComplete`
change) is now correctly detected by `progressChanged`, so the customer-facing streaming surface
pushes the update instead of leaving a stale progress bar.

## Links & Resources

- `packages/loopover-engine/src/loop-progress.ts` (`progressChanged`, ~line 79; `ProgressSnapshot`
  type, ~line 28)
- `src/api/routes.ts` (consumer of `buildProgressSnapshot`/`progressChanged`, for context)
- `test/unit/loop-progress.test.ts` (existing test file to extend)

philluiz2323:
﻿## Problem

Closes #9323.

`progressChanged` compared phase, status, iteration, and the activity tail — but not `maxIterations` or `percentComplete`. When an operator raised the iteration budget mid-run while those other axes stayed the same, the customer-facing progress bar went stale because no push was emitted.

## Fix

Include `maxIterations` and `percentComplete` in the `progressChanged` comparison, matching the displayed snapshot axes.

## Tests

- Extend the existing change-detection test to cover `maxIterations`.
- Regression: raising `maxIterations` from 5→10 at the same iteration pushes even when activity is unchanged.

Reverting either comparison fails the regression test. `git diff --check` clean.

## Validation

- [x] `vitest run test/unit/loop-progress.test.ts` (14 pass)
- Engine progress-only; unrelated UI/MCP checks not exercised.

## Safety

- [x] No secrets exposed; no UI changes (N/A evidence)

### DIFF
--- a/packages/loopover-engine/src/loop-progress.ts
+++ b/packages/loopover-engine/src/loop-progress.ts
@@ -78,13 +78,16 @@ function activityChanged(prev: readonly LoopProgressActivity[], next: readonly L
 
 /** True when `next` differs from `prev` in a way worth pushing to the customer — so the surface streams
  *  ON CHANGE instead of polling on a fixed interval (#4800's acceptance). A null `prev` (the first snapshot)
- *  always pushes. Compares the displayed axes: phase, status, iteration, and the activity tail's contents. */
+ *  always pushes. Compares the displayed axes: phase, status, iteration, maxIterations, percentComplete,
+ *  and the activity tail's contents. */
 export function progressChanged(prev: ProgressSnapshot | null, next: ProgressSnapshot): boolean {
   if (prev === null) return true;
   return (
     prev.phase !== next.phase ||
     prev.status !== next.status ||
     prev.iteration !== next.iteration ||
+    prev.maxIterations !== next.maxIterations ||
+    prev.percentComplete !== next.percentComplete ||
     activityChanged(prev.recentActivity, next.recentActivity)
   );
 }
--- a/test/unit/loop-progress.test.ts
+++ b/test/unit/loop-progress.test.ts
@@ -56,11 +56,20 @@ describe("progressChanged — push on change, not on a fixed interval (#4800)", 
     expect(progressChanged(null, base)).toBe(true);
   });
 
-  it("pushes when phase, status, iteration, or the activity tail changes", () => {
+  it("pushes when phase, status, iteration, maxIterations, percentComplete, or the activity tail changes", () => {
     expect(progressChanged(base, buildProgressSnapshot(running({ phase: "reviewing", recentActivity: [{ step: "a" }] })))).toBe(true);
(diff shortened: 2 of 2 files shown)


philluiz2323:
Closing: engine patch coverage requires packages/loopover-engine tests; re-scoping.

Issue #9273, opened 2026-07-27
ledger: git-commit anchoring backend, cross-verified against GH Archive / Software Heritage


JSONbored:
### Context

Sub-issue of #9267 (external decision-ledger anchoring). Secondary, complementary anchoring backend: a git commit to a public repo, appending each checkpoint's anchor payload. Alone this is weaker than Rekor (GitHub is a trusted third party; `git push --force` rewrites it) — but per the research on #9267, it becomes genuinely strong combined with mirrors nobody at LoopOver controls: GH Archive (hourly `PushEvent` records) and Software Heritage's "Save Code Now" (on-demand public-repo archival). A rewrite becomes independently detectable.

### Requirements

1. **Append, never force-push or rewrite.** Each checkpoint appends one line/entry to `anchors.jsonl` (or one commit) in a public repo, via the Contents API using the existing GitHub App installation token (`@octokit/core` already a dependency — no new one).
2. **The committed content is the same canonicalized anchor payload as #9270**, plus the signature — identical to what gets submitted to Rekor, so the two backends anchor the same fact.
3. **On any error, write a `status: 'failed'` row via #9271's persistence**, same posture as the Rekor backend — this must not throw past the caller.
4. **Document the independent-mirror verification step**: cross-checking a specific push against `data.gharchive.org`'s hourly export, so a verifier isn't solely trusting the anchors repo's own git history.

### Deliverables

- Git-commit backend implemented, persisted via #9271
- Tests: successful commit is recorded with its SHA; a commit failure (rate limit, auth) records a failure row and does not throw past the caller
- Documented cross-mirror verification procedure (GH Archive query) in this issue's own PR

### Expected outcome

A second, independent, freely-verifiable anchor for the same checkpoint, whose tampering is detectable by anyone checking an archive LoopOver does not control.

### References

Sub-issue of #9267. Depends on #9270 (payload), #9271 (persistence). Full mechanism research: decision-record comment on #9267.

JSONbored:
Landed on main via #9397's squash-merge (originally #9403, merged into the intermediate `feat/ledger-anchor-rekor` branch, so the auto-close keyword never fired against the default branch). Verified present on `main`: `submitToGitAnchor`, `buildAnchorLogLine`, `GitHubContentsRequester`.

Issue #9272, opened 2026-07-27
ledger: Rekor v2 hashedrekord anchoring backend


JSONbored:
### Context

Sub-issue of #9267 (external decision-ledger anchoring). Primary anchoring backend, per the research on #9267: Rekor v2's `hashedrekord` accepts a plain hash + signature — content-agnostic, Rekor never sees the payload itself — and needs zero new dependencies since ECDSA P-256/SHA-256 is native to Workers via `crypto.subtle`.

### Requirements

1. **Submit `POST /api/v2/log/entries`** with a `hashedRekordRequestV002`: base64 SHA-256 digest of the canonicalized anchor payload (#9270), signature over the payload bytes, and the verifier public key (SPKI DER) + `keyDetails: PKIX_ECDSA_P256_SHA_256`. No Fulcio, no OIDC — the self-managed key from #9270 is what `verifier.publicKey` expects.
2. **Do not hardcode a log URL.** Shards rotate annually (`log2025-1`, `log2026-1`, ...); resolve the current shard at call time or via a small pinned config, not a literal constant that goes stale.
3. **Store the full response**: `logIndex`, `logId.keyId`, `kindVersion`, `inclusionProof` (hashes + signed checkpoint), `canonicalizedBody` — this is what makes verification possible fully offline later, without trusting Rekor's continued availability.
4. **Timeout ≥ 20s** — v2 batches submissions; a short timeout will falsely read as failure.
5. **On any error, write a `status: 'failed'` row via #9271's persistence** — this backend must never throw in a way that blocks the anchoring job from also trying #9272.

### Deliverables

- Rekor v2 submission implemented against the anchor payload from #9270, persisted via #9271
- Tests: successful submission stores every field needed for later offline verification; a submission failure (timeout, non-2xx) records a failure row and does not throw past the caller
- Documented offline-verification procedure (checkpoint signature against the sigstore trust root, then Merkle inclusion, then digest match) in this issue's own PR

### Expected outcome

Every checkpointed anchor carries a Rekor entry independently verifiable by anyone, including fully offline, without contacting LoopOver.

### References

Sub-issue of #9267. Depends on #9270 (payload), #9271 (persistence). Full mechanism research + example request/response shapes: decision-record comment on #9267.

Issue #9270, opened 2026-07-27
ledger: versioned + signed anchor payload, public key + rotation route


JSONbored:
### Context

Sub-issue of #9267 (external decision-ledger anchoring). `buildLedgerAnchorPayload` (`src/review/decision-record.ts:376`) returns `{ seq, rowHash, at }` and is unused. Per the research on #9267, this shape cannot survive being read by a stranger in two years — it names neither the ledger it's from nor a schema version.

### Requirements

1. **Extend the payload to a self-describing, versioned shape**: `{ v: 1, ledger: "loopover.decision_ledger", seq, rowHash, totalCount, at }`.
2. **Canonicalize and sign**: reuse the existing `canonicalJson` (`decision-record.ts`) over the payload, sign with **ECDSA P-256 via Workers' native `crypto.subtle`** — no new dependency.
3. **Key management**: generate a signing keypair, store the private key as a Worker secret, and **publish the public key (SPKI, DER)** at a stable public route with an explicit rotation history (a verifier two years from now must be able to find the key that was valid when a given anchor was made).
4. **`GET /v1/public/decision-ledger/anchor-key`** (or similar) serving the current + any historical public keys, each with its validity window.

### Deliverables

- Versioned/signed payload builder replacing the current `buildLedgerAnchorPayload`
- Key generation + storage procedure (documented, since this is operational, not just code)
- Public key route + tests, including a rotation scenario (old anchors still verify against the historical key)

### Expected outcome

Every anchor is independently verifiable as genuinely from this ledger, signed by a key a verifier can locate even after key rotation.

### References

Sub-issue of #9267. Depends on #9269 conceptually (same trust surface) but can be built in parallel. `buildLedgerAnchorPayload`: `src/review/decision-record.ts:376`. `canonicalJson`: same file, `:51`.

Issue #9267, opened 2026-07-27
orb(trust): external decision-ledger anchoring — Rekor + git-commit, from tamper-evident to tamper-proof


JSONbored:
### Context

`migrations/0180_decision_ledger.sql` (#8843) shipped a hash-chained, append-only decision ledger and named its own honest limit up front: **tamper-evident, not tamper-proof** — an operator can still delete the chain wholesale and re-chain from genesis undetected, because nothing outside the operator's own database holds an independent record of any past tip. `buildLedgerAnchorPayload` (`src/review/decision-record.ts:376`) was written for exactly this and has been called from nowhere since. #9186's verification matrix and #9215's eval-interface spec both now state this gap in public-facing terms, which is what surfaced it as blocking.

### Decision: which anchoring mechanisms to use, and why (research complete — see linked comment)

Full mechanism research (Rekor v2, CT/tlog-tiles + witness networks, Bittensor on-chain commitment, OpenTimestamps, git-commit anchoring) is posted as a decision-record comment on this issue. Summary:

- **Ship Rekor v2 (`hashedrekord`) + git-commit anchoring together, as complementary primary mechanisms.** Both are free, need no new runtime dependencies (Rekor: one `fetch` + `crypto.subtle` ECDSA P-256, already native to Workers; git: reuses the existing GitHub App auth), and both are independently offline-verifiable or third-party-mirrored (GH Archive, Software Heritage).
- **Add OpenTimestamps as a backstop**, not a primary — free, needs zero trusted party at verification time, but hours of latency and the weakest verifier UX (needs Bitcoin block data).
- **Defer tlog-tiles + witness-network anchoring** — the strongest possible guarantee (genuine non-equivocation, defeats split-view/rewind) but requires either a Go sequencer or a hand-rolled implementation. Revisit exactly when `migrations/0180`'s own stated trigger fires: when tenants exist.
[...]

PR #9395, opened 2026-07-27
feat(ledger): anchor persistence + public attempt log, success and failure alike

### LINKED ISSUE #9271: ledger: anchor persistence + public attempt log (success and failure both visible)
### Context

Sub-issue of #9267 (external decision-ledger anchoring). Every anchor backend (#9272 Rekor, #9273 git) needs somewhere to record what it did — including when it failed. Per the research on #9267: **a failed anchor attempt must be recorded and exposed publicly**, or an operator could make every anchor silently "fail" and quietly regress back to tamper-evident-only with no visible signal.

### Requirements

1. **New migration**: `decision_ledger_anchors` — `seq, row_hash, payload_json, signature, backend ('rekor'|'git'|'ots'), backend_ref, proof_r2_key, status ('ok'|'failed'), error, created_at`.
2. **`backend_ref` for Rekor must store the full resolvable reference**: `{ shardBaseUrl, logIndex, logIdKeyId, uuid }` — a bare `logIndex` becomes unresolvable after the annual shard rotation.
3. **Failures are rows, not silence.** A backend call that errors writes a `status: 'failed'` row with the error, on the same schedule as a success would have.
4. **Public read surface**: `GET /v1/public/decision-ledger/anchors` (paginated) exposing every attempt, success or failure, so "anchoring has been failing for a week" is a fact anyone can observe, not something only the operator's own logs show.

### Deliverables

- Migration + drift-check compliance (contiguous `NNNN_*.sql`, per house convention)
- Persistence helper used by both backend issues
- Public listing endpoint + OpenAPI entry
- Tests: success row, failure row, pagination, and that a failure is queryable identically to a success (no special-casing that could hide it)

### Expected outcome

Anchoring's own health is itself a publicly verifiable fact, closing the "silent regression" hole the research identified.

### References

Sub-issue of #9267. Feeds #9272 (Rekor backend), #9273 (git-commit backend). Depends on #9270 (payload shape) for `payload_json`'s contents.

JSONbored:
## Summary
Third sub-issue of #9267 (external ledger anchoring), stacked on #9270 (#9392). `migrations/0195` adds `decision_ledger_anchors`: every backend (#9272 Rekor, #9273 git) calls the **same** `recordLedgerAnchorAttempt` whether it succeeded or failed — there is no separate failure path a backend could simply not call.

**Why this matters, per the mechanism research on #9267**: anchoring is best-effort — it must never block a review — which means a failure could otherwise be silent. An operator whose anchoring quietly fails could regress the ledger back to tamper-evident-only with no visible signal. Recording every attempt on the same public listing, in the same shape regardless of outcome, is what makes anchoring's own health a fact anyone can observe, not something only the operator's own logs show.

`GET /v1/public/decision-ledger/anchors` serves the log, paginated newest-first (`?before=` cursor, `?limit=`), filterable by `?backend=`. `backend_ref` stores the **full** resolvable reference — for Rekor, shard base URL + log index + key id + uuid, not a bare index that becomes unresolvable after the annual shard rotation.

One bug caught and fixed during testing: `recordLedgerAnchorAttempt` originally called `nowIso()` internally with no way to override it, so several inserts in a tight loop could land on the same millisecond and sort arbitrarily. Made `createdAt` injectable, matching `loadPublicRulePrecision`'s existing pattern for exactly this class of testability.

Closes #9271.

## Test plan
- [x] `npx vitest run test/unit/ledger-anchor-persistence.test.ts` — 8/8, **100% statement/branch/function/line coverage**
- [x] **The load-bearing test**: a failed attempt is served on the public listing with the exact same field set as a success — no special-cased shape that could hide it
- [x] A failed attempt is queryable identically to a success when filtering by backend
- [x] Pagination proven end-to-end: cursor advances the page, terminates with `nextBefore: null`
- [x] `limit` clamped to `[1, 200]`; malformed `backend_ref` degrades to `null` rather than throwing
- [x] `npx vitest run test/integration/public-ledger-anchors-route.test.ts` — 6/6: no-auth-required, failure served on the public route, `?backend=`/`?before=`/`?limit=` query handling, invalid backend value falls back to unfiltered rather than erroring
- [x] Added to the shared exempt-public-routes table
- [x] `npm run db:migrations:check` — 199 migrations OK, contiguous
- [x] `npx tsc --noEmit` clean (including an `exactOptionalPropertyTypes` fix in the route handler) · `npm run ui:openapi:check` clean

### DIFF
--- a/apps/loopover-ui/public/openapi.json
+++ b/apps/loopover-ui/public/openapi.json
@@ -434,6 +434,82 @@
               "merged"
             ]
           },
+          "rulePrecision": {
+            "type": "object",
+            "properties": {
+              "windowDays": {
+                "type": "number"
+              },
+              "rules": {
+                "type": "array",
+                "items": {
+                  "type": "object",
+                  "properties": {
+                    "ruleId": {
+                      "type": "string"
+                    },
+                    "decided": {
+                      "type": "number"
+                    },
+                    "confirmed": {
+                      "type": "number"
+                    },
+                    "precision": {
+                      "type": "number",
+                      "nullable": true
+                    }
+                  },
+                  "required": [
+                    "ruleId",
+                    "decided",
+                    "confirmed",
+                    "precision"
+                  ]
+                }
+              },
+              "reversals": {
+                "type": "object",
+                "properties": {
+                  "reopened": {
+                    "type": "number"
+                  },
+                  "reverted": {
+                    "type": "number"
+                  },
+                  "superseded": {
+                    "type": "number"
+                  }
+                },
+                "required": [
+                  "reopened",
+                  "reverted",
+                  "superseded"
+                ]
+              },
+              "latestBacktestRun": {
+                "type": "object",
+                "nullable": true,
+                "properties": {
+                  "corpusChecksum": {
+                    "type": "string"
+                  },
+                  "at": {
+                    "type": "string"
+                  }
+                },
+                "required": [
+                  "corpusChecksum",
+                  "at"
+                ]
+              }
+            },
+            "required": [
+              "windowDays",
+              "rules",
+              "reversals",
+              "latestBacktestRun"
+            ]
+          },
           "byProject": {
             "type": "array",
             "items": {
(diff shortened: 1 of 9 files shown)


JSONbored:
Stacked on #9392 (#9270) — the base branch is `feat/ledger-anchor-payload`, not `main`, since this needs the `LedgerAnchorPayload`/`SignedLedgerAnchor` types from that PR. Diff shown here should only include this PR's own 6 files once GitHub picks up the new base. Merge order: #9392 first, then this.

Issue #9464, opened 2026-07-27
gate(visual): browserless render failure is swallowed, not deferred → screenshot gate false-closes legitimate PRs (#9030/#9207 unpatched sibling)


JSONbored:
Parent: #9461

## Summary

#9030 and #9207 taught the screenshot-table gate to defer its close when the bot's own capture pipeline blips — but only for failure classes that either report `previewPending` or **throw**. A browserless render failure does **neither**: it is swallowed per-shot and `buildCapture` returns normally. The gate then reads "no visual evidence, no retry pending" and **closes a legitimate visual PR one-shot** — precisely the false-close class #9030 exists to prevent, arriving through an unpatched door.

## Mechanism (verified at HEAD, 776c4147a)

`src/review/visual/capture.ts:408`:
```ts
const { png, authWalled } = await captureShot(env, page, viewport, ...).catch(() => ({ png: null, authWalled: false }));
```

The failure is swallowed per route/viewport and substituted with a non-persisted on-demand URL. Consequently:

- `hasSuccessfulBotCapture` requires **persisted** pairs (`src/review/visual/capture.ts:112-128`) ⇒ false.
- `buildCapture` returns normally with `previewPending: false` and throws nothing.
- The #9030 retry is scheduled only on `previewPending` or a **thrown** error (`src/queue/processors.ts:12422-12480`) ⇒ never fires, so `visualCaptureRetryPendingSha` is never set.
- The next maintenance pass reads "not satisfied, no retry pending" (`processors.ts:3449-3453`) ⇒ `screenshotTableMatch` matched ⇒ **close**.

Trigger: the preview URL resolves fine (the deploy succeeded), but browserless is down, saturated, or times out per shot. The author relied on bot capture rather than pasting a hand-built table — which is the intended workflow.

## Why the existing guards miss it

#9207 correctly classified "preview still building" and "capture threw" as blips. This third class — *preview fine, renderer produced nothing, no exception* — was not modelled, because the `.catch()` at `capture.ts:408` converts it into an apparently-successful capture with empty output.

## Compounding factor: nothing would alert on it

[...]

Issue #9030, opened 2026-07-26
orb(gate): screenshot-table gate closes a legitimate visual PR when the capture service blips


JSONbored:
Visual capture runs in-pass (`src/queue/processors.ts` ~11324-11400). On failure it yields `beforeAfter=[]` and never calls `markPullRequestVisualCaptureSatisfied` (~11374-11385). The same webhook's maintenance pass then reads `botCaptureSatisfied` as false (~3103-3123) and, with `action: "close"`, issues a **one-shot close**.

There is no grace pass and no distinction between "capture concluded: this PR has no visual routes" and "the capture pipeline errored" (browserless down, timeout, transient network). A contributor's correct visual PR is closed because an internal service blipped — and under one-shot rules their recovery is to open a whole new PR.

Given `missing before/after screenshot table` was the single most common close reason in the 2026-07-26 ledger, any false-positive rate here is expensive.

## Fix
On capture **error** (as opposed to a genuine no-visual-routes result) hold instead of close, or defer the gate one pass with a persisted attempt marker. Only close when the capture pipeline actually succeeded and found the table missing.

Issue #4110, opened 2026-07-08
feat(review): wire bot-generated visual captures into screenshotTableGate


JSONbored:
Part of #3607 (visual-capture convergence epic).

## Context

`src/review/screenshot-table-gate.ts`'s `evaluateScreenshotTableGate` (as of 2026-07-07: `input: { config, prBody, prLabels, changedFiles }`, no bot-capture input) is a deterministic, presence-only check on the PR BODY text — it has no way to be satisfied by an automated bot-rendered before/after capture, even when the visual-capture pipeline (`src/review/visual/capture.ts`/`shot.ts`) successfully produced one. A bot-rendered before/after currently CANNOT satisfy this gate; only a hand-authored markdown table in the PR body can.

Separately: `ScreenshotTableGateConfig`'s `action: "request_changes" | "comment"` values are fully typed/validated but have ZERO runtime effect — `src/queue/processors.ts` (~line 2787-2795 as of 2026-07-07) is the only reader of the gate's result and only branches on `=== "close"`. This is dead config surface that should either be wired up for real or removed.

## Requirements

- Add a `source: "bot" | "contributor"` (or similar) input to `evaluateScreenshotTableGate` so a successful automated capture (from the visual-capture pipeline, when it ran for this PR) can satisfy the gate instead of requiring a hand-authored table.
- Wire this through `src/settings/agent-actions.ts`'s `screenshotTableMatch` handling (~line 668-689 as of 2026-07-07) so a successful bot capture doesn't sit as a purely decorative comment section alongside a gate that still demands manual evidence.
- Resolve the dead `request_changes`/`comment` actions: either give them real runtime effect in `processors.ts`, or remove them from `ScreenshotTableGateConfig` as unused surface — do not leave them silently inert either way.
- Consider whether `src/signals/focus-manifest.ts`'s `pre_merge_checks`/`manifestPolicy` mechanism could serve as a cheaper interim "screenshot present" requirement ahead of full gate integration, if that turns out simpler than extending the gate directly.

## Deliverables
- [ ] `evaluateScreenshotTableGate` accepts and correctly evaluates a bot-capture-result input
- [ ] `agent-actions.ts` wiring updated
- [ ] `request_changes`/`comment` either wired to real effect or removed, with the decision documented
- [ ] Tests: bot capture satisfies the gate; contributor-authored table still satisfies it; absence of both still triggers the configured action; both branches of every new predicate

## Expected outcome

[...]

PR #9501, opened 2026-07-27
fix(gate): stop two subsystem failures from becoming wrong auto-verdicts (#9460, #9462)

### LINKED ISSUE #9460: gate: AI blocker silently collapses to a clean pass when its text trips the public-safe filter (false merge, live config)
## Summary

In the `single` combine strategy — **the strategy the live deployment uses** — a reviewer that names a real blocker produces a **clean pass** whenever the blocker's text happens to contain an ordinary word the public-safe filter rejects (`score`, `ranking`, `reward`, `cohort`, `payout`, `farming`, `reviewability`). The PR auto-merges with the defect unreported.

## Mechanism (verified at HEAD, 776c4147a)

`synthesizeDefect` drops the entire defect when the blocker title is not publishable:

`src/services/ai-review.ts:2561-2577`
```ts
const title = toPublicSafe(primary);
if (!title) return null; // unsafe title → drop the block entirely (fail-safe)
```

`combineReviews`, `single` branch — `src/services/ai-review.ts:2598-2608`:
```ts
const r = present[0];
if (!r) return { defect: null, split: false, inconclusive: true };
return {
  defect: r.blockers.length > 0 ? synthesizeDefect([r]) : null,
  split: false,
  inconclusive: false,
};
```

When `synthesizeDefect` returns `null`, the trio is `{defect: null, split: false, inconclusive: false}` — **indistinguishable from "the reviewer found nothing"**. `runAiReviewForAdvisory` then pushes no finding at all (`src/queue/ai-review-orchestration.ts:897-988`), notes are composed from the surviving sentences, and the row is written **durably cacheable** (`:1032-1042`, no `cacheable` field ⇒ true) — so the wrong verdict is replayed on later passes.

The comment calls this "fail-safe". In the `consensus` branch it is; in `single`/`synthesis` it is **fail-open**.

## Why the existing guards do not cover it

- The `consensus` branch **is** fail-closed for exactly this case: `src/services/ai-review.ts:2642` sets `split = !defect && realBlockers > 0`, converting a dropped defect into a hold. `single` and `synthesis` have no equivalent.
- `composeAdvisoryNotes` passes `allowBareScoreTerm` (`:2029-2035`), but `synthesizeDefect` and `consensusDefectOf` call `toPublicSafe` with **no options** (`:2242`, `:2573`) — so even a repo on the score-terms allowlist gets strict filtering on the one string that carries the gate decision.

## Why this is live, and why it got worse two commits ago

- `AI_PROVIDER=claude-code,ollama` ⇒ one reviewer slot, `combine: "single"` (`src/selfhost/ai.ts:1596-1613`). The live path is the vulnerable one.
- The rejection set is ordinary review vocabulary matched with `.includes()`, plus `/\bscore\w*\b/i` unless the repo is in `LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS`. In a codebase whose domain is scoring, ranking and rewards, blocker titles like *"`computeScore` divides by zero when `items` is empty"* or *"the ranking comparator drops the tie-break"* are routine.
- **#9445 (`6269d27a9`, two commits before this audit) widened reachability.** Before per-sentence filtering, a filtered review often lost its whole narrative, leaving no public notes and routing the PR to an `ai_review_inconclusive` hold. Now the narrative survives, notes exist, and the pass completes cleanly with no defect. That commit's own data — the filter engaging on roughly 40% of reviews on this repo — is the base rate for how often blocker text carries these words.

## Requirements

1. A blocker that exists but cannot be published must **never** resolve to "no defect".
2. The fix must apply to **both** `single` and `synthesis` (both call `synthesizeDefect` and both currently collapse); `consensus` already handles it and must not change behaviour.
3. The unpublishable case must be observable, not silent.

## Deliverables

- [ ] In `combineReviews`, when `blockers.length > 0 && synthesizeDefect(...) === null`, return a hold rather than a pass — `{defect: null, split: true, splitConfidence: <flagging reviewer's confidence>}` (reusing the existing split→hold path) or `inconclusive: true`. Pick one and state the rationale in the code comment.
- [ ] Same treatment in the `synthesis` branch, both the `either` and `both` arms.
- [ ] A counter (e.g. `loopover_ai_review_unpublishable_blocker_total`) incremented at the drop site, so the rate is measurable.
- [ ] Correct the `synthesizeDefect` doc comment: it is fail-safe only for `consensus`.
- [ ] Consider whether the gate-bearing title should be filtered with `allowBareScoreTerm` for allowlisted repos, matching `composeAdvisoryNotes`. If not, document why the two differ.

## Tests (must fail against current main)

- [ ] `single` + one reviewer + one blocker whose title contains `score` (repo NOT allowlisted) ⇒ result holds; asserts `defect === null && (split || inconclusive)`.
- [ ] Same for each of `reward`, `ranking`, `cohort`, `farming`, `payout`.
- [ ] `synthesis`/`either` and `synthesis`/`both` equivalents.
- [ ] `consensus` unchanged (regression guard).
- [ ] A publishable blocker still produces a defect in all three strategies.
- [ ] The cache row written for a dropped-blocker review is **not** durably cacheable, or is keyed so the hold is reproduced.

## Expected outcome

A reviewer-named blocker always produces either a published defect or an explicit hold. It can never produce a silent pass. The rate of unpublishable blockers becomes visible in metrics rather than invisible in merges.

## Verification

Reproduce on a scratch PR by configuring a repo off the score-terms allowlist and inducing a blocker whose title contains "score"; confirm the PR holds instead of merging.

### LINKED ISSUE #9462: gate(visual): capture-retry marker never cleared + pending marker skips instead of holds → screenshot gate silently disabled (false merge)
Parent: #9461

## Summary

The screenshot-table gate's close-deferral marker (`visualCaptureRetryPendingSha`, added by #9030) is **never cleared when the capture retry budget is exhausted**, and a head carrying a stale marker causes the gate to be **skipped entirely rather than held**. Combined, a PR whose preview never renders can auto-merge in violation of a gate that is configured to `close`.

This is two defects that compound; both need fixing.

## Part A — the marker is never cleared on exhaustion

`src/queue/processors.ts:9469-9491` (`scheduleVisualCaptureRetry`):
```ts
if (args.previewPollAttempt >= MAX_PREVIEW_POLL_ATTEMPTS) return;
if (args.pr.headSha) {
  await markPullRequestVisualCaptureRetryPending(env, args.repoFullName, args.pr.number, args.pr.headSha)...
}
```

The doc comment at `:9464-9467` claims:

> once `MAX_PREVIEW_POLL_ATTEMPTS` is reached, the marker is deliberately left unset so the gate falls through to its normal (accurate) evaluation on this final attempt

**This is false in code.** The early return only skips setting the marker *again*; the marker written by the *previous* attempt is still there. Nothing clears it.

Exhaustive verification — grep for every reader/writer of the column across `src/`:
- **One** clear: `src/db/repositories.ts:4489` — `markPullRequestVisualCaptureSatisfied` sets `visualCaptureRetryPendingSha: null`, and it runs **only on a successful capture**.
- **One** setter: `src/db/repositories.ts:4504`.
- No other writer anywhere.

So if capture never succeeds for a head, `pr.visualCaptureRetryPendingSha === pr.headSha` holds **permanently** for that head. The same permanent state arises from a lost `recapture-preview` job (crash between the marker write at `:9481` and the enqueue at `:9492`, or a dropped queue message) — with zero attempts actually spent.

## Part B — a pending marker skips the gate instead of holding

`src/queue/processors.ts:3449-3453`:
```ts
const botCaptureRetryPending = Boolean(pr.headSha) && pr.visualCaptureRetryPendingSha === pr.headSha;
const screenshotTableMatch =
  screenshotTableGateResult.violated && screenshotTableGateConfig.action === "close" && !botCaptureRetryPending
    ? { matched: true, reason: screenshotTableGateResult.reason }
    : undefined;
```

In `close` mode the planner's `screenshotTableMatch` close is the gate's **only** enforcement (`src/settings/agent-actions.ts:977-992` is the sole consumer). When `botCaptureRetryPending` is true, `screenshotTableMatch` is left `undefined` and the plan falls through to ordinary merit/CI evaluation — **a green PR plans a merge**, with no finding, no hold, and no blocker representing "screenshot evidence unresolved".

Contrast the correct precedent in the same codebase: an unresolvable enforced pre-merge check yields `PRE_MERGE_CHECK_UNRESOLVED_CODE` → NEUTRAL → held, documented as "never silently skipping a hard requirement (auto-merge bypass)" (`packages/loopover-engine/src/review/pre-merge-checks.ts:9-13,38-49`). Same "cannot evaluate the evidence" state, opposite resolution — a disposition inversion.

## Why this is live

Verified on `edge-nl-01` (2026-07-27): `screenshotTableGate.action: close` is set in the private per-repo config for **all three** repos (`/opt/loopover/loopover-config/jsonbored__{loopover,metagraphed,awesome-claude}/.loopover.yml`), and the gate was re-enabled two commits before this audit in 73d4e9420 (#9455).

Note this defect makes a PR **merge**; its sibling (#9464) makes a PR **close**. They are different failure classes of the same subsystem.

## Requirements

1. An exhausted retry budget must restore accurate gate evaluation, exactly as the doc comment already promises.
2. A gate that is violated and cannot be evaluated must **hold**, never silently skip.
3. A lost retry job must not be able to neutralise the gate indefinitely.

## Deliverables

- [ ] Clear `visualCaptureRetryPendingSha` in the exhausted branch of `scheduleVisualCaptureRetry` (or wherever `previewFailed` resolves on the final attempt), so the marker cannot outlive the retry budget.
- [ ] Belt-and-braces: bound the marker by age at the read site (`processors.ts:3449`) so a marker that survives any future code path cannot neutralise the gate forever. This needs a timestamp — either add a column alongside the SHA or probe the corresponding audit event.
- [ ] While `botCaptureRetryPending && violated && action === "close"`, emit a **neutral hold** (a finding mirroring `pre_merge_check_unresolved`) instead of leaving `screenshotTableMatch` undefined. The deferral must suppress the close *and* the merge.
- [ ] Correct the two doc comments (`processors.ts:9464-9467` and `:3444-3448`) — both currently assert a self-healing property the code does not have.

## Tests (must fail against current main)

- [ ] Capture fails for all `MAX_PREVIEW_POLL_ATTEMPTS`; assert `visualCaptureRetryPendingSha` is null afterwards.
- [ ] A PR with a stale marker for the current head, gate violated, `action: close` ⇒ plan contains **neither** merge **nor** close, and carries an explicit unresolved-evidence finding.
- [ ] Marker set but retry job never runs (simulate drop) ⇒ gate still enforces after the age bound elapses.
- [ ] Happy path unchanged: successful capture clears the marker and the gate evaluates normally.

## Expected outcome

The screenshot-table gate can be deferred by a transient capture failure, but never disabled by one. A PR that violates the gate either closes (when evidence is genuinely absent) or holds (when evidence cannot be determined) — it never merges.

JSONbored:
## Summary

Two verified paths by which a **subsystem failure became a verdict** instead of a "cannot determine". Both are from the 2026-07-27 adversarial audit, and both are live on this deployment's current configuration.

The codebase already models the correct shape in `pre-merge-checks.ts:9-13,38-49` — an unresolvable enforced check yields `PRE_MERGE_CHECK_UNRESOLVED_CODE` → NEUTRAL → held, documented explicitly as "never silently skipping a hard requirement (auto-merge bypass)". Neither of these two followed it.

Closes #9460
Closes #9462

### #9460 — an AI blocker silently collapsed to a clean pass (false merge)

`synthesizeDefect` returns `null` for two situations that must not share an outcome: nobody named a real blocker (a genuine clean pass), and a real blocker whose title `toPublicSafe` refuses to publish. The `single` and `synthesis` strategies collapsed both to `{defect: null, split: false, inconclusive: false}` — indistinguishable from "the reviewer found nothing" — so the PR auto-merged with the defect unreported, and the row was cached durably.

`single` is **this deployment's live strategy** (`AI_PROVIDER=claude-code,ollama` → `resolveAiReviewerPlan`, `src/selfhost/ai.ts:1596-1613`). `consensus` already failed closed via its own `split` arm and is untouched.

Both strategies now resolve to `inconclusive` — deliberately, rather than `split`: the situation genuinely *is* "no usable public verdict for this head", which is `ai_review_inconclusive`'s own semantics and copy, and routes to a neutral gate → human hold. `ai_review_split`'s copy instead asserts a disagreement between two reviewers that never happened, which would be actively misleading in single-reviewer mode.

A blank-only blockers array is still **not** a flag (`realBlockersOf` filters it) — that stays a clean pass, unchanged.

**One correction to the issue as filed:** its example (`computeScore divides by zero`) would *not* actually be filtered — `BARE_SCORE_TERM_PATTERN` is `/\bscore\w*\b/i` and `computeScore` has no word boundary before `Score`. What genuinely trips it is `reward`/`rewards`/`payout`/`farming`/`ranking`/`cohort`/`reviewability` (plain `.includes()`) and a bare `score`/`scores`/`scored`/`scorer`. All are now pinned as regression tests, and the camelCase non-match is pinned too so any future widening of the pattern is a deliberate, visible decision. Also worth noting: `synthesizeDefect` calls `toPublicSafe` with **no options**, so the gate-bearing title is filtered even on a repo in `LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS` — the allowlist never rescued this path.

### #9462 — a stuck capture marker silently disabled the screenshot gate (false merge)

Two compounding defects:

**The marker outlived its retries.** `scheduleVisualCaptureRetry`'s budget-exhausted early return only skipped writing `visualCaptureRetryPendingSha` *again* — the previous attempt's marker stayed. The only other clear (`markPullRequestVisualCaptureSatisfied`) requires a **successful** capture, which is by definition the thing not happening. Verified by exhaustive grep: exactly one clear and one setter across `src/`. So a head whose capture never succeeded stayed marked forever, and the doc comment claiming the gate "falls through to its normal (accurate) evaluation" was false in code.

**The deferral skipped instead of holding.** With the marker set, `screenshotTableMatch` was left `undefined` and the plan fell through to ordinary merit/CI evaluation — so a green PR planned a **merge**, with no finding, hold, or blocker representing "screenshot evidence unresolved". That is a disposition inversion: a PR the gate would have *closed* merged instead.

The gate is configured `action: close` for all three repos in the private server-side config, and was re-enabled two commits before the audit (73d4e9420, #9455).

## Scope note — #9464 deliberately excluded

The audit also found a false-**close** sibling (browserless render failures being swallowed). I implemented it, discovered it was wrong, and backed it out rather than ship it.

Inferring the failure downstream from "routes resolved but zero rendered pairs" cannot distinguish a browserless blip from a capture that legitimately found nothing renderable — and it broke the existing `#4110` test covering exactly that legitimate case, where the gate *should* close. #9464's real deliverable is to surface an explicit `renderFailed` signal from `capture.ts` where the failure is actually known. That needs its own change and its own tests, so it stays open for a follow-up.

## Validation

- `npx tsc --noEmit -p tsconfig.json` — clean
- `test/unit/ai-review.test.ts`, `test/unit/agent-actions.test.ts`, `test/unit/queue-3.test.ts` — **740 passed**

New regression tests:

- `combineReviews`: unpublishable blocker holds in `single`, `synthesis/either`, and `synthesis/both`; four parameterised real-vocabulary titles; publishable blocker still yields a defect; camelCase identifier still publishable; blank-only blockers still pass.
- Planner: unresolved evidence plans neither merge nor close; holds even with a green gate and merge autonomy (the exact bypass); does **not** hold for owner/admin/automation-bot; an explicit `false` or absent flag leaves the close path untouched.

**One test changed rather than added:** `"synthesized defect drops a blocker whose only finding is blank or unsafe (fail-safe...)"` asserted `inconclusive: false` for the unsafe case — i.e. it pinned the bug as correct behaviour, using the fixture literal `"Boost your reward payout"`. It has been split so the blank case keeps its old assertion and the unsafe case asserts the hold.

### Coverage gap I want to flag honestly

The marker-clearing path (`clearPullRequestVisualCaptureRetryPending` and its call site) has **no direct test**. The existing `#9030` exhaustion test starts from a PR that never had a marker written, so it exercises the early return but not the clear. The test that would cover it is the marker-SET-then-budget-EXHAUSTED sequence; I drafted it, found it needed a fetch-stub helper that does not exist (the neighbouring tests inline their stubs), and removed it rather than commit a broken file. Happy to add it in this PR before merge — flagging rather than letting `codecov/patch` surface it.

### DIFF
--- a/src/db/repositories.ts
+++ b/src/db/repositories.ts
@@ -4505,6 +4505,22 @@ export async function markPullRequestVisualCaptureRetryPending(env: Env, fullNam
     .where(and(eq(pullRequests.repoFullName, fullName), eq(pullRequests.number, number), eq(pullRequests.headSha, headSha)));
 }
 
+/** #9462: clear the retry marker once the bounded recapture budget is EXHAUSTED. Without this the marker set by
+ *  the previous attempt outlives the retries that justified it: `scheduleVisualCaptureRetry` early-returns at the
+ *  budget, which only skips writing it AGAIN, so `visualCaptureRetryPendingSha === headSha` stayed true forever
+ *  for that head. The only other clear is markPullRequestVisualCaptureSatisfied's, which requires a SUCCESSFUL
+ *  capture -- exactly the thing that is not happening. A permanently-marked head silently disabled the
+ *  screenshotTableGate's close (see the botCaptureRetryPending branch in processors.ts), so the deferral, meant
+ *  to be temporary, became a permanent bypass. Not scoped to headSha on the write: the caller has already
+ *  established which head it is finishing, and a row whose head moved on has a stale marker worth clearing too. */
+export async function clearPullRequestVisualCaptureRetryPending(env: Env, fullName: string, number: number, headSha: string): Promise<void> {
+  const db = getDb(env.DB);
+  await db
+    .update(pullRequests)
+    .set({ visualCaptureRetryPendingSha: null, updatedAt: nowIso() })
+    .where(and(eq(pullRequests.repoFullName, fullName), eq(pullRequests.number, number), eq(pullRequests.visualCaptureRetryPendingSha, headSha)));
+}
+
 /** Screenshot-table PRESENCE-mode staleness correlation (#stale-screenshot-table-fix): record the (headSha,
  *  evidenceFingerprint) checkpoint at which evaluateScreenshotTableGate's presence-mode check just satisfied
  *  the gate for this PR (see that function's `presenceModeSatisfiedState` result field and staleness comment).
--- a/src/queue/processors.ts
+++ b/src/queue/processors.ts
@@ -56,6 +56,7 @@ import {
   markPullRequestReviewsInvalidated,
   markPullRequestSurfacePublished,
   markPullRequestVisualCaptureSatisfied,
+  clearPullRequestVisualCaptureRetryPending,
   markPullRequestVisualCaptureRetryPending,
   markPullRequestScreenshotTablePresenceSatisfied,
   getLatestRegatedAt,
@@ -2759,6 +2760,7 @@ function buildAgentMaintenancePlanInput(args: {
   requiredContexts: Set<string> | null;
   blacklistEntry: ReturnType<typeof findBlacklistEntry>;
   screenshotTableMatch: AgentActionPlanInput["screenshotTableMatch"];
+  screenshotTableEvidenceUnresolved: AgentActionPlanInput["screenshotTableEvidenceUnresolved"];
   contributorCapMatch: AgentActionPlanInput["contributorCapMatch"];
   linkedIssueHardRule: AgentActionPlanInput["linkedIssueHardRule"];
   linkedIssueRulesConfig: Awaited<ReturnType<typeof loadLinkedIssueHardRules>>;
@@ -2790,6 +2792,7 @@ function buildAgentMaintenancePlanInput(args: {
     requiredContexts,
     blacklistEntry,
     screenshotTableMatch,
+    screenshotTableEvidenceUnresolved,
     contributorCapMatch,
     linkedIssueHardRule,
     linkedIssueRulesConfig,
@@ -2856,6 +2859,7 @@ function buildAgentMaintenancePlanInput(args: {
       : {}),
     copycatGateMode: settings.copycatGateMode,
     ...(screenshotTableMatch !== undefined ? { screenshotTableMatch } : {}),
+    ...(screenshotTableEvidenceUnresolved !== undefined ? { screenshotTableEvidenceUnresolved } : {}),
     ...(contributorCapMatch !== undefined ? { contributorCapMatch } : {}),
     // Always threaded (the DB layer populates it, default "over-contributor-limit"); the planner applies its
     // own fallback.
@@ -3451,7 +3455,11 @@ async function runAgentMaintenancePlanAndExecute(
     screenshotTableGateResult.violated && screenshotTableGateConfig.action === "close" && !botCaptureRetryPending
       ? { matched: true, reason: screenshotTableGateResult.reason }
       : undefined;
-  if (screenshotTableGateResult.violated && screenshotTableGateConfig.action === "close" && botCaptureRetryPending) {
+  // #9462: deferring the CLOSE is only half a deferral -- on its own it let the plan fall through to a MERGE.
+  // Thread the unresolved state into the planner so it holds the PR instead of silently skipping the gate.
+  const screenshotTableEvidenceUnresolved =
+    screenshotTableGateResult.violated && screenshotTableGateConfig.action === "close" && botCaptureRetryPending;
+  if (screenshotTableEvidenceUnresolved) {
     await recordAuditEvent(env, {
       eventType: "github_app.screenshot_table_close_deferred_capture_retry",
       actor: null,
@@ -3592,6 +3600,7 @@ async function runAgentMaintenancePlanAndExecute(
       requiredContexts,
       blacklistEntry,
       screenshotTableMatch,
+      screenshotTableEvidenceUnresolved,
       contributorCapMatch,
       linkedIssueHardRule,
       linkedIssueRulesConfig,
@@ -9476,7 +9485,26 @@ async function scheduleVisualCaptureRetry(
     previewPollAttempt: number;
   },
 ): Promise<void> {
-  if (args.previewPollAttempt >= MAX_PREVIEW_POLL_ATTEMPTS) return;
+  if (args.previewPollAttempt >= MAX_PREVIEW_POLL_ATTEMPTS) {
+    // #9462: the budget is spent, so the marker the PREVIOUS attempt wrote must be cleared here. Returning
+    // without clearing only skips re-writing it -- it left `visualCaptureRetryPendingSha === headSha` standing
+    // forever (the sole other clear needs a successful capture, which by definition never came), which silently
+    // and permanently disabled the screenshotTableGate's close for that head. Best-effort, matching the mark
+    // write below: a failed clear only means the gate stays deferred until the head moves, never a crash.
+    if (args.pr.headSha) {
+      await clearPullRequestVisualCaptureRetryPending(env, args.repoFullName, args.pr.number, args.pr.headSha).catch((error) => {
+        console.log(
+          JSON.stringify({
+            event: "visual_capture_retry_pending_clear_failed",
+            repoFullName: args.repoFullName,
+            pull: args.pr.number,
+            message: errorMessage(error).slice(0, 200),
+          }),
+        );
+      });
+    }
+    return;
+  }
   if (args.pr.headSha) {
     await markPullRequestVisualCaptureRetryPending(env, args.repoFullName, args.pr.number, args.pr.headSha).catch((error) => {
       console.log(
--- a/src/selfhost/metrics.ts
+++ b/src/selfhost/metrics.ts
@@ -129,6 +129,7 @@ export const DEFAULT_METRIC_META: readonly (readonly [string, MetricMeta])[] = [
   ["loopover_ai_review_non_cacheable_total", { help: "AI reviews skipped by cacheability rules.", type: "counter" }],
   ["loopover_ai_review_force_bypass_total", { help: "AI review cache force-bypass events.", type: "counter" }],
   ["loopover_ai_review_inconclusive_total", { help: "AI review inconclusive outcomes.", type: "counter" }],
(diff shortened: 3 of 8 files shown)


JSONbored:
Coverage gap closed. The marker-clearing path now has a direct regression test (\`screenshot-table gate (#9462): a marker written by an earlier attempt is CLEARED once the budget is exhausted\`).

It drives the real failing sequence the existing #9030 test could not reach — attempt 0 errors and writes the marker, then the final attempt exhausts the budget with the pipeline still failing — rather than starting from a PR that never had a marker written.

Verified it actually pins the fix: with the clear neutralised it fails with `expected 'vis62' to be null`, and passes with the fix in place. 158 tests green in that file.

Issue #9470, opened 2026-07-27
retention: dedupeSignalSnapshots keeps the STALE row on Postgres (MAX(rowid)→MAX(ctid)) — confirmed deleting fresh snapshots in production


JSONbored:
## Summary

`dedupeSignalSnapshots` keeps "the latest row per key" by selecting `MAX(rowid)`. On Postgres — **the primary production database** — the dialect shim rewrites `rowid` to `ctid`, which is a *physical heap location*, not insertion order. The dedupe therefore keeps a **stale** snapshot and permanently deletes the fresh one for a large fraction of keys.

**This is empirically confirmed on the production database, not inferred.** Measured on `edge-nl-01` on 2026-07-27, ~36% of multi-row keys for `contributor-evidence-graph` have `ctid` ordering that disagrees with actual recency.

## Mechanism (verified at HEAD, 776c4147a)

`src/db/retention.ts:256`:
```ts
const staleCondition = `signal_type = ?1 AND rowid NOT IN (SELECT MAX(rowid) FROM signal_snapshots WHERE signal_type = ?1 GROUP BY target_key)`;
```

`src/selfhost/pg-dialect.ts:192-194`:
```ts
export function translateRowid(sql: string): string {
  return sql.replace(/\browid\b/gi, "ctid");
}
```

`ctid` is `(page, tuple)`. After the age-prune frees pages, a **newer** row inserted into a reclaimed early page receives a **lower** `ctid` than an older row sitting on a later page. `MAX(ctid)` then selects a stale row, and the `NOT IN` delete removes the genuinely newest snapshot. The loss is permanent.

### The shim's own documentation forbids this usage

`src/selfhost/pg-dialect.ts:185-191` states `ctid` is safe only for

> internal bookkeeping resolved within one statement … **never for durable application-facing row identity**

But `MAX(rowid)` here decides **which row survives a DELETE** — durable row identity by definition. Meanwhile `retention.ts:236-239` asserts the invariant:

> "Latest" is the highest rowid: signal_snapshots is populated by a single sequential batch job, so insertion order and generated_at agree, and rowid (unlike generated_at) can never tie.

That is a SQLite-only truth which became false the moment the pg dialect shim was introduced. Neither comment references the other.

## Production evidence (edge-nl-01, `loopover-postgres-1`, 2026-07-27)

Comparing, per `(signal_type, target_key)` with more than one row, the row chosen by `ctid` against the row chosen by recency:

```sql
[...]

PR #9532, opened 2026-07-28
fix(ledger,retention): stop retention manufacturing tamper signals, and make the public counter durable (#9474, #9489)

### LINKED ISSUE #9474: retention: pruning breaks two consumers that assume permanence — ledger verification (false tamper signal) and the cumulative public counter
## Summary

Two retention rules interact badly with consumers that assume their tables are permanent. Both are latent today and become visible on a schedule, which is why they are worth fixing before they surface.

## 1. `decision_records` 180-day prune manufactures a false tamper signal

`src/db/retention.ts:57` prunes `decision_records` at 180 days. But `verifyDecisionLedger` requires the record for **every** ledger row — `src/review/decision-record.ts:460-463` maps `recordsById.get(row.recordId)` and yields `{kind: "missing_record"}` when absent — and **ledger rows are never pruned**. Verification also walks from `afterSeq = 0` genesis (`:412`).

So roughly 180 days after the rule first bites, any full-chain verification reports the chain broken at the first pruned record. For a **public** tamper-evidence feature, that is the worst possible false positive: a legitimate retention policy is indistinguishable from evidence of tampering, and the feature's own documentation teaches readers to treat it as the latter.

The retention comment describes `decision_records` as "a contributor's evidentiary trail" kept longer than most — the omission is that the ledger references it forever.

**Deliverables**
- [ ] Choose and implement one: exempt records still referenced below the newest anchor; teach the verifier a `pruned-before-<watermark>` tolerance tied to the retention cutoff; or start verification at the oldest surviving record's seq. Whichever is chosen must be legible to an external skeptic — the point of the feature is that a third party can check it.
- [ ] Document the chosen semantics in the public verification contract alongside the existing `short_tail` semantics.
- [ ] Test: prune a record below the anchor watermark, then verify ⇒ chain reports valid, with the pruning explicit rather than silent.

## 2. `orb_pr_outcomes` 90-day prune will shrink the cumulative public counter

#9415 added `{ table: "orb_pr_outcomes", column: "occurred_at", days: 90 }` (`src/db/retention.ts:77`). But `getOrbGlobalStats` SUMs the **entire table** — `src/orb/outcomes.ts:63-92`:
```sql
SUM(CASE WHEN o.outcome = 'merged' THEN 1 ELSE 0 END) AS merged,
SUM(CASE WHEN o.outcome = 'closed' THEN 1 ELSE 0 END) AS closed,
COUNT(*) AS total
FROM orb_pr_outcomes o ...
```
and `src/review/public-stats.ts:465-468` folds that into the **cumulative** homepage totals:
```ts
totals.merged  += orb.merged;
totals.closed  += orb.closed;
totals.handled += orb.total;
```

Once rows begin aging past 90 days, the "all-time" counters plateau and then **decrease** — a public number visibly going backwards, roughly from **2026-10-25** given #9415's merge date.

**Deliverables**
- [ ] Fold pruned rows into a durable running total before deletion (a small `orb_outcome_totals` roll-up updated by the prune), or exclude `orb_pr_outcomes` from retention and bound it another way.
- [ ] Test: prune rows older than the window ⇒ the cumulative public total is unchanged.
- [ ] Check the same pattern for the other cumulative public surfaces that fold pruned tables — `computeFleetAnalytics` is the obvious neighbour to audit.

## Expected outcome

Retention never changes the meaning of a published number, and never produces a signal that reads as tampering.

### LINKED ISSUE #9489: ledger: four anchoring defects — unretried failures at a quiet tip, a tip-read race publishing a false tamper signal, 1MB file truncation, short_tail blind spots
## Summary

Four defects in the decision-ledger anchoring feature that shipped 2026-07-27 (#9413/#9274). Each one degrades the property the feature exists to provide — an external, tamper-evident anchor a skeptic can verify — and two of them produce signals that read as **evidence of tampering** when nothing was tampered with.

## 1. A failed anchor at a quiet tip is never retried

`src/review/ledger-anchor-scheduler.ts:39-42` decides `tipUnchanged` by comparing the rowHash against the last attempt **regardless of its status** (`src/review/ledger-anchor-persistence.ts:161-167`).

Scenario: Rekor returns 429 at seq N; the ledger then goes quiet (a weekend); every hourly tick returns `unchanged`; the tip therefore has **no valid external anchor indefinitely** — exactly the unanchored window the feature exists to bound.

It is also backend-blind: git succeeding at seq N masks Rekor's failure at seq N, because there is no per-backend watermark.

- [ ] Retry when the last attempt **for that backend** at this rowHash was `failed`, with backoff.
- [ ] Track the watermark per backend.
- [ ] Test: rekor fails, ledger stays quiet ⇒ the next tick retries rekor.

## 2. A tip read race publishes a signed, unretractable false tamper signal

`src/review/decision-record.ts:346-354` issues the `COUNT(*)` and the tip row as two **separate** queries under `Promise.all` — no snapshot isolation. A concurrent `appendDecisionLedger` landing between them yields `totalCount = seq + 1` paired with the **old** rowHash.

Per the payload documentation (`src/review/ledger-anchor.ts:29-31`), that seq/totalCount pairing is precisely how a verifier detects truncation-and-rechaining. So the race publishes a **signed, Rekor-anchored, unretractable** false tamper signal.

- [ ] Read the tip and count in one statement (subselect) or a `DB.batch`.
- [ ] Test: concurrent append during the tip read ⇒ the anchored payload is internally consistent.

## 3. The git backend truncates `anchors.jsonl` past 1 MB

`src/review/ledger-anchor-git.ts:66-75`: for a file between 1 MB and 100 MB, GitHub's Contents API returns `content: ""` with `encoding: "none"`. The code checks `typeof data.content === "string"`, which `""` satisfies, so `existingContent` becomes `""` and the subsequent PUT **rewrites the file to a single line** — recording `status: "ok"`.

At roughly 300 B per line and hourly cadence, that is about 4–5 months out. History survives in the git commits, but the file a skeptic is told to read **shrinks** — indistinguishable from the tampering signal the module's own header teaches them to look for. The whole-file read-modify-write is also O(n²) in bytes over time.

- [ ] Refuse (or rotate) when `data.content === "" && data.size > 0`; consider the Git Data API instead of Contents.
- [ ] Test: simulate the empty-content/non-zero-size response ⇒ no truncating write occurs.

## 4. `short_tail` is transiently false-positive and permanently blind to interior orphans

`src/review/decision-record.ts:487-493`: a verification running between the record INSERT and the ledger append — or after a permanently failed append (`:225-236`) — reports `short_tail`, i.e. "tampering", on a **public** endpoint.

Conversely, once any newer row chains, an unchained record's `created_at` falls behind `lastVerifiedCreatedAt` and that orphan becomes **invisible forever**, because the completeness reconciliation only ever examines the tail.

- [ ] Distinguish "append in flight" from "record missing from the chain" (a short grace window on recency, or an explicit in-flight marker).
- [ ] Give completeness reconciliation a path that can find interior orphans, not only the tail.
- [ ] Test: verify during an in-flight append ⇒ no tamper signal; an interior orphan is detected.

## Minor

`loadPublicLedgerAnchors`'s keyset cursor is `created_at < ?` with `ORDER BY created_at DESC, id DESC` (`src/review/ledger-anchor-persistence.ts:106-127`) — ties on `created_at` at a page boundary are skipped. Use a composite `(created_at, id)` cursor.

## Verified solid (do not re-audit)

The signing and verification crypto in `src/review/ledger-anchor.ts` is sound: canonical serialization shared by sign and verify, fail-closed key parsing, ambiguity-safe `currentAnchorKey`, and `parseRekorResponse` degrading to a recorded failure. One documented nuance: nothing checks `payload.at` against a key's `[notBefore, notAfter]` window at verification — weak protection anyway given an attacker-chosen `at`, reasonable to leave as-is.

JSONbored:
## Summary

Retention and verification were written as if the other did not exist. Both consumers that assumed permanence get fixed here, plus #9489's final defect — the `short_tail` false positive — which belongs with them because it is the same failure shape: **a legitimate operation reading as tampering on a public endpoint.**

Closes #9474
Closes #9489

### #9474a — the 180-day record prune would have broken every full-chain verification

`decision_records` prunes at 180 days; ledger rows are kept forever; `verifyDecisionLedger` demands the record for **every** ledger row and walks from genesis. ~180 days after the rule first bit, every verification would report `missing_record` at the first pruned row — a false tamper signal manufactured by the published retention policy, on the one endpoint whose entire point is that a skeptic can trust it.

The verifier now tolerates a missing record **only** when the ledger row's `createdAt` is older than the published `decision_records` window, counting it in a new `prunedRecords` result field. Two properties make this legible rather than a loophole:

- The tolerance keys on the **ledger row's** timestamp, which is inside the hash chain — backdating it to sneak a fresh deletion under the cutoff breaks `row_hash_mismatch` first. A recent out-of-band deletion is still `missing_record`.
- The cutoff comes from `retentionCutoffIsoForTable("decision_records")`, reading `RETENTION_POLICY` itself — the verifier and the prune cannot drift apart again, because they now share one source of truth.

What is genuinely given up is exactly what pruning gives up: the content re-check for that row. The chain checks still run over it, and the committed digest stays published, so a challenger holding the original preimage can still prove a historical rewrite by hand. Documented in `what-you-can-verify.mdx` and the OpenAPI descriptions.

### #9474b — the cumulative public counter would have visibly shrunk from ~2026-10-25

`getOrbGlobalStats` SUMs the **entire** `orb_pr_outcomes` table and public-stats folds it into the homepage's *all-time* merged/closed/handled totals — but #9415 gave the table a 90-day window, so the "all-time" numbers would have plateaued and then **decreased**.

The prune now folds every row it deletes into a durable `orb_outcome_rollups` table (migration 0198) **in the same batch transaction as the delete** — a fold and delete that could commit separately would either double-count or under-count — and the stats query adds the rollup back. Three semantics preserved exactly:

- Only rows the live query would have counted are folded (registered installs, no published review surface); everything else is deleted without folding, so the public total does not *jump* on prune day either.
- Rollups key on the lowercased `account_login`, so `excludeAccount` de-dup keeps working after the raw rows are gone.
- `orb_pr_outcomes` moved to the **top** of `RETENTION_POLICY`, above `audit_events`: the fold's own-ledger exclusion needs audit rows that the audit prune — same 90-day window — would otherwise have deleted moments earlier in the same pass. An ordering-guard test pins this.

**Neighbour audit (deliverable 3):** `computeFleetAnalytics` is clean — every `orb_signals` read is explicitly windowed (`received_at >= cutoff`) and `orb_signals` has no retention rule at all, so there is no unwindowed aggregate over a pruned table there.

### #9489d — `short_tail` was transiently false-positive and permanently blind to interior orphans

Two independent defects in the completeness reconciliation:

- **Grace.** A record and its ledger append are two writes milliseconds apart, and a verify landing between them reported `short_tail` — "tampering" — for a state every healthy write passes through. Records younger than a 5-minute `LEDGER_APPEND_GRACE_MS` are now simply not yet due; a genuinely failed append still surfaces on the next verify after the grace lapses (and its own error-level alarm fired the moment it happened).
- **Interior orphans.** The old check compared `created_at` against the verified **tail** only, so the moment any newer record chained cleanly, an unchained record behind it became invisible *forever*. The reconciliation now asks the real question — *does any ledger row vouch for this record?* — via a NOT EXISTS anti-join (indexed by the new `decision_ledger_record_id` index). The newest orphan decides the break kind: past the tail it is the truncated-tail signature `short_tail` always meant; behind it, a new `unchained_record` break carrying the record id — the failed-append signature.

## Validation

- `npx tsc --noEmit`, `db:migrations:check` (contiguous 0001..0198), `selfhost:env-reference:check`, `ui:openapi` regenerated, `git diff --check` — all clean
- **260 passed** across the decision-record, retention, public-stats, orb-outcomes, pg-retention and queue-4 suites
- Patch coverage against this diff: **0 uncovered changed lines**

**Regressions, each verified to fail against the unfixed code:** a pruned record verifies clean and is counted, not reported as tampering; a verify during an in-flight append reports ok; an interior orphan is detected as `unchained_record`; the cumulative public total is byte-identical before and after a prune with the aged rows gone.

**Invariants:** a *recent* record deleted out of band is still `missing_record` — the tolerance cannot launder fresh deletions; past the grace window an unchained record IS reported — the grace bounds the blind spot, it does not remove the check; a fully healthy chain reports `prunedRecords: 0` and both new mechanisms inert; rows the live query never counted (unregistered installs, own-ledger-published PRs) are deleted **without** folding; a second prune run folds nothing new; `excludeAccount` still de-dups against folded totals; dry-run neither folds nor deletes; and the policy-ordering guard.

The pg mock pool gained `connect()` (the fold's batch transaction checks out a client) and the fold statements — both still route through the same query fn, so the `rowid`-regression guard sees batch-issued SQL too.

### DIFF
--- a/apps/loopover-ui/content/docs/what-you-can-verify.mdx
+++ b/apps/loopover-ui/content/docs/what-you-can-verify.mdx
@@ -29,8 +29,24 @@ hash, so any edit to history breaks the chain at a point you can locate.
 curl -s "https://api.loopover.ai/v1/public/decision-ledger/verify" | jq
 ```
 
-Returns `{ ok, checked, nextAfterSeq, tipSeq, tipHash, totalCount }`, and a `break` object with a
-`409` status if the chain is inconsistent. No API key — **anyone** can run it.
+Returns `{ ok, checked, nextAfterSeq, tipSeq, tipHash, totalCount, prunedRecords }`, and a `break`
+object with a `409` status if the chain is inconsistent. No API key — **anyone** can run it.
+
+Two response fields encode deliberate, published semantics rather than tolerance for tampering:
+
+- **`prunedRecords`** — decision *records* (the preimages) are retained for 180 days; ledger rows are
+  kept forever. A ledger row whose hash-chained `createdAt` is older than that window may therefore
+  reference a pruned record: the chain checks still hold for it, only the content re-check is
+  impossible, and it is counted here instead of reported as `missing_record`. The tolerance keys on
+  the *ledger row's* timestamp, which is inside the hash chain — backdating it to sneak a fresh
+  deletion under the window breaks `row_hash_mismatch` first. What pruning genuinely gives up: the
+  row's committed digest stays published, but only a challenger holding the original preimage can
+  still prove a historical rewrite of it.
+- **Append grace (5 minutes)** — a record and its chain row are two writes moments apart, so a
+  record younger than the grace window with no chain entry is "append in flight", not a break. Older
+  than that, it is reported: past the verified tip as `short_tail` (the truncated-tail signature),
+  or behind it as `unchained_record` (the failed-append signature — interior orphans are found by an
+  anti-join over the whole record set, not just the tail).
 
 <Callout variant="note">
   **Trust assumption: tamper-evident, externally anchored.** The check above still catches sequence
--- a/apps/loopover-ui/public/openapi.json
+++ b/apps/loopover-ui/public/openapi.json
@@ -20104,10 +20104,10 @@
         "summary": "Verify a window of the hash-chained decision ledger (resumable via afterSeq)",
         "responses": {
           "200": {
-            "description": "Window verified clean; nextAfterSeq is the resume cursor (null at the tip). Every response also carries tipSeq/tipHash/totalCount for third-party checkpointing."
+            "description": "Window verified clean; nextAfterSeq is the resume cursor (null at the tip). Every response also carries tipSeq/tipHash/totalCount for third-party checkpointing, and prunedRecords — the count of rows whose record preimage was legitimately pruned by the published retention window (chain checks still hold for them; only the content re-check is impossible, and the committed digest stays published)."
           },
           "409": {
-            "description": "First break found: sequence_gap | predecessor_mismatch | row_hash_mismatch | short_tail (a record exists past the verified tip with no chain entry)"
+            "description": "First break found: sequence_gap | predecessor_mismatch | row_hash_mismatch | missing_record | content_mismatch | short_tail (a record newer than the verified tip has no chain entry — the truncated-tail signature) | unchained_record (an INTERIOR record has no chain entry — the failed-append signature). Records younger than the 5-minute append grace window are not reported: the record insert and its chain append are two writes moments apart, and a verify landing between them is not evidence of tampering."
           }
         },
         "security": [
--- a/migrations/0198_orb_outcome_rollups.sql
+++ b/migrations/0198_orb_outcome_rollups.sql
@@ -1,0 +1,25 @@
+-- #9474: durable running totals for orb_pr_outcomes' CUMULATIVE public consumer.
+--
+-- #9415 gave orb_pr_outcomes a 90-day retention window, but getOrbGlobalStats SUMs the ENTIRE table and
+-- public-stats folds that into the homepage's all-time merged/closed/handled counters. Once rows began aging
+-- past 90 days (~2026-10-25 given #9415's merge date) the "all-time" numbers would have plateaued and then
+-- visibly DECREASED -- a published cumulative counter going backwards.
+--
+-- The retention prune now folds each about-to-be-deleted row into this table first, atomically (same batch
+-- transaction as the delete -- see pruneExpiredRecords' orb_pr_outcomes special case), and getOrbGlobalStats
+-- adds these totals to its live scan. Keyed per LOWERCASED account_login (matching the stats query's own
+-- LOWER() comparison; '' for a NULL login) so its excludeAccount de-dup keeps working after the raw rows are
+-- gone. Only rows the live query would have counted are folded: registered installations, no published
+-- review surface.
+CREATE TABLE IF NOT EXISTS orb_outcome_rollups (
+  account_login TEXT PRIMARY KEY,
+  merged INTEGER NOT NULL DEFAULT 0,
+  closed INTEGER NOT NULL DEFAULT 0,
+  total INTEGER NOT NULL DEFAULT 0,
+  updated_at TEXT NOT NULL
+);
+
+-- #9489/#9474: verifyDecisionLedger's completeness reconciliation now asks "does ANY ledger row vouch for
+-- this record" (a NOT EXISTS anti-join finding interior orphans, not just tail ones); without this index
+-- that is a full ledger scan per candidate record.
+CREATE INDEX IF NOT EXISTS decision_ledger_record_id ON decision_ledger (record_id);
--- a/scripts/check-schema-drift.ts
+++ b/scripts/check-schema-drift.ts
@@ -54,6 +54,9 @@ export const RAW_SQL_ONLY_TABLES: Set<string> = new Set([
   "orb_export_cursor",
   "orb_github_installations",
   "orb_instances",
+  // #9474: durable running totals folded from orb_pr_outcomes (itself raw-SQL-only, below) by the
+  // retention prune, and summed back in by getOrbGlobalStats -- both via env.DB.prepare, no Drizzle use.
+  "orb_outcome_rollups",
   "orb_pr_outcomes",
   "orb_relay_failures",
   "orb_reuse_counters",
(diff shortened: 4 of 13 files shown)


JSONbored:
Addressed the batching concern — it was right, and "safe at today's volume" is exactly the argument that ages badly (one fleet-wide backfill turns that statement into a multi-million-row delete).

The prune is now **bounded like every other table**, without giving up the atomicity the fold requires. Each slice picks a **timestamp boundary** and both statements share the identical `occurred_at` predicate, so the fold and the delete provably see the same rows while still committing together in one `batch()`. Keying the slice on the retention column rather than on rows avoids both alternatives' hazards: `rowid`/`ctid` is rewritten by pg-dialect to a *physical* location (the #9470 bug), and row-value tuple syntax has uneven SQLite support.

The boundary is taken **inclusively** (`<=` on the `occurred_at` at `OFFSET batchSize - 1`), which is load-bearing rather than cosmetic: an exclusive `<` against a run of rows sharing one timestamp selects zero rows and spins forever. I verified that directly — flipping it to `<` hangs the new tie test until the timeout. A slice is therefore `batchSize` rows plus any ties at the boundary, and the final slice takes the remainder.

Three tests added:
- **REGRESSION** — 25 aged rows at `batchSize: 10` drain across multiple bounded slices, with the cumulative public total unchanged.
- **INVARIANT** — 12 rows sharing one timestamp still make progress (the anti-spin guard).
- **INVARIANT** — `maxPerTable` still caps a single run, and the public total is exact *at every point*, not only once the cohort fully drains.

152 tests green across the retention, pg-retention, orb-outcomes, decision-record and public-stats suites; patch coverage still 0 uncovered changed lines.

Issue #9492, opened 2026-07-27
maintainability: processors.ts (16.2k lines, 437 touches) is the structural cause of this audit's drift and determinism gaps; plus a confirmed dead wire


JSONbored:
## Summary

Two maintainability findings with measured churn evidence, plus one confirmed piece of dead code. Filed because both are the *structural* cause of defects found elsewhere in this audit — not as generic cleanup.

## 1. `src/queue/processors.ts` — 16,200 lines, 437 touches since 2026-07-01

86 of those touches came in the last week alone. It is where most of this audit's confirmed bugs physically live:

- the decision-clock violations (replay determinism)
- both sibling-drift families (#9491) — five near-identical command handlers, two panel twins, three advisory runners, all drifting **inside one file**
- the cap-close vs accept-path parity gap (#9482)
- the visual capture-blip wiring (#9462, #9464)

The drift is mechanical: a guard is added to the instances someone greps for, and the next one is 8,000 lines away in the same file.

**Highest-value splits, in order:**
- [ ] The ~9 slash-command and panel handlers behind one shared dispatch owning state, redelivery and rate-limit guards once. (Also delivers most of #9491.)
- [ ] The three `run*ForAdvisory` spend gates behind one shared precondition struct.
- [ ] The plan-and-execute pass into its own module with a **required, typed decision-pass context** (clock, `authorLogin`, `baseRef`) that both the live path and the approval-accept path must construct. This one change structurally fixes #9482 and roughly half the determinism gap below, because both are caused by ad-hoc optional context fields.

## 2. Replay determinism: the recorded clock reaches one consumer

#9256/#9028 shipped "replayable time", but the captured `decisionClock` is threaded into exactly one consumer (`maybeForceFreshRebase`, `processors.ts:3911`). At least three other clock reads still decide outcomes:

- `processors.ts:2907` — `activeMergeBlockedSha(pr, pr.headSha, Date.now())` feeds the plan and the recorded reason code; it is a **second, unrecorded** clock read in the same pass that captures `decisionClock` at `:3184`.
- `processors.ts:3498` — account-age `Date.now()` → `isNewAccount` → halves the contributor cap → cap close. `decisionClock` is in scope and unused; the twin `src/queue/account-age-throttle.ts:13` has no `nowMs` seam at all.
[...]

Issue #9482, opened 2026-07-27
approval-queue: accept path builds a thinner executor context — #9159 mutex, moderation ladder and #9055 base guard all silently inert


JSONbored:
## Summary

`decidePendingAgentAction` builds a **thinner** executor context than the live path, so three separately-shipped protections are silently inert on every approval-queue accept. Two independent audit sweeps found this by different routes, which is worth noting: it is structural, not incidental.

## Mechanism (verified at HEAD, 776c4147a)

The accept-path context (`src/services/agent-approval-queue.ts:490-531`) omits `authorLogin`, `expectedBaseRef` and `moderationSettings`. Every one of them is optional in the type, so nothing forces parity with `runAgentMaintenancePlanAndExecute`'s context (`src/queue/processors.ts:3065`, `:3129`, `:3927`).

### 1. #9159's cap-lock mutex provides zero exclusion on accepts

Executor step 8c claims `claimContributorCapLock(env, repo, ctx.authorLogin ?? "")` (`src/services/agent-action-executor.ts:830-851`). With `authorLogin` undefined the key becomes `contributor-cap-lock:<repo>:` — an **empty author**. The live cap-close locks `…:<author>` (`src/queue/transient-locks.ts:197-206`).

So the mutex #9159 shipped ("the two can never both act on a stale view") does nothing on this path: the #7284 race between an accepted merge and a sibling's cap-close for the same author stays fully open. Additionally, *all* accept-path merges in a repo now contend on one shared empty-author key.

### 2. Moderation escalation silently no-ops

`maybeEscalateModeration` early-returns on `!args.authorLogin` (`src/services/agent-action-executor.ts:987`). Every blacklist, contributor-cap or review-nag close accepted from the queue therefore records **no moderation violation** — so it never counts toward the warning/ban threshold or auto-blacklist. A contributor whose enforcement closes always route through approval accumulates zero standing violations. The file explicitly supports staging cap closes (`agent-approval-queue.ts:505-508`), so this is a reachable path, not a theoretical one.

### 3. #9055's base-retarget guard is inert

`fetchPullRequestFreshness` only performs the `base_changed` check when `expectedBaseRef` is set (`src/github/pr-freshness.ts:96-99`). On accepts it is `undefined`, so a contributor who **retargets the PR base after staging** (head unchanged, head CI still green, head-SHA pin still matching) is merged into a base that the reviewed diff and CI never targeted — on a maintainer accept made against stale information. This is a wrong-merge class.

[...]

Issue #9028, opened 2026-07-26
Replay v2: re-query action-match mode + wall-clock capture for staleness rules


JSONbored:
Parent: #8828. Successor to #8838 (whose deterministic core — input capture, pure replay, CLI, CI smoke — ships in the closing PR); this issue carries the two deliberately deferred pieces so #8838 closes clean.

## Requirements
1. **Re-query mode** (`replay-decision.ts --requery`): re-runs the model for the same target and reports an ACTION-MATCH RATE — never labeled "reproducibility" (hosted inference is not bit-deterministic even at temperature 0; the docs must state why). Blocked on: persisting the full prompt text alongside the existing `promptDigest` (privacy + row-size call: prompt text embeds the diff — likely another private sibling table with retention bounds, mirroring decision_replay_inputs' posture).
2. **Wall-clock capture**: when staleness rules (`requireFreshRebaseWindowMinutes`, `staleBaseAheadByThreshold`) join the replay scope, the evaluation must read time through an injected `nowMs` recorded into `decision_replay_inputs.replay_json` — a seeded regression test per rule proving replay at the recorded instant is bit-exact and replay at a different instant is NOT silently accepted.

## Test Coverage Requirements
99%+ patch coverage, branch-counted, on all changed `src/**` lines; both arms of every time-dependent branch.

## Expected Outcome
The replay harness covers every decision input including time, and operators get an honest action-match metric for the one irreducibly nondeterministic stage.

Issue #9537, opened 2026-07-28
contract(stdio): migrate the stdio MCP server's 102 tools to @loopover/contract — typed handlers, no hand-mirrored shapes


JSONbored:
### Context

Part of #9515; the stdio third of what #9518 originally covered, split out so each migration is a single reviewable PR.

`packages/loopover-mcp/bin/loopover-mcp.ts` registers **102 tools**, of which (after #9517's pilot) only a handful declare an `outputSchema`. Its ~85 shape constants (`:371-1104`) hand-mirror the remote server's — several say so in their own comments (`:376`, `:381`, `:407`, …) — and the registration helper erases all typing (`:1740-1742`, three `any`s), so every handler takes `(input: any)`. `STDIO_TOOL_DESCRIPTORS` (`:1108-1668`) is a third hand-maintained list of name/category/description for all 102.

Two divergences #9517 found and modelled as unions, which converge here: `loopover_get_repo_context` and `loopover_get_pr_reviewability` return **different payloads** from this server than from the remote one (this server proxies the REST route; the remote builds its own object). Resolving them is wire-visible and belongs in this issue's batches, not the keystone's.

### Requirements

1. **Migrate all 102 tools to register from `@loopover/contract`**, by category, reusing the entries #9518 lands for tools that exist on both servers rather than adding a second definition.
2. **Typed handlers.** The registration helper stops erasing types; handlers derive their input type via `z.infer`. No `(input: any)` remains — enforce with a lint/grep gate.
3. **Real output schemas on every tool**, replacing unschematized `structuredContent`.
4. **Delete `STDIO_TOOL_DESCRIPTORS`.** Name, category, and description come from the registry; the `tools` CLI command and `_meta.category` read it.
5. **Converge the two divergent payloads** with the remote server, so one tool name means one shape. Whichever side changes, the change is wire-visible and gets called out explicitly in the PR body.
6. **`buildPlanDag`/`validatePlanDag`** (`:173-215`, hand-duplicated from `src/services/plan-dag.ts`) move to one shared home and the duplicate is deleted.
7. No wire regression otherwise: names, descriptions, and input shapes preserved or widened, asserted by a `tools/list` snapshot per batch. Full branch-counted coverage.

### Non-goals

- The remote (#9518) and miner server migrations.
- The typed API client and CLI-surface generation (#9521) — though this issue's contract entries are what make that possible.

### Deliverables

- [ ] All 102 tools registered from the contract, handlers typed, `any` gate in CI
[...]

PR #9538, opened 2026-07-28
fix(replay,maintainability): thread the recorded decision clock into three unrecorded reads, delete a dead wire, and add a reachability check (#9492)

### LINKED ISSUE #9539: replay: the recorded decision clock reaches one consumer while three other reads still decide outcomes
Parent: #9492

## Summary

#9028/#9256 shipped "replayable time": exactly ONE `Date.now()` per decision pass, captured as `decisionClock` and recorded into `DecisionReplayInput`. That guarantee is worth nothing if a **second, unrecorded** clock read in the same pass also decides an outcome — the recorded instant is then not the instant that decided, and the decision is unreplayable.

The captured clock reached exactly **one** consumer (`maybeForceFreshRebase`, `processors.ts:3911`). At least three other reads still decided outcomes.

## Why this fails silently, in the worst direction

`replayDecision` pins `action` and only re-derives `evaluateGateCheck`. So replaying a decision that was actually made on an unrecorded clock read reports **`verdict: "match"`** — a **false certification**. The determinism feature actively vouches for the pipeline precisely where the pipeline is not deterministic. A divergence that is reported as a match is strictly worse than no replay at all.

## The three reads

1. **`processors.ts:2907` — `activeMergeBlockedSha(pr, pr.headSha, Date.now())`.** Feeds the plan AND the recorded reason code. It is a second, unrecorded clock read in the very same pass that captures `decisionClock` at `:3184`.
2. **`processors.ts:3498` — account age.** `Date.now()` → `isNewAccount` → halves the contributor cap → can flip a cap **close**. `decisionClock` is already in scope and simply unused. The twin helper `src/queue/account-age-throttle.ts:13` has no `nowMs` seam at all.
3. **`src/review/unlinked-issue-guardrail.ts:277`** (also `:192`, `:92`) — a wall-clock gap chooses **close vs hold**.

There is also a class no JS-side grep can see: SQL-side `datetime('now', ?)` in `src/review/submitter-reputation.ts`.

## Requirements

1. Every clock read that can change a decision **inside the recorded pass** must consume the captured instant.
2. A read deliberately left on the live clock must be **named** in `decision-replay.ts`'s honest-scope note. The current note does not list them, and a scope claim that silently widens is worse than a narrow one.
3. No behaviour change for callers outside the decision pass.

## Deliverables

- [ ] Thread `decisionClock.nowMs` into the merge-blocked, account-age and unlinked-issue paths.
- [ ] Give `isBelowAccountAgeThreshold` a `nowMs` seam.
- [ ] Make the plan-input clock field **required**, not optional — optionality is how this class of context field drifts (the #9482 lesson); an optional clock lets a future call site reintroduce the gap with no type error.
- [ ] Extend the honest-scope note at `src/review/decision-replay.ts:26-33` to name every read still on the live clock.

## Tests (must fail against current main)

- [ ] The velocity exception evaluated at a supplied instant diverges from the live clock. Use a **confirmed miner** — the exception only applies to one, so it is the only configuration where the gap is decisive; an unconfirmed author closes on repeat regardless of gap, which would make the assertion pass without exercising the clock at all.
- [ ] The repeat **lookback** shares that same instant, not just the velocity gap.
- [ ] Account age judged at a supplied instant flips `isNewAccount`.
- [ ] Invariant: omitting the instant preserves live-clock behaviour byte-identically at every new seam.

## Expected outcome

A decision made in the recorded pass is replayable from the recorded instant, and `verdict: "match"` means what it claims.

### LINKED ISSUE #9540: maintainability: a dead wire ships green, and no check exists for the class (26 wires, no registration protocol)
Parent: #9492

## Summary

`src/review/issue-rag-wire.ts` has **zero** production importers. Both real consumers import `packages/loopover-engine/src/issue-rag-query` directly; only the wire's own test imports the wire.

Test coverage **camouflages** this rather than catching it: a dead module's own test exercises it perfectly well, so it reports green while contributing nothing to the running system. Commit `39cc9583c` shows this class has bitten before.

There are 26 `*-wire.ts` files and no registration protocol; ~18 are direct-import-only, so a forgotten import is a **silently dead feature**. The one real registry (`src/review/feature-activation.ts:80-89`, a total `Record` over `CONVERGED_FEATURE_KEYS`) is compile-time drift-proof and is the right model — but it covers only converged features.

No mechanical check exists for the class: there is no knip or ts-prune, and none of the ~15 `scripts/check-*-drift.ts` covers reachability.

## Requirements

1. Delete the dead wire (or wire it up, if it was meant to be used).
2. Add a check that makes the class **fail CI**, not merely be noticed by a future reader.
3. The check must not be defeated by an allowlist of individual dead files — that would defeat its own purpose. A false positive is fixed at the root (a legitimate importer root, or a declared entry point with a stated reason), never by exempting the file.

## Deliverables

- [ ] Delete `src/review/issue-rag-wire.ts` and its test.
- [ ] Add `scripts/check-dead-source-files.ts`: every `src/**` file must have at least one importer outside its own test. `scripts/check-import-specifiers.ts` is a ready template (injectable `listFiles`/`readFile` seams, pure core, `main()` guard).
- [ ] Wire it into `test:ci` **and** the CI workflow, alongside the sibling drift checks.
- [ ] Deliberately narrow scope: no re-export chasing, no unused-export analysis. Answer only "is this file imported by anything outside its own test" — the shape that actually failed here, cheap enough to run per-PR.
- [ ] Fix whatever the first run surfaces, at the root.

## Tests

- [ ] Both dead shapes are flagged, with distinct reasons: "only its own test imports it" and "no importer at all".
- [ ] One production importer is enough.
- [ ] A test that is NOT the module's own counts as a real importer (a shared helper is alive).
- [ ] `scripts/**` counts as a production importer — several CLIs are the sole consumer of a live `src/**` module.
- [ ] Declared entry points are never flagged.
- [ ] Barrels keep their members alive via `export ... from`.
- [ ] A specifier reaching outside `src/**` is ignored rather than miscounted.
- [ ] The real repo tree is clean, so the check stays green in CI.

## Expected outcome

A module that stops being reachable fails CI on the PR that orphans it, instead of sitting green for months.

JSONbored:
## Summary

Two of #9492's three findings, both structural rather than cosmetic: the recorded decision clock reached exactly one consumer while three other clock reads still decided outcomes, and a dead feature file sat green in CI for months with no mechanism that could have noticed.

Closes #9539
Closes #9540

### #9539 — replay determinism: three unrecorded reads decided outcomes

#9028/#9256 shipped "replayable time": one `Date.now()` per decision pass, recorded into the replay input. Its value is destroyed entirely by a **second, unrecorded** read in the same pass — the recorded instant is then not the instant that decided.

Worse, this failed silently in the direction that matters. `replayDecision` pins `action` and only re-derives `evaluateGateCheck`, so replaying a decision made on an unrecorded read reports **`verdict: "match"`** — a false certification, not a caught divergence. The feature would have reported the pipeline deterministic precisely where it wasn't.

Three reads now take the recorded instant:

- **`activeMergeBlockedSha`** — fed the plan *and* the recorded reason code, from a second `Date.now()` in the very function whose caller captures `decisionClock`. Threaded as a **required** `decisionNowMs` field, deliberately: optionality is how this class drifts (the #9482 lesson), and an optional clock lets a future call site reintroduce the gap with no type error.
- **Account age → `isNewAccount`** — halves the contributor cap and can therefore flip a cap **close**. `decisionClock` was already in scope, unused. The twin helper `isBelowAccountAgeThreshold` had no `nowMs` seam at all; it has one now.
- **`resolveUnlinkedIssueMatchDisposition`** — its velocity-exception gap chooses **close vs hold**. All three of its clock reads now share one instant, so the DB lookback windows and the gap agree.

The honest-scope note in `decision-replay.ts` is updated to name what is *still* live-clock — the three `isBelowAccountAgeThreshold` callers outside this pass (each its own pass, no captured instant, nothing to be consistent with) and the SQL-side `datetime('now', ?)` in `submitter-reputation.ts`, a class no JS-side seam can cover. A scope claim that silently widens is worse than a narrow one.

### #9540 — a dead wire, and the missing check for the class

`src/review/issue-rag-wire.ts` had **zero** production importers: both real consumers imported the engine module it re-exported directly, and only its own test imported the wire. Coverage cannot catch this — a dead module's own test exercises it perfectly well.

- **Deleted** it.
- **`scripts/check-dead-source-files.ts`** (new, in `test:ci` and CI): every `src/**` file must have at least one importer outside its own test. Deliberately narrow — no re-export chasing, no unused-export analysis — it answers only the question #9492 actually found, cheaply enough to run per-PR.
- Running it surfaced **two more** dead files and **one genuinely dead feature**:
  - `src/review/content-lane/index.ts` — a barrel with no importers, and **already drifted**: it omits `survivingExistingEntries`, which predates this PR. Deleted.
  - **`src/miner/soft-claim.ts` was a shipped-but-unregistered feature.** `buildSoftClaimSpec` landed in #2813 and no MCP tool ever exposed it, so the capability simply did not exist at runtime. **Wired up** as `loopover_post_soft_claim`, following the identical `local-write-tools` pattern its nine siblings already use.
- The checker's four apparent false positives were fixed at the root rather than allowlisted: `scripts/**` is a legitimate importer root (`check-migrations.ts` and `draft-issue.ts` are the sole consumers of two live `src/**` modules), and `server.ts` plus the two esbuild alias targets are declared entry points with reasons.

## Scope

#9492 originally bundled three findings in one issue, which is why this PR had nothing it could close *in full*. It is now an epic over three sub-issues — #9539 and #9540, both closed here **in their entirety**, and **#9541**, the `processors.ts` splits.

#9541 is deliberately not in this PR. Extracting ~9 command handlers behind a shared dispatch, three advisory spend gates behind one precondition struct, and the plan-and-execute pass into its own typed-context module is a large behaviour-preserving refactor of a 16.2k-line file. Bundling it with behavioural fixes would make both unreviewable, and the risk profiles are entirely different.

Both fixes here *reduce* that refactor's risk: the **required** `decisionNowMs` field is a first instance of exactly the "typed decision-pass context both paths must construct" #9541 calls for, and the reachability check makes a module accidentally orphaned during a split fail CI instead of silently dying.

## Validation

- `npx tsc --noEmit`, `dead-source-files:check`, `db:migrations:check`, `selfhost:env-reference:check`, `actionlint`, `git diff --check` — all clean
- **264 passed** across the guardrail, account-age, clock-threading, checker, merge-block, MCP write-tools, soft-claim and queue-3 suites
- Patch coverage against this diff: **0 uncovered changed lines**

**Regressions, verified to fail against the unfixed code:** the velocity exception evaluated at a supplied instant diverges from the live clock (using a *confirmed miner*, the only configuration where the gap is decisive — an earlier draft of this test passed for the wrong reason and was fixed); the repeat lookback shares that instant; account age judged at a supplied instant flips `isNewAccount`; and the checker flags both the "only its own test imports it" and "no importer at all" shapes.

**Invariants:** omitting `nowMs` preserves live-clock behaviour byte-identically at every new seam, so the callers outside the decision pass are untouched; one production importer is always enough; a *non*-own test still counts as a real importer (a shared helper is alive); `scripts/**` rescues a module that is otherwise reported dead; declared entry points are never flagged; barrels keep their members alive via `export ... from`; specifiers reaching outside `src/**` are ignored rather than miscounted; and the pass still captures exactly one clock instant.

The two `processors.ts`-internal sites have no export seam, so they are pinned by a **source-parity guard** — the same convention `db-parsers.test.ts` already uses for `STALE_RECHECK_DENIAL_DETAIL_PATTERN`, and the honest alternative to exporting internals purely for a test.


JSONbored:
CI failure fixed and pushed.

`loopover_post_soft_claim` — the tool this PR wires up after finding it shipped-but-unregistered — needed an entry in `MCP_TOOL_CATEGORIES`. The `mcp-tool-categories` drift guard caught it, which is the guard working exactly as intended: a newly registered tool with no category is precisely what it exists to block.

That is a fair hit on me. The irony is not lost — this PR adds a reachability check for *silently dead* code, and I introduced a *silently uncategorized* tool doing it. The lesson is the same one #9540 encodes: a registry with no compile-time totality (this map is a plain `Record`, not a total `Record<ToolName, Category>`) drifts the moment a producer is added elsewhere. `feature-activation.ts`'s total-`Record`-over-a-key-union pattern is the fix for that class, and it is worth applying to this map — noted in **#9541**, since it belongs with the structural work rather than bolted onto this PR.

I also checked whether the stdio MCP server (`packages/loopover-mcp/bin/loopover-mcp.ts`) needed the same entry. It maintains its own independent tool list, and the full CI suite flagged only the categories map, so no parity check spans the two servers — that hand-mirroring is exactly what **#9537** tracks. Left alone deliberately rather than widened here.

Rebased onto current `main`; `mcp-tool-categories` and both write-tools suites green.

Issue #9543, opened 2026-07-28
Runtime credential rotation for subscription CLI providers (no restart, single-box + fleet)


JSONbored:
## Problem

Rotating the subscription-CLI credentials on an ORB review host (`CLAUDE_CODE_OAUTH_TOKEN`, and the codex `auth.json`) is a manual, multi-step, footgun-laden process. A live rotation on the hosted box surfaced two failure modes, both of which are **silent** — the container stays healthy and the status stays green while reviews quietly degrade:

1. **Content shape.** `loadFileSecrets` (`src/selfhost/load-file-secrets.ts`) only `.trim()`s the file. It does not strip comments or select a line. A human-added label line above the value (`# some-account`) becomes *part of the credential*. With `AI_PROVIDER=claude-code,ollama` the result is a failed auth that silently falls through to the Ollama fallback — no error surfaced, reviews just get worse.

2. **Inode pinning.** A Compose `secrets:` entry is a plain bind mount pinned to the **inode**. Writing the file *in place* propagates to the running container immediately; writing via write-new-then-rename (`mv`, and the default save behaviour of several editors) leaves the container serving the **old** content. `docker compose up -d` does not fix this — it reports `Container Running` and changes nothing, because the Compose config is unchanged. Only `--force-recreate` re-establishes the mount.

On top of that, the rotation currently *requires* a container recreate at all, which drops in-flight reviews for a credential swap that should be a no-op.

## Why a restart shouldn't be needed

The two providers are structurally different, and one is already doing the right thing:

| | Claude | Codex |
|---|---|---|
| Storage | docker secret file → env var, materialised **once at boot** | `auth.json` in the `loopover-data` volume |
| Read | `parentEnv.CLAUDE_CODE_OAUTH_TOKEN` inside `run()` (`src/selfhost/ai.ts`) | path re-resolved per call (`resolveCodexAuthPath`) |
| Restart to rotate? | Yes | **No** |

Claude is *nearly* hot already: `src/server.ts` passes the **live** `process.env` object into `createSelfHostAi`, and the token is read inside `run()` per call, not captured at construction. The only missing piece is that nothing ever re-reads the file after boot.

## Scope

Both deployment shapes need to be covered — a single self-hosted box and a multi-instance fleet.

### 1. Call-time credential resolution

Resolve the Claude token at call time with an explicit precedence:

1. **Fleet** — a DB-backed, encrypted-at-rest credential, injected via a Workers-safe registry slot.
[...]

PR #9550, opened 2026-07-28
feat(observability): distinguish a rejected credential from an exhausted quota

### LINKED ISSUE #9549: Distinguish a rejected credential from an exhausted quota in provider-failure signals
## Problem

A provider auth failure and a provider quota exhaustion are the same shape in every surface we expose:

- **Metrics:** `loopover_ai_provider_failures_total{provider}` has no reason dimension.
- **Logs:** `selfhost_ai_provider_failed` carries the raw `error` string only.

So `claude_code_error_401` and `claude_code_error_429` are indistinguishable without grepping raw logs and decoding the status by hand. The two need opposite first responses:

| | 401 / 403 | 429 |
|---|---|---|
| Meaning | credential rejected | credential fine, plan limit spent |
| Fix | rotate the credential | wait for the window, or raise the limit |
| Self-heals? | **No** — degrades every review until someone acts | Yes |

This is not hypothetical. A burst of `claude_code_error_429` on the hosted ORB box read as "the token is dead" and prompted a credential rotation that could not possibly have helped — the credential was valid the whole time.

## Scope

- Classify a provider failure into an actionable reason (`credential_invalid`, `quota_exhausted`, `timeout`, `not_configured`, `other`), matching on the structured error shapes the module itself throws rather than free-text provider prose.
- Expose it as a `reason` label on a **new** counter (not a new label on the existing failures counter, which shipped alert rules and dashboards already query) and as a `reason` field on the failure log.
- Two alert rules with the severity split that encodes the difference: a rejected credential pages, an exhausted quota warns, and each runbook says explicitly whether rotating helps.

## Notes

Anything unrecognised must stay `other`. A confidently wrong label sends an operator to the wrong runbook, which is worse than an honest unknown.

Follow-up to #9543.

JSONbored:
## Summary

A provider `401` and a provider `429` need opposite first responses, but every surface we expose rendered them identically — `loopover_ai_provider_failures_total` has no reason dimension, and `selfhost_ai_provider_failed` carries only the raw error string. Decoding which one you're looking at meant grepping logs for a status code.

That cost us real time: a burst of `claude_code_error_429` on the hosted ORB box read as a dead token and prompted a credential rotation. The credential was valid the whole time — a 429 is a quota signal, and a new token on the same account behaves identically.

`classifyProviderFailure` maps a failure to `credential_invalid` / `quota_exhausted` / `timeout` / `not_configured` / `other`. It matches on the **structured error shapes this module itself throws** (`claude_code_error_<status>`, the auth/timeout sentinels) rather than free-text provider prose, which is not a stable contract. The HTTP providers, which report auth failures as prose rather than an `error_NNN` code, get a separate narrower arm.

Anything unrecognised stays `other` — a confidently wrong label sends an operator to the wrong runbook, which is worse than an honest unknown.

## Two deliberate choices

**A new counter, not a new label.** `loopover_ai_provider_failures_total` is already referenced by shipped alert rules and dashboards. Adding a `reason` label would silently multiply it into one series per reason and change what those existing queries return. `loopover_ai_provider_failure_reason_total{provider,reason}` is purely additive and breaks nothing.

**Severity encodes the difference.** `LoopoverAiProviderCredentialInvalid` is `critical` — it does not self-heal, and its runbook points straight at the rotation path from #9543 (noting no restart is needed). `LoopoverAiProviderQuotaExhausted` is `warning` with a longer `for:`, since a short burst against a rate limit is normal and self-correcting, and its runbook opens by saying explicitly **not** to rotate.

## Validation

- `npm run typecheck` — clean (0 errors)
- `npx vitest run test/unit/selfhost-ai.test.ts` — **216 tests pass**
- `npm run selfhost:validate-observability` — dashboards and alert rules valid

Tests cover every classification arm and both sides of the branches, including: each status code, the prose-based HTTP-provider arm, non-`Error` and nullish rejections, and the `other` fallback. Three pin behaviour that matters specifically:

- **Regression:** a 429 classifies as `quota_exhausted`, not as a credential problem — the exact misreading this closes.
- **Invariant:** an unrecognised failure stays `other` rather than being mislabelled.
- **No false positive:** `"failed after 401 bytes of output"` classifies as `other`. The status pattern requires the `error_` prefix, so a PR number or byte count can never masquerade as an HTTP status.

Plus an assertion that the credential never leaks into the new metric's labels.

Closes #9549. Follow-up to #9543.

### DIFF
--- a/prometheus/rules/alerts.yml
+++ b/prometheus/rules/alerts.yml
@@ -604,6 +604,37 @@ groups:
           description: "Provider {{ $labels.provider }} has failed repeatedly and its circuit breaker is skipping calls fast during its cooldown (sustained 5m)."
           runbook: "Check that provider's credentials/reachability (CLI auth for claude-code/codex, or the configured API key/base URL for HTTP providers) via loopover_ai_provider_failures_total{provider=\"...\"} and recent selfhost_ai_provider_failed logs."
 
+      # The next two rules exist to separate the only two provider failures whose FIRST RESPONSE differs
+      # (#9549). Both previously surfaced as an opaque `claude_code_error_NNN` under the same counter, so a
+      # quota exhaustion looked exactly like a dead credential -- and rotating a perfectly good token is the
+      # natural, wasted next move. Distinct severities encode the difference: one needs a human NOW, the
+      # other resolves itself when the window rolls over.
+      - alert: LoopoverAiProviderCredentialInvalid
+        # A 401/403 from a provider: the credential is rejected outright. This does NOT self-heal -- every
+        # review degrades to the fallback provider (or fails) until someone rotates it. Any occurrence in
+        # 15m is actionable; `for: 5m` only filters a single blip during an in-flight rotation.
+        expr: increase(loopover_ai_provider_failure_reason_total{reason="credential_invalid"}[15m]) > 0
+        for: 5m
+        labels:
+          severity: critical
+        annotations:
+          summary: "loopover AI provider {{ $labels.provider }} credential is being rejected"
+          description: "Provider {{ $labels.provider }} returned {{ $value | printf \"%.0f\" }} auth failure(s) in 15m (sustained 5m). The credential is invalid or expired — this will not recover on its own."
+          runbook: "Rotate the credential: `./scripts/rotate-secret.sh claude_code_oauth_token` (or the loopover_admin_rotate_secret MCP tool). No restart is needed — the token is re-read per AI call. Confirm recovery via loopover_ai_provider_failure_reason_total{reason=\"credential_invalid\"} going flat."
+
+      - alert: LoopoverAiProviderQuotaExhausted
+        # A 429: the credential is fine, the plan's limit is spent. Rotating it changes NOTHING, so this is a
+        # warning and the runbook says so explicitly. Sustained 15m rather than any-occurrence, because a
+        # short burst against a rate limit is normal and self-correcting.
+        expr: increase(loopover_ai_provider_failure_reason_total{reason="quota_exhausted"}[15m]) > 0
+        for: 15m
+        labels:
+          severity: warning
+        annotations:
+          summary: "loopover AI provider {{ $labels.provider }} is out of quota"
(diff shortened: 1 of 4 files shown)


JSONbored:
Conflicts resolved and both CI failures fixed; rebased onto current `main`.

**1. Unregistered metric.** `loopover_ai_provider_failure_reason_total` was emitted and referenced by two alert rules but never added to `DEFAULT_METRIC_META`, so the metrics-drift guard failed. Registered, with help text stating why it exists as a separate counter rather than a new label on `loopover_ai_provider_failures_total` (that series is already in shipped alerts and dashboards — adding a label would silently change what those queries return).

**2. The runbook pointed at tooling that does not exist.** The `LoopoverAiProviderCredentialInvalid` runbook told an on-call operator to run `./scripts/rotate-secret.sh` and mentioned a `loopover_admin_rotate_secret` MCP tool. I checked: **neither exists** — not in the repo, not on `edge-nl-01`, and not in `MCP_TOOL_CATEGORIES`. The metric-name checker only caught it because the tool name happens to be `loopover_`-prefixed, but the real defect is worse than a lint failure: a runbook that sends someone to a nonexistent script during a credential outage is worse than no runbook.

**A correction I had to make to my own fix.** My first rewrite said a `docker compose up -d --force-recreate` was required, because a bind-mounted secret file rewritten in place leaves the process holding the stale inode. That *was* true — but **#9543 landed on `main` in the meantime**, and `resolveClaudeOauthToken` now re-reads `CLAUDE_CODE_OAUTH_TOKEN_FILE` on every AI call, with an empty/unreadable file falling back to the last known good value rather than dropping a working credential mid-rewrite. So the original annotation's "no restart is needed" claim was right, and my correction would have shipped an unnecessary-restart instruction. The runbook now states the real procedure and cites #9543 for why no restart is needed. Caught this because the rebase conflict was *in* #9543's new tests — the conflict was the signal.

**Conflicts.** `test/unit/selfhost-ai.test.ts` — both sides append independent `describe` blocks; kept both. The conflict region had swallowed the closing braces of #9543's block, so `tsc` failed until those were restored; all 227 tests in that file now run (previously the file would not parse).

284 tests green across `selfhost-ai`, `alerts-metric-name-references`, and `selfhost-metrics`; typecheck clean.

Issue #9432, opened 2026-07-27
AI review intermittently publishes no narrative summary (~40% of reviews) — root cause unknown


JSONbored:
## Summary

Some AI reviews publish a real narrative summary; others publish the generic placeholder ("The AI review returned non-blocking notes for this change but did not include a separate narrative summary") or "AI review is unavailable for this PR head" — under what appears to be identical configuration. Root cause is **not** established.

## Evidence (2026-07-27, all after `AI_MAX_OUTPUT_TOKENS` was restored to 4096)

| PR | cached summary state | time (UTC) |
|---|---|---|
| loopover#9426 | REAL_SUMMARY | 18:21 |
| awesome-claude#5599 | **UNAVAILABLE** | 18:20 |
| loopover#9423 | REAL_SUMMARY | 18:03 |
| loopover#9417 | **FALLBACK_PLACEHOLDER** | 18:01 |
| loopover#9420 | REAL_SUMMARY | 17:58 |
| loopover#9415 | REAL_SUMMARY | 17:56 |
| loopover#9422 | **FALLBACK_PLACEHOLDER** | 17:55 |
| loopover#9421 | **FALLBACK_PLACEHOLDER** | 17:49 |

Roughly 40% degraded. Query used:

```sql
SELECT repo_full_name, pull_number, created_at,
  CASE WHEN notes LIKE 'AI review is unavailable%' THEN 'UNAVAILABLE'
       WHEN notes LIKE 'The AI review returned%did not include a separate narrative%' THEN 'FALLBACK_PLACEHOLDER'
       ELSE 'REAL_SUMMARY' END AS summary_state
FROM ai_review_cache WHERE created_at > '2026-07-27T17:37:00Z' ORDER BY created_at DESC;
```

## Ruled out

- **Provider auth** — `claude` CLI in the container works when invoked as the app does (`CLAUDE_CODE_OAUTH_TOKEN` sourced from `/run/secrets/claude_code_oauth_token` → returns `OK`). A bare `docker exec claude -p ...` reports "Not logged in", but that is a test artifact: `loadFileSecrets` populates `process.env` inside the Node process at boot and `docker exec` gets a fresh shell.
[...]

PR #9556, opened 2026-07-28
fix(gate,review,ops): add glob exclusions, tier the public-comment vocabulary, and make inert config queryable (#9554, #9555, #9433)

### LINKED ISSUE #9433: Config-dependent fixes can ship inert: audit env-gated behaviour for boot-time validation
## Summary

Several production-behaviour fixes are gated on server env vars that were never set, so the shipped code path stayed **inert** and the deployment silently behaved exactly like the pre-fix build. No unit test can catch this class: tests verify the mechanism works *when enabled*, never that an operator set the variable.

This is a process/observability gap, not a single bug. It has already caused at least one "fixed" bug to stay live for weeks.

## Confirmed instance: bare-`score` public-comment filter

`LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS` was **unset** on `edge-nl-01`. It fails closed, so `BARE_SCORE_TERM_PATTERN = /\bscore\w*\b/i` was enforced for every repo.

`sanitizePublicComment` **throws** on a match and `toPublicSafe` catches → returns `null`. It is all-or-nothing, so any AI review narrative containing **score / scores / scoring / scored** had its **entire summary discarded** and replaced by the generic placeholder. Nits survived because they are filtered individually.

The `composeAdvisoryNotes` doc comment already documents this exact failure:

> "Metagraphed#8038-class bug: without this, ANY review of code with a legitimately-public `score`-named field ... had its entire narrative assessment silently discarded in favor of the generic 'did not include a separate narrative summary' fallback -- observed live, recurring."

The `allowBareScoreTerm` mechanism was built to fix it and is correctly wired at `ai-review.ts` (`allowBareScoreTerm: isPublicScoreTermSafeForRepo(env, input.repoFullName)`) — but with the env var unset it never engaged. **The fix shipped and was inert.**

Mitigated by setting the allowlist to the three gated repos. A boot-time warning (`selfhost_public_score_terms_allowlist_unset`) is added in the linked PR so this specific one can't silently recur.

## The general problem

Other config-dependent safety/behaviour paths have the same shape and no equivalent signal. Worth auditing for:

- fail-closed defaults where "unset" silently disables a fix rather than failing loudly
- env vars whose absence changes *review output* or *gate disposition* (not just performance)
- anything whose doc comment says a bug was "observed live, recurring"

## Proposed work

1. Audit `src/env.d.ts` for vars whose unset state silently disables a shipped fix, and classify each: hard-fail at boot / warn at boot / genuinely optional.
2. Extend `assertSelfHostPreflight` (or the boot warning block in `server.ts`) to cover the ones that change output correctness.
3. Consider surfacing "inert config" state on `/health` or the ops dashboard so it is visible without reading boot logs.

## Precedent to follow

`shouldWarnRagEmbedUnavailable` (`src/selfhost/ai.ts` → shouted in `src/server.ts`) is the existing pattern: a pure predicate plus a single structured boot-time `console.error`, warn-only so a legitimate configuration is not blocked. The new `shouldWarnPublicScoreTermsAllowlistUnset` mirrors it exactly.

### LINKED ISSUE #9554: gate(visual): the path matcher is positive-only, so a gate cannot exclude a generated artifact from an otherwise-visual directory
Parent: #9434

## Summary

`matchesAny` (`packages/loopover-engine/src/signals/change-guardrail.ts`) is **positive-only**: no `!` negation, and `{a,b}` braces are regex-escaped to literals. There is no way to express "this directory, except these generated artifacts".

That gap is what let `apps/loopover-ui/public/**` sit in a `screenshotTableGate.whenPaths` list. The directory is overwhelmingly **non-visual** — a generated 536KB `openapi.json`, `robots.txt`, `sitemap.xml`, `sw.js`, `manifest.webmanifest`, favicons — and `CLAUDE.md` **requires** contributors to run `npm run ui:openapi` on any API/schema change, which regenerates exactly that file. Following the documented contribution steps therefore tripped a gate demanding a Desktop/Tablet/Mobile × Dark before/after screenshot matrix **of a JSON spec file**. With `action: close` and the gate being one-shot, 5 contributor PRs were auto-closed with no recovery path (#9414, #9411, #9410, #9409, #9405).

The only available workaround was to drop the whole directory from scope — losing genuine coverage of the visual assets that *do* live there.

## Deliverables

- [ ] Add exclusion support so a gate can say "this directory except these generated artifacts" without enumerating every visual extension.
- [ ] It must be **additive**: `matchesAny` is also the hard-guardrail matcher, where an unrecognized glob failing *toward* matching is the safety-correct default. Changing `matchesAny` itself would mean a maintainer's literal path beginning with `!` is silently reinterpreted as an exclusion rather than guarded — so exclusion semantics must be opt-in for callers that want them.
- [ ] Wire it into the screenshot-table gate's `whenPaths` matching, in **both** places that read that list (`isScreenshotTableGateInScope` and `hasCommittedImageFile`). They must share one matcher — disagreeing on what "scoped" means would let a path excluded from gate scope still count as a stray committed image.

## The fail-direction requirement (the subtle part)

The include and exclude halves need **opposite** fail directions for an over-complex (ReDoS-capped) glob:

- **Include**: failing toward "matches" is safe — worst case, more paths are considered in scope. This is `matchesAny`'s existing behaviour and must be preserved.
- **Exclude**: failing toward "matches" is **wrong** — an over-complex exclude glob resolving to "matches everything" would silently *widen* what gets excluded, shrinking a safety gate's coverage. That is the opposite of what the gate exists for.

So the exclude half must fail toward excluding **nothing**, meaning gate coverage can only ever end up too *wide* from a malformed exclude glob, never too narrow.

## Tests

- [ ] The live shape: a directory glob plus one generated-file exclusion.
- [ ] With no exclusions, behaviour is identical to `matchesAny`.
- [ ] A path matching no include is false regardless of excludes; an all-exclude list is not an implicit wildcard include.
- [ ] Multiple excludes all apply.
- [ ] A bare `!` stays a literal include, not a malformed exclusion.
- [ ] **SECURITY**: an exclude glob is bound by the same wildcard cap as an include, resolves instantly against an adversarial multi-KB path, and fails toward excluding nothing.

## Also in scope: the sibling repos (#9434 follow-up 1)

The parent issue flagged that `JSONbored/metagraphed` and `JSONbored/awesome-claude` private configs were **not yet verified** for the same over-broad glob.

- [ ] Verify and fix them.

### LINKED ISSUE #9555: review(public-safety): ordinary English ('reviewability', 'ranking', 'reward') still drops narrative sentences on every repo
Parent: #9432

## Summary

#9445 fixed the catastrophic half of #9432 — the narrative assessment is now filtered **per sentence** (`toPublicSafeBySentence`) instead of all-or-nothing, so a single flagged word no longer discards the entire summary.

A residual remains. `FORBIDDEN_PUBLIC_COMMENT_WORDS` is matched with a plain case-insensitive `.includes()`, and it contains **ordinary English**:

`reward`, `rewards`, `ranking`, `rankings`, `cohort`, `farming`, `reviewability`

Only bare `score` has a per-repo escape hatch (`allowBareScoreTerm`, driven by `LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS`). Its siblings have none — so on **every** repo, allowlisted or not, a sentence like:

> "This change improves reviewability and updates the ranking comparator so ties resolve deterministically."

is still dropped from the published narrative. That is a perfectly safe review of this codebase's own gate/scoring code, and `reviewability` in particular is unavoidable code-review vocabulary.

The result is no longer a *missing* summary (that was #9445) but a **quietly incomplete** one — arguably harder to notice, because the output looks fine.

## Requirements

1. Extend the existing per-repo exemption to the ambiguous ordinary-English terms, using the **same** allowlist and the same positively-confirmed-per-repo requirement. No new config surface, no blanket default.
2. Terms that name a private concept outright must stay enforced for **every** repo, allowlisted or not.
3. Default behaviour for a non-allowlisted repo must be byte-identical to today.

## Why the split is safe (the load-bearing argument)

The risk this filter exists to stop is a leaked private **value**. A value never appears as a bare noun — it appears **qualified**:

- `reward estimate 12 TAO` → contains `reward estimate`
- `raw trust score 0.82` → contains `raw trust score`
- `score preview`, `estimated score`, `private ranking`, `wallet`, `hotkey`, `coldkey`, `seed phrase`, `mnemonic`, `payout`

Every one of those qualified forms belongs in the always-forbidden tier. A bare `reward` or `ranking` with no number attached leaks nothing on its own. This generalizes the judgment `allowBareScoreTerm` **already** made for `score` to its siblings rather than inventing a new one.

## Deliverables

- [ ] Split the vocabulary into `ALWAYS_FORBIDDEN_PUBLIC_COMMENT_WORDS` (never exemptible) and `AMBIGUOUS_PUBLIC_COMMENT_WORDS` (exemptible via the existing allowlist).
- [ ] Keep `FORBIDDEN_PUBLIC_COMMENT_WORDS` exported and derived from both, so existing consumers and the redaction-parity test keep one canonical vocabulary.
- [ ] Gate the ambiguous tier on the existing `allowBareScoreTerm` flag.

## Tests

- [ ] **SECURITY**, exhaustive over the real list: every always-forbidden phrase still throws even with the exemption on — asserted over the array itself, so a term moved into the wrong tier fails immediately rather than passing a hand-picked sample.
- [ ] **SECURITY**: a qualified private value stays blocked for an allowlisted repo even though its bare noun is exempt (`reward` exempt, `reward estimate 12 TAO` blocked).
- [ ] An allowlisted repo keeps a sentence using each ambiguous term.
- [ ] Without the exemption, the ambiguous tier is still enforced — the default is unchanged.
- [ ] The two tiers are disjoint and together exactly equal the combined list — no term silently lost or double-listed.

### LINKED ISSUE #9560: ci(turbo): three typecheck tasks lack a build edge to @loopover/contract, failing validate-code on any cache miss
## Summary

`validate-code` fails intermittently with errors in files the PR never touched:

```
../../src/mcp/server.ts(22,8): error TS2307: Cannot find module '@loopover/contract/tools' or its corresponding type declarations.
../../src/mcp/server.ts(2239,14): error TS7006: Parameter 'input' implicitly has an 'any' type.
...four more implicit-any errors, all downstream of the first
```

#9530 added `@loopover/contract`, and `src/mcp/server.ts` imports `@loopover/contract/tools`. Three turbo `typecheck` tasks pull that file into their compilation program with **no build edge to the package**:

| task | how it reaches `src/mcp/server.ts` | why `^build` does not cover it |
| --- | --- | --- |
| `@loopover/ui#typecheck` | `apps/loopover-ui/tsconfig.json` includes `$TURBO_ROOT$/worker-configuration.d.ts`, which imports `./src/index` — so the **entire Worker** is in the UI's program (`src/index.ts` → `src/api/routes.ts` → `src/mcp/server.ts`) | `apps/loopover-ui` has no `package.json` dependency on `@loopover/contract` |
| `//#typecheck` | root tsconfig includes `src/**` directly | a root (`//#`) task gets no build edge from a root `package.json` dependency |
| `@loopover/ui-miner#typecheck` | reaches `packages/loopover-miner/lib/**`, which imports the package | miner-ui has no dependency on it either |

The UI chain was confirmed with `tsc --explainFiles`, not inferred.

## Why it reads as flakiness

Turbo caches these tasks, so the failure only appears on a **cache miss**. PRs that hit the cache are green, which is why this presents as an unrelated PR "randomly" failing.

## Prior art — this exact class is already documented in the file

All three tasks already carry an explicit `@loopover/engine#build` edge, and `turbo.json`'s own comment on `@loopover/ui-miner#typecheck` describes the identical scheduling race for engine:

> Without it, turbo runs engine build CONCURRENTLY with this typecheck; on an engine-build cache miss the typecheck can race ahead of the dist emit and fail with phantom "Cannot find module '@loopover/engine'" errors (observed intermittently in CI's validate-code job [...]).

`@loopover/contract` reintroduced the same gap for a new package.

## Fix

Add `@loopover/contract#build` to all three tasks' `dependsOn`.

Verified both directions from a clean contract build state (`dist/` **and** `.tsbuildinfo` removed — turbo caches both, so removing only `dist/` produces a state turbo never creates, and `tsc` then silently no-ops):

- **with** the edges: 4 tasks successful, contract built first, typecheck passes
- **without** them: the identical six errors CI reported

JSONbored:
## Summary

Three defects that share one shape: a safety mechanism that is **correct in aggregate but too blunt at the edges**, silently destroying good content or hiding its own inertness.

Closes #9554
Closes #9555
Closes #9433

### #9554 — the path matcher cannot exclude, so a gate auto-closed PRs for a JSON file

`matchesAny` is positive-only: no `!` negation, and `{a,b}` braces compile to literals. So `apps/loopover-ui/public/**` could not carve out the generated `openapi.json` inside it — and `CLAUDE.md` *requires* contributors to regenerate that file on any API change. Following the documented contribution steps tripped a gate demanding a Desktop/Tablet/Mobile × Dark before/after screenshot matrix **of a JSON spec**, and with `action: close` being one-shot, 5 contributor PRs were auto-closed with no recovery path.

`matchesAnyWithExclusions` is **additive** — `matchesAny` is unchanged, because it is also the hard-guardrail matcher where a maintainer's literal path starting with `!` must stay guarded, not be silently reinterpreted.

The load-bearing subtlety is that the two halves need **opposite** fail directions for an over-complex (ReDoS-capped) glob. An unsafe *include* fails toward matching — safe, more paths in scope, and `matchesAny`'s existing deliberate behaviour. An unsafe *exclude* must fail toward excluding **nothing**: failing it toward "matches everything" would silently *widen* the exclusion and shrink gate coverage, the opposite of what a safety gate is for. My first implementation reused `matchesAny` for both halves and got this backwards; the security test caught it.

**Sibling repos, verified on the server (parent issue's follow-up 1).** `metagraphed` and `awesome-claude` both carried the *identical* over-broad glob — so this was never loopover-specific and both were live-exposed with `action: close`. Narrowed both to the actually-rendered UI, matching the fix already proven on loopover, with a comment recording why and backups taken first. No restart needed: `private-config.ts` reads these fresh per lookup.

### #9555 — ordinary English still drops narrative sentences

#9445 fixed the catastrophic half of #9432 (per-sentence filtering instead of all-or-nothing). The residual: `FORBIDDEN_PUBLIC_COMMENT_WORDS` is matched with a plain `.includes()` and contains ordinary English — `reward`, `ranking`, `cohort`, `farming`, **`reviewability`**. Only bare `score` had a per-repo escape hatch, so on *every* repo a sentence like "this improves reviewability and updates the ranking comparator" was still dropped. No longer a missing summary — a quietly incomplete one, which is harder to notice.

Split into `ALWAYS_FORBIDDEN_PUBLIC_COMMENT_WORDS` (never exemptible) and `AMBIGUOUS_PUBLIC_COMMENT_WORDS` (exemptible via the **existing** allowlist — no new config surface).

**Why the split is safe:** the risk is a leaked private *value*, and a value is never a bare noun — it is qualified (`reward estimate 12 TAO`, `raw trust score 0.82`, `score preview`). Every qualified form stays in the always-forbidden tier, enforced for every repo. A bare `reward` with no number leaks nothing. This generalizes the judgment `allowBareScoreTerm` already made for `score` rather than inventing a new one.

### #9433 — a report, not more boot warnings

The confirmed instance (an unset score-terms allowlist silently replacing every narrative containing "score") sat live for weeks with a green suite, because tests verify a mechanism *when enabled*, never that an operator set the variable.

**I did not follow the issue's obvious path of adding one boot warning per var, because verifying first showed it would be actively harmful.** `LOOPOVER_PUBLIC_STATS_REPOS` is unset on the ORB self-host box and that is **correct** — the public-stats surface is served by the Cloudflare Worker, where `wrangler.jsonc` sets it to four repos (confirmed live: `byProject` returns 3 entries). An unconditional warning would fire on every self-host deployment forever for a non-problem — the same alert fatigue that let a genuinely broken backup alert be ignored for 8 days.

What is actually missing is **answerability**: no way to ask "which config-gated behaviours are inert on this box?" without reading long-scrolled boot logs. `loopover_inert_config` answers exactly that, on demand, with zero steady-state noise and cardinality bounded by construction (the key set is a fixed list in code). Every entry is classified `deployment-specific` precisely so nothing here is auto-escalated to a warning.

Scope is deliberately narrow: only vars whose unset state changes **output correctness**. The ~100 `Default OFF` convergence flags are excluded — listing them would bury the few that matter.

## Validation

- `npx tsc --noEmit`, `selfhost:env-reference:check` (169 vars), `dead-source-files:check`, `actionlint`, `git diff --check` — all clean
- **558 passed** across guardrail, screenshot-gate, queue-intelligence, inert-config, metrics, ai-review and ai-summaries
- Patch coverage: **0 uncovered changed lines** (the one reported line is in `src/server.ts`, which is explicitly Codecov-`ignore`d as boot wiring covered by the Docker boot smoke test)

**Regressions:** the exclusion carves a generated file out of a matching directory glob, in both places that read `whenPaths`; an allowlisted repo keeps a sentence using each ambiguous term; the report names the confirmed #9433 instance.

**Invariants:** with no exclusions, behaviour is identical to `matchesAny`; an all-exclude list is not an implicit wildcard include; a bare `!` stays a literal include. **SECURITY**, asserted exhaustively over the real arrays so a term in the wrong tier fails immediately: every always-forbidden phrase still throws with the exemption on; a qualified private value stays blocked even though its bare noun is exempt; an exclude glob is bound by the same ReDoS cap and fails toward excluding nothing. The two tiers are disjoint and together exactly equal the combined list. Gauge samples and entries stay in lockstep, so three surfaces read one answer rather than three drifting lists.

Closes #9560

### DIFF
--- a/apps/loopover-ui/src/lib/selfhost-env-reference.ts
+++ b/apps/loopover-ui/src/lib/selfhost-env-reference.ts
@@ -286,6 +286,14 @@ export const SELFHOST_ENV_REFERENCE_ROWS: SelfHostEnvReferenceRow[] = [
     firstReference: "src/server.ts",
   },
   {
+    name: "LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS",
+    firstReference: "src/selfhost/inert-config.ts",
+  },
+  {
+    name: "LOOPOVER_PUBLIC_STATS_REPOS",
+    firstReference: "src/selfhost/inert-config.ts",
+  },
+  {
     name: "LOOPOVER_REPO_CONFIG_DIR",
     firstReference: "src/server.ts",
   },
@@ -296,6 +304,10 @@ export const SELFHOST_ENV_REFERENCE_ROWS: SelfHostEnvReferenceRow[] = [
   {
     name: "LOOPOVER_REVIEW_RAG",
     firstReference: "src/selfhost/ai.ts",
+  },
+  {
+    name: "LOOPOVER_REVIEW_SAFETY",
+    firstReference: "src/selfhost/inert-config.ts",
   },
   {
     name: "LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS",
@@ -744,9 +756,12 @@ export const SELFHOST_ENV_REFERENCE_MARKDOWN = [
   "| `LOOPOVER_ENABLE_UNSAFE_CODEX_REVIEWER` | `src/selfhost/ai.ts` |",
   "| `LOOPOVER_MCP_TOKEN` | `src/selfhost/preflight.ts` |",
   "| `LOOPOVER_METRICS_REPO_LABELS` | `src/server.ts` |",
+  "| `LOOPOVER_PUBLIC_SCORE_TERMS_ALLOWED_REPOS` | `src/selfhost/inert-config.ts` |",
+  "| `LOOPOVER_PUBLIC_STATS_REPOS` | `src/selfhost/inert-config.ts` |",
   "| `LOOPOVER_REPO_CONFIG_DIR` | `src/server.ts` |",
   "| `LOOPOVER_REVIEW_CONTINUOUS` | `src/queue/processors.ts` |",
   "| `LOOPOVER_REVIEW_RAG` | `src/selfhost/ai.ts` |",
+  "| `LOOPOVER_REVIEW_SAFETY` | `src/selfhost/inert-config.ts` |",
   "| `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS` | `src/server.ts` |",
   "| `LOOPOVER_SINGLE_INSTANCE` | `src/selfhost/redis-cache.ts` |",
   "| `LOOPOVER_VERSION` | `src/selfhost/otel.ts` |",
--- a/packages/loopover-engine/src/review/screenshot-table-gate.ts
+++ b/packages/loopover-engine/src/review/screenshot-table-gate.ts
@@ -1,4 +1,4 @@
-import { matchesAny } from "../signals/change-guardrail.js";
+import { matchesAnyWithExclusions } from "../signals/change-guardrail.js";
 import type { ScreenshotTableGateAction, ScreenshotTableGateConfig } from "../types/manifest-deps-types.js";
 
 export type { ScreenshotTableGateAction, ScreenshotTableGateConfig } from "../types/manifest-deps-types.js";
@@ -174,7 +174,12 @@ export function hasCommittedImageFile(changedFiles: string[], scopedPaths: strin
   return changedFiles.some((file) => {
     const lower = file.toLowerCase();
     if (!IMAGE_EXTENSIONS.some((ext) => lower.endsWith(ext))) return false;
-    return scopedPaths.length === 0 || matchesAny(file, scopedPaths);
+    // #9434: exclusion-aware, so an operator can scope "this directory except these generated files"
+    // (e.g. "apps/loopover-ui/public/**" minus "!apps/loopover-ui/public/openapi.json") instead of being
+    // forced to enumerate every safe subpath. Must stay the SAME matcher isScreenshotTableGateInScope uses
+    // below -- both read config.whenPaths, and disagreeing on which paths count as "scoped" would make a
+    // path excluded from gate SCOPE still count as a stray committed image, or vice versa.
+    return scopedPaths.length === 0 || matchesAnyWithExclusions(file, scopedPaths);
   });
 }
 
@@ -312,7 +317,8 @@ export function isScreenshotTableGateInScope(config: ScreenshotTableGateConfig, 
   if (config.whenLabels.length === 0 && config.whenPaths.length === 0) return true;
   const wantedLabels = new Set(config.whenLabels.map((label) => label.toLowerCase()));
   const labelMatch = config.whenLabels.length > 0 && prLabels.some((label) => wantedLabels.has(label.toLowerCase()));
-  const pathMatch = config.whenPaths.length > 0 && changedFiles.some((file) => matchesAny(file, config.whenPaths));
+  // #9434: exclusion-aware -- see hasCommittedImageFile's own comment on why the two must share one matcher.
+  const pathMatch = config.whenPaths.length > 0 && changedFiles.some((file) => matchesAnyWithExclusions(file, config.whenPaths));
   return labelMatch || pathMatch;
 }
 
--- a/packages/loopover-engine/src/signals/change-guardrail.ts
+++ b/packages/loopover-engine/src/signals/change-guardrail.ts
@@ -117,6 +117,47 @@ export function matchesAny(path: string, globs: string[]): boolean {
 }
 
 /**
+ * #9434: `matchesAny` is deliberately positive-only — there is no way to say "this directory, except these
+ * generated files" without enumerating every safe subpath by hand. That gap is exactly what let
+ * `apps/loopover-ui/public/**` (a directory that is overwhelmingly non-visual: a generated openapi.json,
+ * robots.txt, sitemap.xml, favicons) sit in a screenshotTableGate `whenPaths` list and auto-close 5
+ * contributor PRs one-shot for regenerating a JSON spec file, with zero recovery — a one-shot close has none.
+ * Confirmed the identical glob shape independently present in two sibling repos' private configs.
+ *
+ * A glob prefixed with `!` here is an EXCLUSION: a path counts as matched only if it hits at least one
+ * INCLUDE glob and hits NO exclude glob. This is deliberately a SEPARATE function, not a change to
+ * `matchesAny` itself — `matchesAny` is also the hard-guardrail matcher, where an unrecognized/malformed
+ * glob shape failing towards "still matches" is the safety-correct default (see its own doc comment); adding
+ * `!`-parsing there would mean a maintainer's literal path starting with `!` (rare, but not impossible) is
+ * silently reinterpreted as an exclusion instead of guarding it. This function is opt-in for callers that
+ * explicitly want exclusion semantics (today: the screenshotTableGate `whenPaths` scope check) and leaves
+ * every existing `matchesAny` caller — including hardGuardrailGlobs — byte-identical.
+ *
+ * The INCLUDE half reuses `matchesAny` unchanged, so an over-complex include glob keeps its existing
+ * fail-toward-matching default. The EXCLUDE half deliberately does NOT reuse `matchesAny` for this: failing
+ * an unsafe glob toward "matches" is safe for an include (worst case, more paths are considered in scope) but
+ * WRONG for an exclude (an over-complex exclude glob resolving to "matches everything" would silently widen
+ * what gets excluded, shrinking a safety gate's coverage — the opposite failure direction from what the gate
+ * exists for). The exclude half therefore compiles each glob with `globToRegExp` directly, whose own
+ * NEVER_MATCHES fallback for an unsafe glob is exactly right here: a malformed/pathological exclude excludes
+ * NOTHING rather than excluding everything, so gate coverage can only ever be too WIDE from a bad exclude
+ * glob, never too narrow.
+ */
+export function matchesAnyWithExclusions(path: string, globs: string[]): boolean {
+  const includes: string[] = [];
+  const excludes: string[] = [];
+  for (const glob of globs) {
+    if (glob.startsWith("!") && glob.length > 1) excludes.push(glob.slice(1));
+    else includes.push(glob);
+  }
+  if (includes.length === 0) return false;
+  if (!matchesAny(path, includes)) return false;
+  if (excludes.length === 0) return true;
+  const canonicalPath = canonicalize(path);
+  return !excludes.some((exclude) => globToRegExp(exclude).test(canonicalPath));
+}
+
+/**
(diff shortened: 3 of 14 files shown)


JSONbored:
Pushed additional coverage for the two engine files Codecov flagged.

**Diagnosis.** Codecov ran on my current head (`29ed46a`), so the report was real rather than stale. Digging into the v8 report, my three actual code lines in `screenshot-table-gate.ts` were all hit (20 hits each) and v8 emits **no entries for comment lines** — lines 177–181 are simply absent from the statement map. So the shortfall was not uncovered logic.

The likely cause is **shard attribution**: `packages/loopover-engine/src/**` and its `src/**` re-export shim are two *distinct* coverage identities, and CI shards the suites — so a shard running only one of the two test files attributes the lines under only one identity. `vitest.config.ts`'s own coverage-include comment documents this exact class from the 2026-07-24 codecov/patch (0%/99%) incident.

**Fix.** The exclusion behaviour is now exercised from **both** import paths rather than one:
- `screenshot-table-gate.test.ts` (the `src/` shim) gained the `whenPaths` exclusion cases, including that scope-matching and the committed-image check cannot disagree about what counts as scoped.
- `screenshot-table-gate-engine.test.ts` (the engine source) gained direct `matchesAnyWithExclusions` cases, including the security one — an over-complex exclude glob excludes *nothing*, so it can only widen gate scope, never shrink it.

Not redundant duplication: each path is its own real file and its own coverage identity, so whichever shard picks up either suite, the lines stay attributed.

237 tests green across the three suites; typecheck clean; both engine files now report zero uncovered changed lines locally.

Issue #8608, opened 2026-07-25
Bump eslint to v10 to resolve brace-expansion/minimatch DoS chain (GHSA-mh99-v99m-4gvg)


JSONbored:
## Context

Split out of #8588 (task 2 of 4) — the remaining `npm audit` findings after #8588's postcss+tar fix are all one
transitive chain: `brace-expansion` (high, DoS via unbounded expansion length, [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg))
pulled in via `minimatch` -> `@eslint/config-array`/`@eslint/eslintrc` -> `eslint`. `npm audit fix --force` resolves it
but bumps `eslint` `10.8.0` — a semver-major jump — so it needs real verification, not a blind force-fix.

## Why this is its own issue

ESLint has a history of breaking flat-config changes across majors. A blind `--force` risks silently breaking
lint enforcement (either failing loudly in CI, or worse, silently linting less than before) across both the
root workspace and `apps/loopover-ui`.

## Task

- [ ] `npm audit fix --force` (or a manual `eslint` bump + matching `@typescript-eslint/*`/plugin bumps) to `eslint@10.8.0`.
- [ ] Review the eslint v9->v10 changelog for flat-config breaking changes relevant to this repo's `eslint.config.*`.
- [ ] Run the full lint suite (root + `ui:lint`) and confirm it still catches what it caught before (spot-check a
      few known lint rules still fire, not just that the command exits 0 — a broken config can silently no-op).
- [ ] Re-run `npm audit --audit-level=moderate` to confirm the brace-expansion/minimatch/eslint chain is gone.
- [ ] `npm run test:ci` green.

Not urgent/blocking.

PR #9596, opened 2026-07-28
fix(ci): cache every workspace's node_modules, not just the root and one app

JSONbored:
## The failure

Every PR's **UI lint** step is failing on `main` right now:

```
> @loopover/ui-miner@0.0.0 lint
> eslint .
Error: Cannot find module '@eslint/js'
Require stack:
- /home/runner/work/loopover/loopover/apps/loopover-miner-ui/eslint.config.ts
```

## Why

`setup-workspace`'s node_modules cache listed only `node_modules` and `apps/loopover-ui/node_modules`. npm decides **per dependency** whether it hoists to the root or nests under a workspace, and that decision moves whenever a version range changes. #8608's eslint-10 bump made npm nest `@eslint/js` under *both* UI apps — so on a cache **hit**, `apps/loopover-miner-ui/node_modules` was never restored (it isn't in the path list) and the config's very first import failed.

It reproduces on no developer machine: a real `npm ci` / `npm install` always creates those directories, so only the cache-hit path is broken. That's also why it slipped through #9589's own green run — that run was a cache *miss* (the lockfile had just changed), which installed everything for real.

## The fix

Glob `apps/*/node_modules` + `packages/*/node_modules` — the same workspace globs `package.json` declares — in both the restore and the save step, so the cache follows npm's layout automatically instead of re-breaking on the next hoist change. `review-enrichment` and `control-plane` are deliberately untouched: they are not workspaces, and they already have their own installs and their own cache steps.

Verified: `actionlint` and `lint:composite-actions` both clean. The real proof is this PR's own CI — a first run installs fresh and saves the widened cache; every subsequent PR then restores a complete tree.

### DIFF
--- a/.github/actions/setup-workspace/action.yml
+++ b/.github/actions/setup-workspace/action.yml
@@ -51,12 +51,23 @@ runs:
       with:
         path: |
           node_modules
-          apps/loopover-ui/node_modules
+          apps/*/node_modules
+          packages/*/node_modules
         # hashFiles('.nvmrc') matters as much as the package manifests and lockfile: a Node bump
         # with no lockfile change would otherwise still hit and silently reuse node_modules whose
         # native addons (sharp, workerd, fsevents) were compiled against the OLD Node's ABI. The
         # manifests matter too because npm ci validates package.json/package-lock.json consistency
         # and runs lifecycle scripts from package.json; a package.json-only change must not skip it.
+        #
+        # The path list must cover EVERY workspace's own node_modules, not just the root and one app.
+        # npm decides per dependency whether it hoists to the root or nests under a workspace, and that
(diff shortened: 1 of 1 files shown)


JSONbored:
Follow-up correction pushed: widening `path` alone was **not** enough. The cache key is computed from the manifests only, so a branch with an unchanged lockfile keeps *hitting* the entry archived under the narrow path list — which restores no `apps/loopover-miner-ui/node_modules`, and being a hit, skips `npm ci` entirely. The tree would have stayed exactly as broken.

Added an explicit `npm-v2-` generation salt to the key so one clean miss re-archives a complete tree under the new list, with a comment stating the salt must be bumped whenever `path` changes.

Issue #9522, opened 2026-07-28
orb(mcp): ORB management tool family — instance doctor/status/logs, queue+kill-switch ops, fleet + hosted-tenant administration


JSONbored:
### Context

Part of #9515 — the ORB half of "MCP as the management plane." The management *capabilities* mostly exist; they're just scattered across interfaces an agent can't reach uniformly:

- **MCP already (self-host only, 4 tools):** `loopover_admin_{get_config,write_config,list_config_backups,trigger_redeploy}` (`src/mcp/server.ts:3295-3336`), gated on `LOOPOVER_MCP_ADMIN_ENABLED` + `mcp-admin` actor, wired through the Workers-safe capability registries (`src/mcp/private-config-admin-registry.ts`, `redeploy-companion-registry.ts`) that only the self-host entry fills.
- **Operator HTTP routes with no MCP equivalent:** operator dashboard (`src/api/routes.ts:1945`), DLQ list/replay/delete/purge (`:1955/:1975/:2001/:2027`), kill-switch get/set (`:2054/:2065`), fleet config-push (`:2112`), bulk agent settings (`:2778`).
- **Fleet/internal routes with no MCP equivalent:** fleet analytics (`:4569`), instances list/register (`:4621/:4641`), installations list/register/backfill (`:4670/:4687/:4703`), enrollments issue/rotate/revoke (`:4721/:4740`), ops stats (`:4754`), ~30 `/v1/internal/jobs/*` (`:4886-5187`) — today driven only by the GitHub Actions fleet matrix (`.github/workflows/self-host-maintenance.yml`).
- **Hosted control plane (`control-plane/src/http-app.ts`):** tenant create/list/patch/delete (`:134/:200/:229/:268`); rollout is a 501 stub (#9143).
- **Confirmed absent everywhere:** instance status rollup (image/version/uptime/container health in one answer), bounded log tail, backup status, an ORB `doctor` (the miner has one; ORB does not — `doctor` in this repo is miner-only).

All new tools ride the standardized registry (#9518) with `auth`/`locality`/`availability` metadata, and the management routes they wrap are specced by #9519.

### Tool catalog (the deliverable — renames/cuts recorded on this issue, not improvised in PRs)

**Self-host instance ops** (availability `selfhost`, auth `mcp-admin`; capability-registry pattern — unset capability ⇒ structured "not configured" result):
[...]

Issue #9521, opened 2026-07-28
contract: generate every derived surface — typed stdio client, CLI from one spec table, tool-reference docs, miner dedup, UI z.infer types


JSONbored:
### Context

Part of #9515 — the "nothing hand-maintained" payoff issue. With the contract registry (#9517/#9518) and the route seam (#9519) in place, every remaining hand-kept surface becomes generatable. Today's inventory of hand-maintained duplicates:

- **Stdio API client:** `apiGet`/`apiPost`/`apiDelete` (`packages/loopover-mcp/bin/loopover-mcp.ts:6950-7002`) return `payload: any`; ~50 endpoints consumed via optional-chaining guesswork (`payload.pendingActions ?? []`, `result.predictedGate`, …). A Worker response rename fails silently at runtime.
- **Four independent CLI command lists:** `CLI_COMMAND_SPEC` (`:112-146`), the 34-branch `runCli` if-chain (`:4169-4207`), `printHelp` (`:5393-5447`), and the README's 47-line command block (`packages/loopover-mcp/README.md:29-76`) — only fragments are pin-tested; `printHelp` already omits commands.
- **CLI flag parsers re-implementing zod shapes:** `branchEligibilityFromOptions` (`:7337-7348`), `validationFromOptions`/`parseValidationEntry` (`:6351-6368`), `parseOptions`' hand-listed repeatable/boolean sets (`:5483+`).
- **No tool-reference doc anywhere:** the README dodges it ("run `loopover-mcp tools`"), 32 tool names are hand-typed across `apps/loopover-ui/content/docs/*.mdx` with zero drift check, and `scripts/check-docs-drift.ts` + `scripts/gen-command-reference.ts` have no MCP coverage at all. The miner README's tool table (`packages/loopover-miner/README.md:239-286`) and 18-row storage table (`:112-131`) are hand-kept prose (name-level parity test only).
- **Miner hand-copies:** `lib/github-token-resolution.ts` re-implements the stdio MCP's config/profile/apiUrl/session resolution by hand (self-declared, `:6-11`); `lib/tenant-credential-resolution.ts:2-3` duplicates `src/orb/broker-client.ts`.
- **UI:** #9282 (accepted decision: share zod schemas, `z.infer` on the UI side) — its pilot (`PublicStats`/`PublicRulePrecision`) plus the drift check belong to this group.
[...]

PR #9607, opened 2026-07-28
ams(mcp): the AMS management tool family — governor-gated mutations, doctor, tenant health (#9523)

### LINKED ISSUE #9523: ams(mcp): AMS management tool family — governor-gated mutations, doctor split, hosted tenant tools
### Context

Part of #9515 — the AMS half of "MCP as the management plane." The miner MCP (`packages/loopover-miner/bin/loopover-miner-mcp.ts`) exposes 11 read-only tools, while the entire mutating ops surface is CLI-only: governor pause/resume (`lib/governor-pause-cli.ts`), queue ops (`lib/portfolio-queue-cli.ts`: `next|claim-batch|done|release|requeue`), claim ops (`lib/claim-ledger-cli.ts`), run-state set (`lib/run-state-cli.ts`), deny-hooks approve/reject (`lib/deny-hooks-cli.ts`), calibration apply/revert (`lib/calibration-cli.ts`, double-gated), right-to-be-forgotten purge (`lib/purge-cli.ts`), store migrations (`lib/migrate-cli.ts`). The safe-dispatch chokepoint already exists: `lib/chat-action-registry.ts` + `chat-action-dispatch.ts` structurally refuse any handler that isn't `governorGatedHandler`-wrapped (built for the dashboard chat, #6521/#6838/#6839), and the miner dashboard already dispatches governor pause/resume + queue release/requeue through it (`apps/loopover-miner-ui` `vite-governor-api.ts:56-58`, `vite-portfolio-queue-actions-api.ts:92-94`).

Hosted AMS bones also exist: `control-plane/` tenant CRUD + cron wake (`ams-wake.ts`), `loopover-miner-hosted` entry with `/health`+`/ready` (`lib/ams-health-server.ts`), tenant CLI (`lib/tenant-cli.ts`) — and #9199 wants tenant-scoped, chat-groundable health tools.

The 11 existing tools gain contract registration + output schemas in #9518 (not here). This issue adds the **mutating and hosted tool families**.

### Tool catalog (the deliverable — renames/cuts recorded on this issue first)

**Local ops, read** (availability `selfhost`, locality `miner`): `loopover_miner_doctor` (dedicated tool — today doctor rides inside `loopover_miner_status`; split so `status` stays cheap and `doctor` can grow checks) · `loopover_miner_get_metrics_snapshot` (the Prometheus families already exported by the `metrics` CLI, as structured JSON).

**Local ops, mutating** (availability `selfhost`, all dispatched through the governor-gated chokepoint — a raw handler that bypasses `chat-action-registry` is defective by definition, mirroring `test/unit/no-direct-octokit.test.ts`'s structural rule):
`loopover_miner_governor_pause` / `loopover_miner_governor_resume` (mirror the dashboard actions; fire the same `ams-notifications` side-channel) · `loopover_miner_queue_release` / `loopover_miner_queue_requeue` (per-item, mirroring the dashboard's two queue actions) · `loopover_miner_claim_release` · `loopover_miner_deny_hooks_decide` (`approve|reject` on a pending synthesized hook) · `loopover_miner_run_migrations` (existing-stores-only, dry-run default like the CLI) · `loopover_miner_purge_repo` (**destructive**: right-to-be-forgotten; requires `confirm: z.literal(true)` + elicitation; returns the same per-store purge report as the CLI).

Deliberately **excluded** from MCP (recorded as decision): calibration `apply-min-rank`/`revert-min-rank` (stays CLI-only behind its existing double gate — an agent-reachable path to loosen calibration floors is not wanted), `state set` (raw run-state mutation is a repair tool, not an operation), kill-switch trip/untrip (one-way breaker semantics stay out of agent reach; pause/resume is the agent-safe control).

**Hosted tenants** (availability `cloud`, auth control-plane admin; served by the hosted/remote side, not the local bin): `loopover_ams_tenant_create` / `loopover_ams_tenant_list` / `loopover_ams_tenant_destroy` (**destructive**, confirm + elicitation) — wrapping `lib/tenant-client.ts`'s control-plane calls · `loopover_ams_tenant_health` (#9199: per-tenant health/status — wake schedule, last cycle outcomes, container health) · `loopover_ams_tenant_wake` (trigger an immediate cycle for one tenant via the `ams-wake` machinery, bounded by the same per-tenant schedule guards).

### Requirements

1. Every tool: `@loopover/contract` entry (real input/output schemas, `auth`/`locality`/`availability`/`destructiveHint`), registered per #9518, smoke-called in `validate:mcp` (#9520) with fakes for stores/control-plane.
2. **Structural governor gating:** the miner MCP's mutating tools register through `chat-action-registry` (or a shared registry both chat and MCP consume — preferred; record the shape in the PR), so the compile-time/structural "only governor-gated handlers" rule covers MCP identically. Add the structural test.
3. Mutating tools write the event ledger exactly as their CLI twins do (same event types), so `loopover_miner_get_audit_feed` shows MCP-initiated actions indistinguishably from CLI ones, with an `actor`/`source` field distinguishing origin.
4. **Tenant isolation for hosted tools** (per #9184's constraints): every hosted tool resolves tenant scope server-side from the authenticated identity; no cross-tenant reads/writes; deny-path tests for wrong-tenant access on every tool.
5. Privacy boundary unchanged: no wallet/hotkey/reward/trust-score data in any tool response (existing `FORBIDDEN_CONTENT` matchers extended to the new tools in the miner contract parity suite).
6. The dashboard's governor/queue endpoints and the new MCP tools call the same underlying action functions — one implementation, asserted by test.
7. Docs regenerate via #9521's generators (tool tables pick the new tools up automatically — no hand-edits).
8. Full branch-counted coverage; fix-what-you-find applies (the `docker-compose.miner.yml`/`k8s` `run`→`loop` defect is already being fixed separately — do not duplicate it here).

### Non-goals

- Registering output schemas for the existing 11 read tools (#9518 does that).
- The hosted AMS chat UI (#9184 consumes these tools).
- New governor policy semantics — MCP only reaches actions the governor machinery already allows.

### Deliverables

- [ ] Local mutating family live behind the structural governor gate, ledger-audited, elicitation+confirm on `purge_repo`
- [ ] `doctor` split + metrics snapshot tool
- [ ] Hosted tenant family incl. #9199's health tool and wake, tenant-isolated with deny-path tests
- [ ] Exclusion decisions above recorded (calibration/state-set/kill-switch stay CLI-only)
- [ ] Generated docs picked up the catalog with zero hand-edits

### Expected outcome

A dual-role operator (or the #9184 hosted chat) can pause/resume the governor, groom the portfolio queue, decide deny-hooks, run store migrations, execute a confirmed right-to-be-forgotten purge, and administer hosted tenants — all over MCP, all schema-validated, all governor-gated and ledger-audited, with the dangerous calibration/kill-switch levers deliberately unreachable by agents.

### References

Part of #9515. Blocked by #9518. Advances #9184 (action catalog) and #5230 (AMS cloud convergence); implements #9199's health-tool half. Relates to #5221/#5226 (fleet rollout/health design).

JSONbored:
Closes #9523. **Stacked on #9604** (`feat/orb-management-tools-9522`) — review that first; this branch contains its commits.

## What changed

The miner MCP exposed 11 read-only tools while its entire **mutating** ops surface was CLI-only. This adds 10: a dedicated `doctor` (split out of `status` so status stays cheap and doctor can grow checks), a structured `metrics_snapshot`, and eight mutations — governor pause/resume, queue release/requeue, claim release, deny-hook decide, store migrations, and a confirmed right-to-be-forgotten purge.

**Every mutation dispatches through the miner's existing governor-gated chat-action chokepoint** — the same boundary the dashboard's own actions use. The MCP layer never touches a store: it dispatches an action *name*, and the registry structurally refuses any handler not produced by `governorGatedHandler()`, whose brand is a private symbol a raw function cannot forge. That is what makes *"an MCP caller cannot reach a write path the dashboard could not"* a property of the code rather than of review discipline, and the structural test asserts it from both ends — including that a blocked gate never reaches the store operation.

## Three claims I had to correct against the real code

Each of these I'd written into the contract from the issue's description, then found the code disagreed:

- **The migrate CLI has no dry-run.** Applying a migration *is* opening the store. So the tool has no `apply` flag either, rather than advertising a safety mode that does not exist and could only be built as a second implementation of the migration walk.
- **The deny-hook store keys proposals by `(repo, id)`.** The decide tool takes the repo rather than scanning every repo's proposals to resolve a bare id.
- **`purgeRepoAcrossStores` is extracted from `runPurge`**, so the tool runs the CLI's own purge over the CLI's own target list — not a second implementation free to miss a store as targets are added.

`collectMinerPredictionMetrics` is likewise extracted in the engine: the Prometheus text renderer now formats those families, so the scrape and the JSON snapshot share one aggregation and cannot disagree about what a counter means. All 762 engine tests pass unchanged.

## validate:mcp earned its keep twice

It caught the two AMS tenant tools declared but never registered, and it caught `loopover_miner_run_migrations`/`_purge_repo` returning fields their output schemas did not declare — the `.shape` re-wrap that drops `looseObject`'s catchall, **the same -32602 class this epic already fixed once**. Both outputs now declare every field their handlers return.

## Catalog decisions, recorded rather than improvised

AMS tenant create/list/destroy are **deliberately absent**: #9522's `loopover_tenant_*` tools are product-parameterized and already serve `product: "ams"`, because the control plane's own routes are. A second AMS-only spelling would be two names for one capability — the rot that kept `loopover_fleet_get_analytics` out of #9522. Only `health` and `wake`, the pair with no ORB counterpart, are added.

The issue's recorded exclusions — calibration floors, raw run-state `set`, the one-way kill switch — stay CLI-only, **pinned by a test** so reversing that decision has to be deliberate rather than accidental.

## Validation

Typecheck clean; `validate:mcp` passes across all three servers. `chat-miner-ops-actions.ts` and `miner-ops-actions.ts` are at **100% statement and branch**, covering the full path (action name → gate → store call), the validator guard clauses, and the fail-closed cases. `paramsOf` lost its `?? {}` — every caller has a validator requiring an object that runs first, so the fallback had no reachable case.

Docs: the miner README's tool list is hand-edited here because #9521's generator is not merged yet; once #9590 lands it owns that table and the entries regenerate.

### DIFF
--- a/apps/loopover-ui/src/lib/mcp-tool-reference.ts
+++ b/apps/loopover-ui/src/lib/mcp-tool-reference.ts
@@ -109,6 +109,20 @@ export const MCP_TOOL_REFERENCE: readonly McpToolReferenceEntry[] = [
     "description": "Create a queued copilot-only LoopOver agent run. The agent plans and explains; it does not edit code or open PRs."
   },
   {
+    "name": "loopover_ams_tenant_health",
+    "category": "tenant",
+    "locality": "remote",
+    "availability": "cloud",
+    "description": "Operator only. One hosted AMS tenant's health: lifecycle state, its cron-wake cadence, when it last woke, that cycle's outcome, and container health. Read-only, and scoped server-side to the authenticated tenant — a name outside that scope is refused rather than answered."
+  },
+  {
+    "name": "loopover_ams_tenant_wake",
+    "category": "tenant",
+    "locality": "remote",
+    "availability": "cloud",
+    "description": "Operator only. Trigger an immediate cycle for one hosted AMS tenant, instead of waiting for its next scheduled wake. Bounded by the SAME per-tenant schedule guards the cron path obeys — a wake too soon after the last one is reported as throttled rather than forced through."
+  },
+  {
     "name": "loopover_apply_labels",
     "category": "agent",
     "locality": "remote",
@@ -697,6 +711,27 @@ export const MCP_TOOL_REFERENCE: readonly McpToolReferenceEntry[] = [
     "description": "Mark a contributor's own delivered notifications as read (clears the badge). Self-scoped; pass `ids` to clear specific notifications or omit to clear all."
   },
   {
+    "name": "loopover_miner_claim_release",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "Release this miner's claim on one issue, so the claim ledger no longer reserves it. Dispatches through the governor-gated chokepoint and is recorded in the event ledger with source=mcp."
+  },
+  {
+    "name": "loopover_miner_deny_hooks_decide",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "Approve or reject one synthesized deny-hook awaiting review. Approving puts it into force for future runs; rejecting discards it. Dispatches through the governor-gated chokepoint and is recorded in the event ledger with source=mcp."
+  },
+  {
+    "name": "loopover_miner_doctor",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "Read-only diagnostic checks for this AMS miner: state directory, engine version match, store reachability, credentials, and configuration. Split out of loopover_miner_status (#9523) so status stays cheap and doctor can grow checks. Every check runs and reports its own pass/warn/fail — nothing is mutated and nothing stops at the first failure."
+  },
+  {
     "name": "loopover_miner_get_audit_feed",
     "category": "agent",
     "locality": "miner",
@@ -725,6 +760,13 @@ export const MCP_TOOL_REFERENCE: readonly McpToolReferenceEntry[] = [
     "description": "Read-only manage-phase status: the per-managed-PR rows `loopover-miner manage status` reports (branch, CI state, gate verdict, outcome, last-polled-at, queue status/priority) plus the run-level portfolio view (one row per tracked repo: run state, updated-at, PR count). Joins the portfolio queue, the append-only event ledger, and run-state by reusing the existing collectManageStatus/collectRunPortfolio aggregators -- no new join logic. Read-only: never calls GitHub, never mutates local stores. Takes no arguments."
   },
   {
+    "name": "loopover_miner_get_metrics_snapshot",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "The same Prometheus metric families the `metrics` CLI exports, as structured JSON — so an agent can read them without parsing the text exposition format. Read-only."
+  },
+  {
     "name": "loopover_miner_get_plan",
     "category": "agent",
     "locality": "miner",
@@ -746,6 +788,20 @@ export const MCP_TOOL_REFERENCE: readonly McpToolReferenceEntry[] = [
     "description": "Read-only per-repo miner run-state (idle/discovering/planning/preparing). Pass repoFullName for a single repo (a null state means none has been recorded for it yet), or omit it to list every repo's state. The read-only analog of ORB's loopover_get_automation_state; adds no state-set or mutation capability."
   },
   {
+    "name": "loopover_miner_governor_pause",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "Pause this miner's governor: no new work is admitted until it resumes. Administrative control, not a content write — the same action the dashboard's pause button dispatches, through the same governor-gated chokepoint, firing the same notification side-channel. Recorded in the event ledger with source=mcp."
+  },
+  {
+    "name": "loopover_miner_governor_resume",
+    "category": "ops",
+    "locality": "miner",
+    "availability": "selfhost",
+    "description": "Resume this miner's governor after a pause, re-admitting work. The same action the dashboard's resume button dispatches, through the same governor-gated chokepoint. Recorded in the event ledger with source=mcp."
+  },
+  {
     "name": "loopover_miner_list_claims",
     "category": "agent",
     "locality": "miner",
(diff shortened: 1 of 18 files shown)


JSONbored:
Both blockers cite files that do not exist in this repo, and I could not reproduce either against the real code. Details, since I'd rather show the evidence than just assert it:

**1. `adminRotateSecretTool.ts:12` — no such file.** The tool lives in `packages/loopover-contract/src/tools/admin-config.ts`.

On the substance: it is correct that there is no `confirm`, but that is not a defect this PR introduced or should change.
- The tool is **pre-existing** (#9543). This PR *migrated it into the contract registry* — its input surface is byte-identical to what already shipped. Adding a required `confirm` here would be a **breaking behavior change to a live tool**, not a fix.
- It is gated by `auth: "mcp-admin"` (`LOOPOVER_MCP_ADMIN_TOKEN`), and only registered at all when `LOOPOVER_MCP_ADMIN_ENABLED` is truthy — it is one of the few conditionally-registered tools on the server.
- Rotation **backs the previous value up first** (`backupPath` in its output), so it is recoverable. That is the material difference from `purge_repo`, which is not, and which is exactly why purge carries the confirm and this does not.
- It already carries `destructiveHint: true`.

**2. `mineralPurgeRepoTool.ts:24` — no such file, and no such symbol.** `grep -rn "mineral"` across `packages/`, `src/`, and `test/` returns nothing. The real tool is `minerPurgeRepoTool` in `packages/loopover-contract/src/tools/miner-ops.ts`.

Both claims about it are wrong:
- It **does** require confirmation — `MinerPurgeRepoInput` is `{ repoFullName, confirm: DestructiveConfirm }`, where `DestructiveConfirm` is `z.literal(true)` (a plain boolean would let an omitted field read as false and pass). Two regression tests cover it: `confirm` absent and `confirm: false` are both rejected **before any dispatch reaches a store**.
- A repo with no rows is **not a silent failure**. `purgeRepoAcrossStores` returns `{ outcome: "purged", totalPurged: 0, stores: [...] }` — an explicit per-store breakdown. That is the CLI's own core, extracted in this PR precisely so both surfaces answer identically. For a right-to-be-forgotten operation, "there was nothing to forget" is a correct answer, not an error.

The two findings also appear to have been swapped: the one told it lacks `confirm` is the one that has it.

I have pushed real fixes for the two genuine problems in the meantime — the merge conflict (resolved against #9522, now merged) and patch coverage, which is now at **zero uncovered changed lines** across every file in the diff.

JSONbored:
Confirming the `runPurge` exit-code question from the review — it is not a behavior change.

`outcome` is computed from the *same* predicate the old code used, at the only site that constructs a `PurgeSummary` ([purge-cli.ts:311-314](https://github.com/JSONbored/loopover/blob/feat/ams-management-tools-9523/packages/loopover-miner/lib/purge-cli.ts#L311-L314)):

```ts
const hadError = perStoreResults.some((entry) => "error" in entry);
return { outcome: hadError ? "partial" : "purged", ... };
```

And `PurgeSummary.outcome` is typed `"purged" | "partial"`, so there is no third value for it to take — `summary.outcome === "partial"` is `hadError`, by construction and by the type. The extraction moved that predicate from `runPurge` into `purgeRepoAcrossStores` precisely so the MCP tool reports the same partial-failure state the CLI exits 2 on, rather than recomputing it.

The 2-on-partial contract stays pinned by the existing suite (`test/unit/miner-purge-cli.test.ts`), including "is audit-observable on a PARTIAL failure: reports what succeeded, flags the failed store, and exits 2" plus the per-store error cases at lines 563/576/602/626.

Issue #9676, opened 2026-07-29
Reversal signal is unavailable on the hosted runtime, so own-ledger accuracy is unmeasurable there


JSONbored:
## What's wrong

`loopover.ai/fairness` currently shows a headline **94.6%** decision accuracy while every by-repository row and every one of the 8 weekly-trend rows shows **100%** with `Reversed: 0`, and the per-rule table shows **62.4%** for two different rules. Four numbers, four different tables, four different definitions of "wrong" — and only one of them is measuring anything live.

Live payload (`GET https://api.loopover.ai/v1/public/stats`) confirms `totals.reversed: 0`, every `byProject[].accuracyPct: 100`, and every `accuracyTrend[].reversed: 0`.

## Root cause

**Every writer of ground-truth "the bot was wrong" evidence lives in the review-execution pipeline, which the hosted Worker acks-and-drops — while every reader of that evidence still runs there and renders its absence as perfection.**

`recordReversalSignals` is called from exactly one place (`src/queue/processors.ts:7039`, inside `handlePullRequestWebhookEvent`), reached only via the `github-webhook` job. That job type is in `REVIEW_EXECUTION_JOB_TYPES` (`src/selfhost/review-runtime.ts:4`), so on the hosted Worker `src/index.ts:61-75` logs `retired_review_job_ignored` and acks it. The Orb App webhook path that *does* still run writes only `orb_pr_outcomes` (`src/orb/outcomes.ts:9-32`) and has no reversal concept.

The repo already records the empirical result — `src/review/reversal-superseded.ts:3-4`: "verified in production: zero reversal events ever."

### Three separate defects fall out of this

**1. By-repository: lifetime denominator ÷ 90-day-pruned numerator.** `reviewed` counts `github_app.pr_public_surface_published`, the one event type exempt from retention (`src/db/retention.ts:14,213-216`), so it is lifetime and immortal. The `reversed` numerator reads `reversal_*` events from `audit_events`, pruned at 90 days (`src/db/retention.ts:25`). Even with a working signal, that ratio drifts toward 100% by construction.

[...]

Issue #8277, opened 2026-07-23
calibration: backfill the slop_gate_score corpus by deterministic replay over archived diffs


JSONbored:
## Problem

The slop knob (#8224) entered the registry report-only, but its corpus only accumulates while slop gating fires live — weeks to reach the 50/12 sample floors. The slop scorer is deterministic and in-repo, and we hold hundreds of historical PR diffs with human-labeled outcomes (the #8130/#8170 raw-context corpus). The evidence can be manufactured honestly today.

## Requirements

> ⚠️ Required pattern. Mirror the phase-1/phase-2 backfill discipline exactly (scripts/backfill-calibration-corpus*.ts): idempotent synthetic events with an explicit provenance marker, dry-run default, resume cursor, both stores (cloud D1 CLI path + --pg for the ORB Postgres), pacing if any GitHub fetches are needed (prefer none — replay archived diffs only).

- Pure core: recompute slop risk over each archived diff via the shipped deterministic scorer; emit synthetic `signal.rule_fired:slop_gate_score` events (`confidence = risk/100`, provenance-tagged) paired with override labels from realized outcomes — the same counterfactual "mapping a" framing the close-confidence backfill used, documented as such in the event metadata.
- Apply on both stores; verify the report-only proposals path and the knobs endpoint surface slop evidence afterward.
- Post the resulting first proposals (or the honest absence) on this issue.

## Deliverables

- [ ] Backfill script (pure core + thin IO) with 100%-branch unit coverage on the core
- [ ] Applied + verified on cloud D1 and the ORB Postgres
- [ ] Evidence summary posted here; if proposals clear, file the slop flip-to-live issue citing them

## Links & Resources

#8224 (the knob), #8157/#8170 (the backfill discipline), scripts/backfill-calibration-corpus-phase2-core.ts

## Boundaries

Report-only evidence generation — no knob movement, no live-mode change.

*maintainer-only — synthetic-evidence provenance is gate-adjacent.*

JSONbored:
## Applied + verified on both stores

`slop_replay_backfill_v1` ran over the full archived corpus (460 cases, manifest checksum `86c3b4b55cae…`) and applied to **cloud D1** (920 rows) and the **ORB Postgres** (460 fired + 460 override rows, verified by direct count). Rows upsert idempotently by deterministic id.

**Replay distribution** (labels: 287 confirmed / 173 reversed):

| band | cases |
| --- | --- |
| zero (0) | 165 |
[...]

Issue #8083, opened 2026-07-22
calibration: pure BacktestCase corpus builder from RuleFiredEvent/HumanOverrideEvent pairs


JSONbored:
## Context

`packages/loopover-engine/src/calibration/signal-tracking.ts` (shipped in #7982) defines `RuleFiredEvent`, `HumanOverrideEvent`, and `computeRulePrecision(ruleId, fired, overrides)`, which pairs fired events with overrides by matching `ruleId` (see the internal `overrideMatchesRule` helper) to compute an aggregate precision number. There is no equivalent that keeps each PAIRED case as an individual labeled record — which is what a backtest needs: a list of concrete "this rule fired against this target, and a human later said it was right/wrong" cases, replayable against a different candidate rule/classifier later (see the parent epic for why).

> ⚠️ Required pattern. This file must be pure TypeScript with no IO, no DB access, and no imports from `src/`, `packages/loopover-miner`, or any host adapter — only the existing `RuleFiredEvent`/`HumanOverrideEvent` types from `signal-tracking.ts`. A PR that adds any database, environment, or network access to this file does NOT satisfy this issue.

## Requirements

- Add a new file `packages/loopover-engine/src/calibration/backtest-corpus.ts`.
- Export a type `BacktestCase`:
  ```ts
  export type BacktestCase = {
    ruleId: string;
    targetKey: string;
    outcome: string;
    label: "reversed" | "confirmed";
    firedAt: string;
    decidedAt: string;
    metadata?: Record<string, unknown>;
  };
  ```
  - `outcome` comes from the matching `RuleFiredEvent.outcome`.
  - `label` comes from the matching `HumanOverrideEvent.verdict`.
  - `firedAt` is the `RuleFiredEvent.occurredAt`; `decidedAt` is the `HumanOverrideEvent.occurredAt`.
  - `metadata` is the fired event's `metadata` field, omitted entirely from the object (not set to `undefined`) when the fired event has none — the same optional-property discipline `RuleFiredEvent` itself already uses.
- Export a pure function:
  ```ts
  export function buildBacktestCorpus(
    ruleId: string,
    fired: readonly RuleFiredEvent[],
    overrides: readonly HumanOverrideEvent[],
  ): BacktestCase[]
  ```
  - Matching rule: a `RuleFiredEvent` and `HumanOverrideEvent` pair into one `BacktestCase` when both have `ruleId` equal to the function's `ruleId` argument AND the same `targetKey`.
[...]

PR #9720, opened 2026-07-29
fix(stats): drop counterfactual-replay rows from the public per-rule precision

JSONbored:
## Summary

The public "Measured accuracy per rule" table published `ai_consensus_defect` and `slop_gate_score` with **byte-identical** `decided: 460, confirmed: 287, precision: 62.4%`. Not a coincidence — `slop_gate_score`'s override rows are the same label set, copied.

#8277's slop backfill re-scores the deterministic slop signals over archived diffs, but takes each label **verbatim** from the `ai_consensus_defect` corpus's human verdict on the same target: `manifestToSourceCases` (`scripts/backfill-slop-corpus.ts:51-58`) passes `backtestCase.label` straight through, and `replaySlopCorpus` writes it unchanged into the new rule's `verdict` (`scripts/backfill-slop-corpus-core.ts:133-134,165`). Only the *score* side is recomputed.

That is exactly what it was built to be — internal evidence for a flip-to-live decision, and by its own module header "a LOWER BOUND on what live scoring would have produced". But the public table describes itself as *"Precision of each automated rule over its human-decided cases in the last 90 days"*, and those were **another rule's** human-decided cases. It also made the table print one number twice, since both rules then shared a label set and a target set.

Corroborating that the published rows cannot be live measurements: `slop_gate_score` is in `GATE_SCORE_SIGNAL_CODES` (`src/rules/advisory.ts:207`), which `recordImplicitTerminalConfirmations` explicitly excludes (`src/review/outcomes-wire.ts:654`), and `recordConfiguredGateBlockerOverrides` only ever writes `reversed`. On the live path that rule has **no route to a `confirmed` verdict at all**, so a 62.4% could only ever have come from replayed rows.

Excluded by **provenance**, not by rule id, so the next cross-rule replay inherits the same protection. `review_targets_decision_level` (#8083's own backfill) is deliberately **not** excluded — its labels come from what actually happened to the PRs that rule fired on, so they do support a precision claim for it, and there is a test pinning that distinction.

Refs #9676

## Scope

## Validation

- [x] `git diff --check`
- [x] `npm run actionlint`
- [x] `npm run typecheck`
- [ ] `npm run test:coverage` locally; **`codecov/patch` requires ≥99% coverage of the lines AND branches you changed** (aim for **100%** on your diff so CI variance does not fail near the threshold). Global coverage is a non-blocking trend with a loose 90% backstop, not the gate.
- [ ] `npm run test:workers`
- [ ] `npm run build:mcp`
- [ ] `npm run test:mcp-pack`
- [ ] `npm run ui:lint`
- [ ] `npm run ui:typecheck`
- [ ] `npm run ui:build`

- Coverage measured **scoped**: `vitest run --coverage` over `public-rule-precision.test.ts` restricted to the changed module reports **100% statements / branches / functions / lines** (19/19, 10/10, 6/6, 17/17). Both sides of the new exclusion are pinned — a cross-rule replay row is dropped, a same-rule synthesized row is kept.
- Ran the neighbouring suites that consume this block (`public-stats`, `public-stats-route`, `public-eval-scores-route`): 59 tests green.
- No UI files touched, so `ui:lint`/`ui:typecheck`/`ui:build` were not run. `ui:openapi:check` passes — response shape is unchanged, only which rows qualify.
- `test:workers`, `build:mcp`, `test:mcp-pack` untouched by this diff; left to CI.

## Safety

No auth/CORS/session change. No UI file is touched — the fairness page already renders whatever rows the block returns, so this removes a row rather than changing a component; that is why there is no UI Evidence table here. The existing `INVARIANT: the public payload never carries target keys, repos, confidences, or private terms` test still passes.

## Notes

On the current production data every `slop_gate_score` row carries this provenance, so that rule drops out of the public table entirely and `ai_consensus_defect` remains at 460 / 287 / 62.4%. That is the honest outcome: one rule with a real (if synthesized-from-its-own-history) measurement, instead of two rules appearing to independently corroborate each other while sharing one label set.

### DIFF
--- a/src/review/public-rule-precision.ts
+++ b/src/review/public-rule-precision.ts
@@ -45,6 +45,23 @@ export type PublicRulePrecision = {
 };
 
 /**
+ * Provenance tags whose override rows carry a verdict that was NOT a human decision about the rule the row is
+ * filed under, and so cannot support a per-rule "measured accuracy" claim.
+ *
+ * `slop_replay_backfill_v1` (#8277) is the counterfactual replay: it re-scores the deterministic slop signals
+ * over archived diffs, but takes each label verbatim from the `ai_consensus_defect` corpus's human verdict on
+ * the same target (`scripts/backfill-slop-corpus.ts`'s `manifestToSourceCases`). That is exactly the evidence
+ * it was built to be -- internal input to a flip-to-live decision, and by its own module header a LOWER BOUND
+ * on live scoring -- but published under "precision of each automated rule over its human-decided cases" it
+ * asserts something untrue: those were another rule's human-decided cases. It also made the public table print
+ * one number twice, since both rules then shared a label set and a target set.
+ *
+ * NOT excluded: `review_targets_decision_level` (#8083's own backfill), whose labels come from what actually
+ * happened to the PRs THAT rule fired on -- synthesized rows, but genuinely about that rule.
+ */
+const NON_ATTRIBUTABLE_OVERRIDE_PROVENANCES = ["slop_replay_backfill_v1"] as const;
+
+/**
  * Load the public per-rule precision block. Fail-safe per section (the same degradation contract as
  * loadCalibrationTrend): a read error yields an empty/absent section, never a thrown public endpoint.
  */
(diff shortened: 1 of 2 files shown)


JSONbored:
Merge-order note: this PR and #9675 both edit `src/review/public-rule-precision.ts`. They originally conflicted — both inserted a constant directly after `HUMAN_OVERRIDE_EVENT_TYPE_PREFIX` — which would have given whichever merged second a base conflict. This branch now declares `NON_ATTRIBUTABLE_OVERRIDE_PROVENANCES` immediately above `loadPublicRulePrecision` instead, leaving the top-of-file constants block byte-identical to `main`.

Verified: all four open branches (#9675, #9718, #9720, #9721) merge cleanly into `main` in both forward and reverse order, and with all four merged together `typecheck` is clean and the eight affected suites pass (122 tests). No merge sequence is required.

PR #9752, opened 2026-07-29
fix(orb): parse the AMS PR outcome from an anchored dedupKey, not a substring scan

### LINKED ISSUE #9703: orb(notifications): stop deriving the AMS PR outcome by substring-scanning a client dedupKey

## Context

`ams_pr_outcome` is the only notification kind whose rendered content depends on a field that is not on the event. `buildAmsPrOutcomeNotification` (`src/notifications/service.ts:89-107`) recovers the decision by substring-scanning the dedup key:

    // buildAmsPrOutcomeEvent embeds `:merged:` or `:closed:` in the dedupKey after the PR number.
    const merged = event.dedupKey.includes(":merged:");

`DetectedNotificationEvent` has no `decision` field, so that string is the entire carrier. Three things make it unsafe:

1. **The validator does not check the shape.** `normalizeAmsNotificationEventInput` (`src/notifications/ams-events.ts:134-156`) type-checks every other field but for `dedupKey` only tests non-empty string (`:142`). The HTTP schema is equally permissive: `dedupKey: z.string().min(1).max(500)` (`src/api/routes.ts:487`). Any `ams_pr_outcome` whose key lacks `:merged:` renders as "AMS recorded close: ... closed without merge" — including for a merge.
2. **The format has two hand-duplicated producers.** `buildAmsPrOutcomeEvent` (`ams-events.ts:107-128`) and `buildAmsPrOutcomePayload` (`packages/loopover-miner/lib/ams-notifications.ts:118-139`) each independently write `ams_pr_outcome:${repoFullName}#${pullNumber}:${decision}:${closedAt}`. Nothing pins them together.
3. **The server-side builders are dead.** `buildAmsPrOutcomeEvent`, `buildAmsAttemptStartedEvent`, `buildAmsAttemptFailedEvent` and `buildAmsGovernorPausedEvent` have zero non-test callers (grep across `src/`, `apps/`, `packages/`); the miner's copies are what actually run. So the format the consumer documents itself against is produced by a file this repo never exercises — if the miner reorders `decision` and `closedAt`, every merge notification silently flips to "closed without merge" and nothing fails.

The repo has a convention for exactly this drift class — see the source-pinning tests at `test/unit/advisory-spend-gate.test.ts:49-71` and the `REVIEWER_VOTE_EVENT_TYPE` note at `src/services/reviewer-routing.ts:102-104`.

## Requirements

- Add to `src/notifications/ams-events.ts`:
  - `export const AMS_PR_OUTCOME_DEDUP_KEY_PATTERN` — an anchored `RegExp` matching exactly `ams_pr_outcome:<repoFullName>#<pullNumber>:(merged|closed):<closedAt>`, with the decision as a capture group.
  - `export function parseAmsPrOutcomeDecision(dedupKey: string): "merged" | "closed" | null` — the single reader of that pattern.
- `buildAmsPrOutcomeEvent` keeps its output, but the file must contain no second literal spelling of the format: the builder and the pattern must derive from one shared prefix/segment constant.
- `normalizeAmsNotificationEventInput` returns `null` for an `ams_pr_outcome` event whose `dedupKey` does not satisfy `parseAmsPrOutcomeDecision`. Every other `AMS_NOTIFICATION_EVENT_TYPES` kind keeps its current validation unchanged — this rule is `ams_pr_outcome`-only.
- `buildAmsPrOutcomeNotification` uses `parseAmsPrOutcomeDecision` instead of `.includes(":merged:")`. A `null` parse renders the **closed** copy (never claim an unproven merge); the tests below must show that arm is unreachable from the ingest route once the validator rejects it.
- Add a producer-parity test pinning `packages/loopover-miner/lib/ams-notifications.ts`'s `buildAmsPrOutcomePayload` output against `AMS_PR_OUTCOME_DEDUP_KEY_PATTERN` for both `merged` and `closed`.

> ⚠️ Required pattern: the source-pinned drift guard at `test/unit/advisory-spend-gate.test.ts:49-71` ("adoption is real, not aspirational") is the convention for the parity test. It does NOT satisfy this issue to add a `decision` field to `DetectedNotificationEvent` / the deliveries schema (a schema migration, out of scope); to validate only in `src/api/routes.ts`'s zod schema while `normalizeAmsNotificationEventInput` stays permissive (the MCP/self-host path reaches the normalizer directly); or to keep `.includes(":merged:")` and merely add tests around it.

## Deliverables

- [ ] `parseAmsPrOutcomeDecision("ams_pr_outcome:owner/repo#7:merged:2026-05-28T12:00:00.000Z")` returns `"merged"`, the `:closed:` twin returns `"closed"`, and `"custom-key"` returns `null` — asserted in `test/unit/notifications-ams-events.test.ts`.
- [ ] `normalizeAmsNotificationEventInput({ ...validAmsPrOutcome, dedupKey: "custom-key" }, "miner")` returns `null`, while the same malformed key on an `ams_attempt_started` event is still accepted — two cases.
- [ ] `buildAmsPrOutcomeNotification` renders the merge copy for a `:merged:` key and the close copy for both a `:closed:` key and an unparseable key — extending `test/unit/notifications-service.test.ts:144-155`.
- [ ] A parity test asserts `buildAmsPrOutcomePayload` produces a `dedupKey` matching `AMS_PR_OUTCOME_DEDUP_KEY_PATTERN` for both decisions.
- [ ] `src/notifications/service.ts` no longer contains the string `":merged:"` (grep-verifiable).

All Deliverables above are required in a single PR. A PR that satisfies only some of them — for example swapping the consumer without tightening `normalizeAmsNotificationEventInput` — does not resolve this issue.

## Test Coverage Requirements

99%+ Codecov patch coverage, branch-counted, on `src/**`; `src/notifications/ams-events.ts` and `src/notifications/service.ts` are inside `coverage.include`, so both arms of every changed conditional need a test. `packages/loopover-miner/lib/**` is ALSO inside `coverage.include` and gated, so the parity test is both mandatory and measured.

## Expected Outcome

A merged/closed AMS notification's wording is decided by a validated, single-sourced key format instead of a substring search over arbitrary client input. A miner-side edit that reorders the dedup key's segments fails CI instead of silently reporting every merged PR as "closed without merge".

## Links & Resources

`src/notifications/service.ts:89-107`; `src/notifications/ams-events.ts:107-128`, `:134-156`; `packages/loopover-miner/lib/ams-notifications.ts:118-139`; `src/api/routes.ts:479-493`, `:3842`; `test/unit/advisory-spend-gate.test.ts:49-71`.

shin-core:
## What & why

Closes #9703.

`ams_pr_outcome` is the only notification kind whose rendered content depends on a field that is not
on the event. `buildAmsPrOutcomeNotification` recovered the merged/closed decision with
`event.dedupKey.includes(":merged:")` — the dedupKey was the entire carrier. Three things made that
unsafe:

1. **Nothing validated the key's shape** — `normalizeAmsNotificationEventInput` only checked non-empty
   string, so any `ams_pr_outcome` whose key lacked `:merged:` rendered as "closed without merge",
   **including for a real merge**.
2. **Two hand-duplicated producers** wrote the format literal with nothing pinning them together.
3. **The server-side builder has no non-test callers** — the miner's copy is what runs, so a reorder
   of `decision`/`closedAt` there would silently flip every merge notification and nothing would fail.

## Change

- `ams-events.ts`: `AMS_PR_OUTCOME_DEDUP_KEY_PATTERN` (anchored, decision as capture group) +
  `parseAmsPrOutcomeDecision` (the single reader). The builder and the pattern derive from one shared
  prefix constant — the format is spelled once, not twice.
- `normalizeAmsNotificationEventInput` returns `null` for an `ams_pr_outcome` whose dedupKey doesn't
  parse (the MCP/self-host path reaches the normalizer directly, so validating only the HTTP zod schema
  isn't enough); every other AMS kind's validation is unchanged.
- `service.ts` uses `parseAmsPrOutcomeDecision`; a `null` parse renders the **closed** copy, never an
  unproven merge — unreachable from ingest now the normalizer rejects such a key upstream.

No `decision` field is added to the event/schema (out of scope).

## Validation

- Pattern/parser tests (well-formed round-trips for both decisions; null for wrong-prefix,
  no-decision, unknown-decision, and the loose-`:merged:`-elsewhere trap).
- Normalizer rejects a malformed `ams_pr_outcome` key while other AMS kinds keep their loose check.
- `buildAmsPrOutcomeNotification` renders the closed copy on a null parse.
- A **producer-parity test** pins the miner's `buildAmsPrOutcomePayload` output against the server
  pattern for both decisions (the drift guard).
- Bug-catch verified: reverting the normalizer/service fails the new assertions.
- 100% patch coverage on both source files.

### DIFF
--- a/src/notifications/ams-events.ts
+++ b/src/notifications/ams-events.ts
@@ -103,6 +103,27 @@ export function buildAmsGovernorPausedEvent(input: {
   };
 }
 
+// The `ams_pr_outcome` dedupKey is the ONLY carrier of the merged/closed decision (DetectedNotificationEvent
+// has no `decision` field), so its exact format is load-bearing. Both the producer below and the reader
+// (parseAmsPrOutcomeDecision) derive from this one prefix so the format is spelled once, not twice (#9703).
+const AMS_PR_OUTCOME_DEDUP_KEY_PREFIX = "ams_pr_outcome";
+
+/** Build the decision-carrying dedupKey. The single source of the `ams_pr_outcome:...:decision:closedAt` shape. */
+function amsPrOutcomeDedupKey(repoFullName: string, pullNumber: number, decision: "merged" | "closed", closedAt: string): string {
+  return `${AMS_PR_OUTCOME_DEDUP_KEY_PREFIX}:${repoFullName}#${pullNumber}:${decision}:${closedAt}`;
+}
+
+/** Anchored matcher for a well-formed `ams_pr_outcome` dedupKey, with the decision as capture group 1. Derives
+ *  its prefix + decision segment from the same constants the producer uses, so the two cannot drift (#9703). */
+export const AMS_PR_OUTCOME_DEDUP_KEY_PATTERN = new RegExp(`^${AMS_PR_OUTCOME_DEDUP_KEY_PREFIX}:.+#\\d+:(merged|closed):.+$`);
+
+/** The single reader of AMS_PR_OUTCOME_DEDUP_KEY_PATTERN: returns the embedded decision, or null when the key is
+ *  malformed (missing/unknown decision segment). A null result must never be rendered as an unproven merge. */
+export function parseAmsPrOutcomeDecision(dedupKey: string): "merged" | "closed" | null {
+  const match = AMS_PR_OUTCOME_DEDUP_KEY_PATTERN.exec(dedupKey);
+  return match ? (match[1] as "merged" | "closed") : null;
+}
+
 /** Miner-local PR outcome change (merged or closed). */
 export function buildAmsPrOutcomeEvent(input: {
   recipientLogin: string;
@@ -120,7 +141,7 @@ export function buildAmsPrOutcomeEvent(input: {
     recipientLogin,
     repoFullName: input.repoFullName,
     pullNumber: input.pullNumber,
-    dedupKey: `ams_pr_outcome:${input.repoFullName}#${input.pullNumber}:${input.decision}:${closedAt}`,
+    dedupKey: amsPrOutcomeDedupKey(input.repoFullName, input.pullNumber, input.decision, closedAt),
     deeplink: githubPullDeeplink(input.repoFullName, input.pullNumber),
     actorLogin: recipientLogin,
     detectedAt,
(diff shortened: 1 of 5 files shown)


JSONbored (commented):
Rejecting for this reason: 

```
The anchored pattern is a real improvement, but because `.+` is greedy and repoFullName/closedAt are still attacker(-miner)-controlled free text reaching the same dedupKey string, a crafted key can still smuggle a fake trailing `#N:decision:` segment that regex backtracking will match over the true one — this narrows but does not fully close the original class of spoofing bug the PR is fixing.
```

Issue #8525, opened 2026-07-24
release: reconcile-stale-releases tags at main head instead of the release PR's merge commit


JSONbored:
## What happened

The `reconcile-stale-releases` self-heal (#7149) creates a missing tag for a merged-but-untagged release PR at the **current main head**, not at the **release PR's own merge commit**.

Observed live today (2026-07-24): the ui-kit v1.1.2 release PR (#8342) merged as 962fa085d during the afternoon's Actions runner starvation, so its release-please tagging run was cancelled before executing (ten consecutive `Package Release Please` runs were cancelled while queued). When the self-heal finally ran, it created `ui-kit-v1.1.2` + the GitHub release at 16:37 pointing at d09fbe22f — main head, two commits later — which silently swept ddb674ebc (`fix(ui-kit): declare react-hook-form as a dependency`, #8494) into the 1.1.2 tag and the published artifact.

## Consequences

1. **Version-content drift**: published 1.1.2 contains a commit that semantically belonged to 1.1.3. Harmless this time (a dependency declaration), but the same mechanism would happily sweep a breaking change into a patch tag.
2. **Zombie release PR**: the already-open v1.1.3 PR (#8506) became permanently unreleaseable — release-please logs `No commits for path: packages/loopover-ui-kit, skipping` (the swept commit is now behind the tag) and neither regenerates nor prunes the PR, so its stale shared-manifest copy sat CONFLICTING with no self-heal path. Closed manually.

## Fix

When reconciling a merged-but-untagged release PR, resolve the PR's `merge_commit_sha` and create the tag there, not at the branch head. If the merge commit can't be resolved, fail loudly rather than tagging head.

## Acceptance criteria

- The reconcile path tags the release PR's `merge_commit_sha`.
- A regression test pins that the tag ref sent to the API equals the resolved merge commit, and that an unresolvable merge commit aborts without creating a tag.

JSONbored:
Verified fully resolved — nothing left to implement.

The fix landed in #8527 (`fix(release): tag and pack the resolved release commit, never the dispatch head`), already merged to main, covering all four published packages (`publish-engine.yml`, `publish-mcp.yml`, `publish-miner.yml`, `publish-ui-kit.yml`) — each now resolves the commit that actually introduced the release version into its `package.json` via `git log -S`, verifies it's reachable from main, checks out and tags/packs *that* commit, and aborts loudly instead of falling back to HEAD if it can't be resolved.

[...]

PR #7086, opened 2026-07-17
chore(release): cut miner v3.1.1

JSONbored:
:robot: I have created a release *beep* *boop*
---

### Fixes

* **miner:** bound oauth-device-flow.js's GitHub fetches with a request timeout ([cd9aedf](https://github.com/JSONbored/loopover/commit/cd9aedf7cceb45146302158225193bb963ab4eca))
* **miner:** fail closed when a chat-action handler throws ([8ba48bd](https://github.com/JSONbored/loopover/commit/8ba48bd707dcf002a31991ac434be8cd6a7822d8))
* **release:** scope MCP publish validation to its own package ([86ee117](https://github.com/JSONbored/loopover/commit/86ee11744f7c7d2f76c9d64345391f61651256c8))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @loopover/engine bumped from ^3.0.0 to ^3.2.1

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #9753, opened 2026-07-29
ci(release): wire @loopover/contract into the release and publish automation

### LINKED ISSUE #9749: @loopover/contract is unpublished but is a runtime dependency of @loopover/mcp and @loopover/miner — the next release of either ships uninstallable
## The problem

`@loopover/contract` is **not on npm**:

```
$ curl -s -o /dev/null -w "%{http_code}" https://registry.npmjs.org/@loopover%2Fcontract
404
$ npm install "@loopover/contract@^0.1.0" --dry-run
npm error 404  '@loopover/contract@^0.1.0' is not in this registry.
```

But both published CLI packages now declare it as a **runtime dependency**, not a dev one:

- `packages/loopover-mcp/package.json` → `"@loopover/contract": "^0.1.0"` in `dependencies`
- `packages/loopover-miner/package.json` → same

And both have **real runtime imports** in code that ships, built with plain `tsc` (no bundler, so the specifiers survive into the emitted JS):

- `packages/loopover-mcp/lib/telemetry.ts` → `@loopover/contract`
- `packages/loopover-mcp/bin/loopover-mcp.ts` → `@loopover/contract`, `/api-schemas`, `/tools`
- `packages/loopover-miner/lib/tenant-credential-resolution.ts` → `@loopover/contract/orb-broker`
- `packages/loopover-miner/lib/github-token-resolution.ts` → `@loopover/contract/cli-config`

## Why it hasn't broken yet

The currently-published versions predate the dependency. `@loopover/mcp@3.14.0` declares only `@loopover/engine`, `@modelcontextprotocol/sdk`, `posthog-node`, `zod`; `@loopover/miner@3.14.1` likewise. Both still install cleanly today — verified.

**It fires on the next release of either package.** That publish will carry `"@loopover/contract": "^0.1.0"`, and `npm install -g @loopover/mcp@latest` will fail with E404 for every external user.

Worth noting this also breaks the documented contributor workflow: `.claude/skills/contributing-to-loopover/SKILL.md` instructs `npm install -g @loopover/mcp@latest` as the pre-submit gate oracle.

## Supporting evidence that publishing was intended

`packages/loopover-contract/package.json` is already shaped for it — no `private: true`, `"publishConfig": { "access": "public" }`, and a `files: ["dist", "CHANGELOG.md"]` allowlist. It is the only workspace package with that shape and **no** `publish-*.yml` workflow (`engine`, `mcp`, `miner`, `ui-kit` all have one). It looks like the publish step was simply never added when the package was split out.

## Options

**A. Publish it** (add `publish-contract.yml` mirroring `publish-engine.yml`). Matches the package's existing shape and the other four packages' pattern. Contract must publish *before* any mcp/miner release, so release ordering becomes a real constraint.

**B. Bundle it into the consumers** — switch mcp/miner from plain `tsc` to a bundler that inlines `@loopover/contract`, and demote it to a `devDependency`. Keeps one less public package, but changes both build pipelines and their `dist` shape.

**C. Vendor the shared types** into each consumer. Cheapest to ship, worst for drift — it re-creates exactly the duplication `#9521` consolidated into this package.

Recommend **A**: it is the smallest change, matches how every sibling package already ships, and preserves the single-source-of-truth the contract package exists to provide.

## Guard worth adding either way

Nothing currently catches this. `test:mcp-pack` / `test:miner-pack` pack the tarball but never install it from a clean registry-only context, so an unresolvable dependency passes. A check that installs the packed tarball with `--omit=dev` against the public registry would have caught it before the dependency ever landed.

JSONbored:
## Summary

`@loopover/contract` is a runtime `dependencies` entry of both `@loopover/mcp` and `@loopover/miner`, but it had **no publish path** — no publish workflow, and no entry in release-please's config or manifest. It was the only workspace package *shaped* for publishing (no `private: true`, `publishConfig.access: public`, a `files` allowlist) with no automation behind it.

Left alone the next mcp/miner release ships uninstallable: both are at **3.15.2 locally against 3.14.x on npm**, so the already-cut publish carries `"@loopover/contract": "^0.1.0"` against a package that did not exist.

`0.1.0` has now been **bootstrap-published by hand** — npm's OIDC trusted publishing cannot create a brand-new package, the same bootstrap `@loopover/engine` needed — and a Trusted Publisher is configured against `publish-contract.yml` (GitHub Actions · `JSONbored/loopover` · environment `release`). **Renaming that file breaks the connection and every publish with it**, which is why the workflow header says so.

Closes #9749

## What's here

**`publish-contract.yml`** mirrors `publish-engine.yml` / `publish-ui-kit.yml`: unprivileged `validate` resolves the version, verifies the commit is on `main`, resolves the release commit *from the version string* rather than tagging HEAD (#8525), builds, packs and smoke-tests; privileged `publish` tags and pushes the exact tested tarball via OIDC behind `environment: release`. No separate typecheck step — this package's `build` **is** `tsc -p tsconfig.json`.

**The secret scan needed one deliberate change.** Unlike its siblings, this package *ships the redaction logic*: `dist/telemetry.js` defines `SECRET_VALUE_PATTERN`, whose source text contains a literal PEM `PRIVATE KEY` header, plus a doc comment quoting a full one describing the bug that alternative fixes. The siblings' verbatim pattern is a **guaranteed false positive on every release** — verified against the real 0.1.0 tarball. So the scan excludes that one module, and a second, narrower token-shaped scan still runs across *everything including* the excluded module, so a genuine leak there still fails the build.

**Prerequisite ordering.** mcp/miner depend on contract exactly as they depend on engine, so the reconciliation job now treats both as prerequisites — contract and engine publish first, and mcp/miner are skipped entirely if *either* fails, rather than resolving a version that exists nowhere and failing ETARGET.

**Versioning.** Contract stays independently versioned (like `ui-kit`) rather than joining the `engine-and-dependents` linked-versions group. It has only just shipped `0.1.0`; renumbering it into the 3.x train before it has consumers costs more than it buys. Worth revisiting at its next major.

**LICENSE.** The package declares `AGPL-3.0-only` but shipped no LICENSE file; `engine`, `mcp` and `ui-kit` all carry one. (`miner` is also missing one — out of scope here, but worth a follow-up.)

## Validation

Everything below was run against the **real** 0.1.0 tarball built from this tree:

| Check | Result |
|---|---|
| `npm run actionlint` | clean |
| `npm run lint:composite-actions` | clean |
| `npm run release-manifest:sync:check` | `checked 5 package version(s), all in sync` (was 4) |
| `release-please-config.json` / `.release-please-manifest.json` | parse clean, contract tracked at `0.1.0` |
| Both workflow files | valid YAML |
| Tarball allowlist rule | passes — only `dist/` + `package.json` + `README.md` |
| Secret scan **with** the exclusion | passes (verbatim sibling pattern **fails**, as expected) |
| Narrow token scan across **all** files | passes |
| Install-from-tarball + import `index` / `public-api` / `api-schemas` | passes, `TOOL_CATEGORIES` = 10 |
| `git diff --check` | clean |

The reconciliation ordering was also behaviorally tested against five scenarios with stubbed dispatch helpers — nothing-to-do, contract+mcp happy path, contract-fails, engine-fails-contract-ok, and mcp-only-with-deps-already-live. The load-bearing case is engine failing while contract succeeded: mcp/miner are still skipped, because both are prerequisites.

## Scope

## Safety

No runtime/`src` code changes and no UI output, hence no coverage obligation and no UI Evidence table. The privileged `publish` job holds `id-token: write` but never runs `npm install` or a build — it only downloads and publishes the tarball the unprivileged job already produced and tested, preserving the same privilege separation as every sibling publish workflow. No credential value appears anywhere in the diff.


JSONbored:
Valid blocker — fixed in the latest push. You were right, and the repo's own documentation settles it against my original reasoning.

`mcp-release-please.yml`'s header states three things that together make this unambiguous:

1. The race is caused by a **real runtime dependency edge** combined with `separate-pull-requests: true` — node-workspace bumps the dependent's range onto the dependent's *own* branch, leaving its `package.json` requiring a version that exists neither locally (the prerequisite's bump is on a different unmerged branch) nor on npm (it hasn't published). `npm ci` then fails ETARGET.
2. node-workspace's own `merge` option **does not** override `separate-pull-requests` — already tried, and verified against release-please's `manifest.ts` and `docs/manifest-releaser.md`. So linked-versions is the *only* mechanism.
3. `ui-kit` is outside the group **because it has no dependency edge to any other package here** — not because independent versioning is a default.

My "independent like ui-kit" justification inverted that last point: ui-kit is independent *precisely because* it lacks the edge contract now has. And this was not hypothetical — the header cites it as confirmed live across several mcp/miner cycles (#7086/#7087/#7107/#7108/#7119/#7120/#7121).

So rather than confirm cross-group reconciliation works, I've added `contract` to the group. I also rewrote the header comment to say that **membership tracks the dependency edge, not release cadence**, and called out "publish it like ui-kit" as the wrong instinct for any future package something here depends on — since that framing is exactly what produced this gap.

The reconciliation ordering added in this PR stays as-is. The two mechanisms cover different stages: the group fixes the **release-PR** stage (prerequisite version and dependents' ranges land in one commit), reconciliation fixes the **publish** stage (dispatch ordering and registry propagation). Engine needs both today; contract needs both for the same reasons.

**Cost, stated plainly:** contract joins the group's version train rather than staying at `0.1.0`, so its next release won't be `0.2.0`. That's a worse version story and a better correctness story, and `0.1.0` is published either way. If you'd rather keep contract on true independent semver, the only clean way is removing the edge — bundling contract into mcp/miner and demoting it to a devDependency (option B on #9749) — which is a real change to both build pipelines, not a config tweak.

`actionlint`, `lint:composite-actions`, and `release-manifest:sync:check` (5 packages, all in sync) still pass.

JSONbored:
Second push: the config fix alone closed today's instance but not the class, so I've added enforcement.

**The actual problem** is that linked-versions membership is a function of the dependency graph, and nothing checked that. #9521 made `@loopover/contract` a runtime dependency of both `@loopover/mcp` and `@loopover/miner` without touching the group, and it sat unnoticed until #9749 — by which point the package wasn't even published, so the next mcp/miner release would have shipped uninstallable. A config edit fixes that once; the next edge added silently reintroduces it.

`scripts/check-release-linked-versions.ts` derives the requirement from the real `package.json` edges between release-please-tracked packages and fails when two are joined by a shipping dependency but not by a linked-versions group. Wired into `test:ci` next to `release-manifest:sync:check`.

It also closes a **second silent gap** I hit while wiring this: `sync-release-manifest.ts` builds its work list from the *manifest's own keys* (`if (!(workspacePath in manifest)) continue`), so a package added to `release-please-config.json` but not to the manifest is silently exempt from the staleness check that stops release-please re-proposing an already-published version. Adding a package to one file and not the other had zero feedback.

**Scope decision, written down at the constant:** only `dependencies`/`peerDependencies`/`optionalDependencies`. `devDependencies` are excluded because they never reach a consumer, and forcing permanent version lockstep over a build-time-only edge costs more than the failure it prevents — which surfaces immediately in CI on the introducing branch rather than in a published artifact. There's a test pinning that exclusion so it can't drift by accident.

`ui-kit` is the case proving this isn't just "group everything": no dependency edge, stays ungrouped, and an invariant test asserts it is never reported. Otherwise the check would drag unrelated packages into lockstep, which is the opposite of the goal.

**Verified both directions:**

| Config state | Exit | Output |
| --- | --- | --- |
| contract removed from group (reproduces #9521) | **1** | names both `mcp→contract` and `miner→contract` edges |
| contract's manifest entry removed | **1** | names the package and the consequence |
| real config | **0** | `5 tracked package(s), all dependency edges linked and manifest-tracked` |

15 unit tests, pure-core/thin-IO split mirroring `sync-release-manifest.ts`. The only uncovered lines are that file's default IO closures (`readFileSync` wrapper, `process.exit`) — the same untestable seam its sibling has. `typecheck`, `actionlint`, `validate:no-hand-written-js`, `coverage-boltons:check`, and both release checks all pass.

PR #9763, opened 2026-07-29
fix(release): publish @loopover/contract, and catch this class of break before it ships (#9749)

### LINKED ISSUE #9749: @loopover/contract is unpublished but is a runtime dependency of @loopover/mcp and @loopover/miner — the next release of either ships uninstallable

JSONbored:
## The break

`@loopover/contract` is a **runtime** dependency of both published CLIs — `mcp` and `miner` import it from code that ships, built with plain `tsc` so the specifiers survive into the emitted JS — but it had no publish workflow and was never on npm.

Nothing has broken *yet*: the currently-published versions predate the dependency and install fine. The **next** release of either would publish `"@loopover/contract": "^0.1.0"` and fail with E404 for every external user — and a published version can't be taken back. It also breaks the documented contributor workflow, since `SKILL.md` instructs `npm install -g @loopover/mcp@latest` as the pre-submit oracle.

## Option A, and why not B or C

The package is already shaped for publishing — `publishConfig.access: public`, a `files` allowlist, no `private: true` — and is the **only** workspace package with that shape and no publish workflow. The step was simply never added when #9521 split it out.

Bundling it into both consumers (B) changes two build pipelines and their `dist` shape. Vendoring the types (C) re-creates exactly the duplication #9521 consolidated into this package.

`publish-contract.yml` mirrors `publish-engine.yml`'s privilege separation: the unprivileged `validate` job builds, packs and smoke-tests; the privileged `publish` job only tags and publishes the tarball that job already tested, so a compromised build dependency can't reach the OIDC token. The smoke test imports both the root export and a subpath (`/public-api`), since consumers use `@loopover/contract/api-schemas` and friends.

## The more useful half

`scripts/check-publishable-deps.ts` — **a published package must not depend on an unpublishable workspace one.**

I proved it catches the real bug rather than assuming: removing the new workflow makes it fail, naming both consumers and the reason.

```
@loopover/mcp -> "@loopover/contract": "^0.1.0"
  @loopover/contract has no .github/workflows/publish-*.yml, so it is not published to npm
@loopover/miner -> "@loopover/contract": "^0.1.0"
  ...
```

It's deliberately about the **class**: any future workspace package that becomes a runtime dependency of a published one is caught the same way, with no list to remember to update. "Published" is derived from the `publish-*.yml` workflows themselves rather than a hand-kept list, so it stays honest when a fifth package appears — and a non-publish workflow mentioning a package can't fake it published. `devDependencies` are ignored: they never ship, so flagging them would be noise that pushes contributors toward the wrong fix. A **private** dependency gets a distinct message, because adding a publish workflow fixes the unpublished case and cannot fix that one.

Wired into `test:ci` beside the other drift checks. 9 unit tests on the pure core.

## Two things this deliberately does not do

Both stated where they matter rather than left to be discovered:

- **It cannot enforce ordering between two publishes.** A contract release must go out *before* an mcp release depending on a new version of it. No static check sees that; it stays a runbook step, called out in the workflow header.
- **npm's OIDC cannot create a brand-new package.** The first contract publish needs a one-time authenticated bootstrap plus a Trusted Publisher entry in npmjs.com's settings — in the workflow header, so it isn't discovered when the first dispatch fails.

Closes #9749


JSONbored:
Rebased onto `main` and addressed the review.

**Review finding — `publishedPackageNames` over-matching.** Fixed. It now reads the package a workflow actually **packs**, from its own `npm pack --workspace <name>` line, instead of any `@loopover/*` mention anywhere in the file. The looser form meant a publish workflow's own build step (`npx turbo run build --filter=@loopover/contract`, `npm run build --workspace @loopover/engine`) would mark a sibling "published" and mask exactly the violation this check exists to catch. A workflow with no recognizable pack line now contributes **nothing** rather than falling back to a looser match — an unreadable workflow should make the check stricter, never quietly more permissive.

Three tests cover it: the regression case (a workflow that builds/tests three siblings and packs one yields only the packed one), the no-pack-line case, and an **invariant** asserting every real `publish-*.yml` in the repo yields exactly one package — so a future workflow that packs two, or none, fails here rather than silently changing what counts as published.

**Descoped against `main`.** #9753 landed the publish workflow and the release-please wiring independently, and its version is equivalent — same two-job privilege split, same pack line, and it does build before packing. I took `main`'s wholesale for `publish-contract.yml`, `release-please-config.json`, and `.release-please-manifest.json`. What remains here is what `main` doesn't have: the two checkers.

**One tightening to the workflow `main` landed.** Its secret scan excluded `telemetry.js`/`.map`/`.d.ts`, leaving the one module that handles secrets as the one module nobody scanned. The exclusion existed because a doc comment quoted a full PEM header; this branch already rephrases that comment to describe it instead, which makes the exclusion unnecessary. Verified against the packed tarball: with all three files scanned, the pattern finds nothing — the surviving `BEGIN [A-Z ]*PRIVATE KEY` is the regex *source*, which doesn't match `BEGIN (RSA |EC |OPENSSH )?PRIVATE KEY`. The release scan now covers all 97 packed files, and the redundant "prove the exclusion stayed narrow" follow-up grep is gone with it.

That removal rests on the doc comment staying descriptive, so it's pinned by a regression test plus an invariant on `FORBIDDEN_CONTENT` — a future widening of the scanner fails in `test:ci` with an explanation rather than failing an irreversible release.

Local: `tsc`, `actionlint`, `publishable-deps:check`, `test:contract-pack`, `release-linked-versions:check`, `validate:no-hand-written-js` all green; 33 unit tests pass.

PR #9764, opened 2026-07-29
feat(api): publish the eval corpus so a stranger can actually verify a review

### LINKED ISSUE #9636: Public verifiability: the published freeze point commits to an empty corpus, and the walkthrough's stats link 404s
## What's wrong

Three defects on the public verifiability surface, verified live against production on 2026-07-28.

### 1. The reproducibility freeze point commits to an empty corpus

`GET https://api.loopover.ai/v1/public/stats` reports:

```json
"latestBacktestRun": { "corpusChecksum": "4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945", "at": "2026-07-28T21:18:14.855Z" }
```

That value is `checksumCases([])` — SHA-256 over the canonicalized empty case list (`printf '[]' | shasum -a 256` reproduces it exactly). The `/fairness` page renders it as a **"Reproducibility freeze point"**, and `GET /v1/public/eval-scores` publishes two records committed to it carrying `trust.tier: "reproducible"` alongside `decided: 460, confirmed: 287`.

A hash over zero cases is byte-identical for every rule, every window and every deployment, so it commits to nothing a reader could re-derive the scores from. The scores and the commitment also come from different datasets: `decided`/`confirmed` are computed live from `signal.human_override:*` audit events over a trailing 90 days, while the checksum is one frozen value from the most recent backtest run.

**Why the corpus is empty:** `.github/workflows/calibration-advisory.yml` runs `backtest-corpus-export.ts` hardcoded to `--rule-id linked_issue_scope_mismatch`. That rule's `signal.rule_fired` events are only written on a path gated on `linkedIssueSatisfactionGateMode === "block"`, which defaults to `"off"`, so `buildBacktestCorpus` returns `[]` and nothing refuses to publish the result.

### 2. The walkthrough's stats reference 404s

`apps/loopover-ui/content/docs/verify-this-review.mdx` step 4 cites a bare `` `/v1/public/stats` ``. A reader on `loopover.ai/docs/verify-this-review` resolves that to `https://loopover.ai/v1/public/stats`, which the UI worker does not serve — only `api.loopover.ai` does. Sibling docs (`what-you-can-verify.mdx`) already use the absolute host.

### 3. The walkthrough overstates who can run it

The page says "Nothing posts anywhere, nothing needs an API key" and `what-you-can-verify.mdx` says "**Anyone** can do this for public repositories". But step 1 (`backtest-corpus-export.ts --remote`) and the track-record command shell out to `wrangler d1 execute --remote`, reading the deployment's own Cloudflare D1 with that deployment's credentials. A third party cannot run step 1 at all. There is no HTTP corpus-export route — any unmatched `/v1/*` path returns `401 {"error":"unauthorized"}` from the catch-all auth middleware, which is what makes a guessed corpus URL look auth-gated rather than absent.

## Scope of this issue

Stop publishing a commitment that verifies nothing, and make the walkthrough's host references and audience claims accurate.

Explicitly **out of scope** (each needs its own decision and issue): publishing an actual anonymous corpus snapshot (`BacktestCase` carries `targetKey` and firing `metadata`, so it needs a redaction design before anything is published), making `corpusChecksum` per-rule and window-bound, and backfilling the missing `rule_fired` signal.

JSONbored:
## Summary

The verifiability walkthrough tells a skeptic to *"export the same corpus the numbers come from"* — but the only exporter is `scripts/backtest-corpus-export.ts`, which shells out to `wrangler d1 execute --remote` and needs **that deployment's own Cloudflare credentials**. Step 1 was never runnable by a third party. #9636 fixed the *claim* by marking the step operator-only; this fixes the *gap*.

`GET /v1/public/eval-corpus?ruleId=<id>` serves one rule's labeled corpus over the same window the published per-rule precision covers, unauthenticated.

Closes #9636

## Redaction is the design — and it costs nothing replay needs

| Field | Treatment | Why |
|---|---|---|
| `targetKey` (`owner/repo#number`) | **dropped, not hashed** | `scoreBacktest` never reads it, and a hash is still a stable per-PR identifier — correlatable, for zero replay benefit |
| `metadata` | narrowed to `confidence`, **kept nested** | `buildConfidenceThresholdClassifier` reads `metadata.confidence` (`backtest-threshold.ts:21`). Flattening it would make every replay hit the `?? 1` fallback and label every case `"confirmed"` — a wrong answer that looks like a working one |
| `firedAt` / `decidedAt` | truncated to the UTC day | nothing in the replay path reads them; full-precision pairs were the one remaining way to line a case up against a private repo's PR timeline |

**The checksum commits to *this* artifact**, not to the internal manifest. A reader can only re-derive what they can download, and a checksum over a preimage nobody can obtain verifies nothing — exactly the failure #9636 fixed on the eval-score records. It uses `canonicalJson` + Web Crypto, the same pair every other digest here uses, so a consumer learns one serialization rule rather than two.

Counterfactual-replay overrides are excluded, matching the per-rule precision this corpus explains — otherwise the corpus and the number it backs would disagree.

## Docs

Step 1 is now a credential-free `curl`. **Step 2's verification snippet imports nothing from LoopOver** — canonical JSON is eight lines, and running our code to check our numbers is not much of a check. I executed that snippet against real output before writing it down; it prints `checksum OK`.

A callout records that the operator CLI's older canonicalization produces a different, deliberately non-comparable checksum, since each commits to the artifact it ships with. Without that note, an operator verifying a downloaded document with `buildBacktestCorpusManifest` would get a false `CHECKSUM MISMATCH`.

`what-you-can-verify.mdx`'s *"private repositories are the exception"* callout is **corrected rather than left flattering**: a private repo's cases **do** appear here, because the published precision counts them and a corpus omitting them would not explain those numbers. They appear stripped of identity. What is still not published is any identifiable per-tenant history.

## Validation

| Check | Result |
|---|---|
| `public-eval-corpus.ts` coverage | **100%** stmts / branches / funcs / lines (22/22, 18/18, 8/8, 21/21) |
| Changed lines in `routes.ts` + `route-auth.ts` | 22 changed lines, **0 uncovered, 0 partial branches** |
| Unit + integration + auth-parity suites | **59 tests pass** |
| `typecheck` · `ui:openapi:check` · `docs:drift-check` · `ui:lint` (0 errors) · `ui:build` · `coverage-boltons` · `validate:no-hand-written-js` · `release-linked-versions:check` | all clean |
| `git diff --check` | clean |

Registered in the table-driven `exemptPublicRoutes` regression test whose stated purpose is *"a future public route that forgets its `requiresApiToken` exemption fails HERE, not in prod"*, and `openapi-security-parity` passes.

Two invariant tests carry the weight: one asserts the serialized response never matches `/acme|private-widgets|4242|targetKey/i` against a deliberately identifying seed, the other pins `metadata.confidence` staying nested.

## Scope

## Safety

This adds a new **unauthenticated** route, so the negative paths are the point: 404 when `LOOPOVER_PUBLIC_STATS` is off (same flag as its siblings), 400 without `ruleId` rather than silently picking a rule, an empty-but-still-checksummed corpus for an unknown rule, and a fail-safe empty corpus when the store read throws rather than a 500 on a public endpoint. The exemption is declared in `route-auth.ts` — the single predicate both the auth middleware and the OpenAPI builder read — not in a second, driftable place.

## UI Evidence

Captured at 1280×800, `deviceScaleFactor: 2`, dark build.

| State | Before | After |
| --- | --- | --- |
| Step 1 — corpus acquisition | <a href="https://raw.githubusercontent.com/JSONbored/loopover/screenshots/before-corpus-step1.jpg">Before: export, operator/self-host only</a> | <a href="https://raw.githubusercontent.com/JSONbored/loopover/screenshots/after-corpus-step1.jpg">After: credential-free curl, anyone</a> |
| Step 2 — checksum verification | <a href="https://raw.githubusercontent.com/JSONbored/loopover/screenshots/before-corpus-step2.jpg">Before: imports buildBacktestCorpusManifest</a> | <a href="https://raw.githubusercontent.com/JSONbored/loopover/screenshots/after-corpus-step2.jpg">After: self-contained, imports nothing from LoopOver</a> |

## Notes

On current production data this endpoint returns a real corpus for `ai_consensus_defect` and an empty one for `slop_gate_score` — correct on both counts, and consistent with what the fairness table now shows after #9720.


JSONbored:
Not previously fixed — you were right to push, and it turned out to be worse than the reviewer framed it. Two defects, both now fixed in the latest push.

## 1. The fallback did mask an outage — and it was the exact bug this endpoint exists to close

`listAuditEventsByType` awaits its drizzle query directly and does **not** swallow, so my `try/catch` was live: a database failure returned `{ caseCount: 0, checksum: sha256("[]"), cases: [] }` with a `200`.

That is a checksummed empty corpus. It verifies perfectly and commits to nothing — **precisely the defect #9636 was opened for, reintroduced by the endpoint built to close it.** A reader downloading during an outage would verify the checksum, replay zero cases, and conclude the rule has no history.

The siblings' degrade-to-empty posture is right for *them*: each is one section of a larger payload. Here the whole response **is** the artifact and it carries a commitment, so the read now propagates and the route returns **503 `eval_corpus_unavailable`**. "The database was down" can no longer be mistaken for "this rule has no history."

## 2. Worse, and not in the review: `truncated` could report `false` over a truncated corpus

`queryRuleHistory` inherits `listAuditEventsByType`'s **default `limit = 500`**. Truncation therefore happened *upstream, before pairing*, invisible in `cases.length` — so `applyPublicEvalCorpusCap` saw a short array, and the response asserted `truncated: false` over a corpus SQL had already cut short. Production's `ai_consensus_defect` sits at 460 decided, so this is live at 501.

A checksummed artifact that silently omits rows while claiming completeness is a worse lie than the empty one.

Fixed with `queryBoundedRuleHistory`, which takes an explicit limit and returns a `saturated` flag. `truncated` is now `capped || saturated`. `PUBLIC_EVAL_CORPUS_MAX_CASES` also drops from an invented 5,000 to `MAX_RULE_HISTORY_EVENTS` (2,000) — `listAuditEventsByType` clamps at 2,000, so a cap above the retrieval ceiling was unreachable by construction and could never have fired.

`saturated` can false-positive when a rule has exactly `limit` events. Over-reporting truncation is the safe direction; the alternative asserts a completeness nobody checked.

The interface method is untouched — its read errors "deliberately propagate" for the decision-time callers, and the 500 default is fine for them; only a *publishing* consumer needs the bound made explicit.

## Validation

| File | Changed lines | Uncovered | Partial branches |
|---|---|---|---|
| `public-eval-corpus.ts` | 157 | none | none |
| `signal-tracking-wire.ts` | 37 | none | none |
| `routes.ts` | 26 | none | none |
| `route-auth.ts` | 4 | none | none |

54 tests green, including two new regressions: the loader now `rejects.toThrow()` on a broken store, and the route returns 503; plus a real-D1 test seeding past the event ceiling to prove `truncated` flips. Rebased onto current `main`; `typecheck` and `ui:openapi:check` clean. The 503 is registered in OpenAPI with the reason.

PR #9718, opened 2026-07-29
fix(stats): withhold reversal-grounded accuracy where a reversal cannot be recorded

JSONbored:
## Summary

`/fairness` showed **100%** accuracy for every repository and all eight weeks of the trend, while the headline read **94.6%**. The 100%s are structural zeros, not measurements.

`recordReversalSignals` has exactly one call site, inside `handlePullRequestWebhookEvent` (`src/queue/processors.ts:7039`), reached only via the `github-webhook` job — a member of `REVIEW_EXECUTION_JOB_TYPES`, which a runtime that does not execute reviews acks-and-drops (`src/index.ts:61-75`). So `reversed` is pinned at 0 while merged/closed keep growing, and `1 - 0/N` renders as a flawless score over a numerator that can never move. The repo already recorded the empirical outcome — `src/review/reversal-superseded.ts:3-4`: "verified in production: zero reversal events ever."

### 1. Accuracy is withheld when a reversal is not observable

`accuracyPct` now returns `null` unless a reversal is observable on this deployment. A recorded reversal proves it outright — and short-circuits the probe, so no deployment that has ever overturned an auto-action pays a query. Otherwise there must be at least one terminal auto-action (`agent.action.close`/`agent.action.merge`) in the retained window for a reversal to attach to.

This is the discipline the surface already applies, extended to the case where the numerator's **writer**, not just its sample, is absent: a sparse rule's precision is null (`PUBLIC_PRECISION_MIN_DECIDED`), and `gamingFlagsCaught` is null below three instances (`src/orb/analytics.ts:343`). Volume columns are measured directly and still publish.

### 2. The weekly trend's denominator asymmetry

`merged`/`closed` folded in registered-Orb rows (`public-accuracy-trend.ts:181-182`) while `reversed` stayed own-ledger-only — the Orb aggregate has no reversal concept — so the denominator grew with every newly registered install while the numerator did not. #7449 fixed exactly this for the lifetime figure (`public-stats.ts:463-470`) and it was never carried across to the trend. Accuracy now divides the own-ledger pairing; displayed volume still includes the fleet.

### 3. UI

- Withheld cells now say *why* instead of leaving a bare em dash beside healthy volume.
[...]

Issue #7449, opened 2026-07-20
fix(review): public-stats accuracyPct denominator includes Orb-fleet merged/closed but numerator (reversed) stays own-ledger-only


JSONbored:
## Context
`src/review/public-stats.ts`'s `getPublicStats` computes the homepage's global `accuracyPct`
(line 401) as `accuracyPct(totals.merged, totals.closed, totals.reversed)`. `totals.reversed` is
accumulated only from loopover's own allowlisted repos (line 358), but `totals.merged`/
`totals.closed` are folded with the registered Orb fleet's aggregate counts first (lines 381-382,
via `getOrbGlobalStats`, which — per this file's own comment at line 378 — "has no reversal concept
at all"). `accuracyPct`'s formula (line 151) is `1 - min(1, reversed / (merged + closed))`. Because
the denominator (`merged + closed`) grows with the whole registered fleet while the numerator
(`reversed`) stays fixed to a handful of own-ledger repos, the published accuracy percentage trends
toward 100% as more self-hosted installations register, independent of the fleet's actual
merge/close reversal behavior — which this number is supposed to represent.

This file's header comment is otherwise very thorough about reasoning through this fold's
tradeoffs (explicitly noting reversals stay own-ledger-only), but doesn't address this specific
downstream skew on the published `accuracyPct` value.

## Requirements
- The publicly-displayed global `accuracyPct` must not be computable in a way where its numerator
  (reversed count) and denominator (merged+closed count) are drawn from structurally different,
  independently-growing populations. Acceptable fixes include (pick one, and document the choice
  in a code comment next to `accuracyPct`'s call site):
  1. Compute the global `accuracyPct` from own-ledger totals only (i.e. call `accuracyPct` with the
     pre-fold `ownLedger`-scoped merged/closed, not the fleet-folded `totals.merged`/`totals.closed`),
     leaving the Orb-fleet merged/closed counts to inflate only the raw `reviewed`/`handled`/
     `minutesSaved` totals as they do today (which have no numerator/denominator mismatch problem);
  2. Or, if `getOrbGlobalStats` can be extended to also report a fleet-wide reversal count, fold
     that into `totals.reversed` alongside `merged`/`closed` so numerator and denominator scale
     together.
  Do not silently drop the Orb-fleet fold from `merged`/`closed`/`reviewed`/`handled` entirely —
  those aggregate totals are working as intended per the file's own comments; only the
  `accuracyPct` computation's numerator/denominator mismatch needs fixing.
[...]

PR #9768, opened 2026-07-29
fix(stats): bound the accuracy denominator to the window its numerator survives in

JSONbored:
## Summary

#9718 gated accuracy on whether a reversal is observable at all, and fixed the weekly trend's Orb-folded denominator. It did **not** fix the third asymmetry in #9676:

`merged`/`closed` come from `github_app.pr_public_surface_published` — the single **retention-exempt** audit event type (`DURABLE_AUDIT_EVENT_TYPES`, `src/db/retention.ts:14`) — so they are lifetime and never shrink. `reversal_*` rows prune with the rest of `audit_events` at 90 days. Pairing an immortal denominator with a 90-day numerator makes `1 - reversed/decided` drift toward 100% as the ledger ages, independent of real reversal behavior.

**Confirmed live after #9718 deployed:**

```
byProject accuracy: [100, 100, 100]   # on 2377 / 602 / 508 reviewed, 0 reversals
```

The observability gate #9718 added *passed* here — the reversal signal genuinely is observable on that deployment — and the ratio was still meaningless. Two different bugs; the first fix couldn't have caught this one.

## The fix

A second, retention-windowed disposition query supplies the accuracy denominator. It shares **both** of the numerator's bounds: own-ledger only, and inside `audit_events`' retention window. `reviewed`/`merged`/`closed` keep publishing lifetime volume, which is measured and correct.

That display-vs-denominator split is the same shape #7449 established for the Orb fold and #9718 carried into the weekly trend. The #7449 snapshot this supersedes (`ownLedgerMerged`/`ownLedgerClosed`) is **removed rather than left dead**, with its reasoning folded into the new comment.

**Boundary case, stated rather than hidden:** a PR published before the window but reversed inside it contributes to the numerator and not the denominator. `accuracyPct`'s existing clamp already handles that — it's the same shape as the reopened-auto-close case its comment already describes — and the alternative, an unbounded denominator, is the bug being fixed.

Refs #9676

## Validation

The stub-based tests route both disposition queries to the same rows, so they pass either way and prove nothing about the windowing. The regression test is therefore **real-D1**: four merged PRs published 200 days ago plus two recent ones (one auto-closed and reversed, one merged and standing).

| Assertion | Value |
|---|---|
| `totals.reviewed` | 6 — lifetime volume unaffected |
| `totals.merged` | 5 — lifetime volume unaffected |
| `totals.reversed` | 1 |
| `byProject[0].accuracyPct` | **0** — windowed denominator is 1, not the lifetime 5 |
| `totals.accuracyPct` | **0** |

Under the old pairing that same fixture published `1 - 1/5 = 80%`.

- **39 changed lines in `public-stats.ts`: 0 uncovered, 0 partial branches.**
- 52 tests green across `public-stats` + `public-stats-route`, including the existing `REGRESSION (#7449)` test, which still passes — the property it pins is preserved, not replaced.
- `typecheck`, `ui:openapi:check`, `git diff --check` all clean.

## Scope

## Safety

No auth/CORS/session change, and no UI file is touched — the fairness page already renders whatever the block returns, and #9718 shipped the copy explaining a withheld or lowered accuracy. `ui:openapi:check` confirms the response shape is unchanged: `accuracyPct` was already nullable, and this narrows what it is computed from, not what it can be.

### DIFF
--- a/apps/loopover-ui/src/routeTree.gen.ts
+++ b/apps/loopover-ui/src/routeTree.gen.ts
@@ -9,83 +9,43 @@
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
 
 import { Route as rootRouteImport } from './routes/__root'
-import { Route as RoadmapRouteImport } from './routes/roadmap'
-import { Route as MinersRouteImport } from './routes/miners'
-import { Route as MaintainersRouteImport } from './routes/maintainers'
-import { Route as InstallRouteImport } from './routes/install'
-import { Route as FairnessRouteImport } from './routes/fairness'
-import { Route as DocsRouteImport } from './routes/docs'
-import { Route as ChangelogRouteImport } from './routes/changelog'
-import { Route as AppRouteImport } from './routes/app'
-import { Route as ApiRouteImport } from './routes/api'
-import { Route as AgentsRouteImport } from './routes/agents'
 import { Route as IndexRouteImport } from './routes/index'
-import { Route as InstallIndexRouteImport } from './routes/install.index'
-import { Route as DocsIndexRouteImport } from './routes/docs.index'
-import { Route as AppIndexRouteImport } from './routes/app.index'
+import { Route as AgentsRouteImport } from './routes/agents'
+import { Route as ApiRouteImport } from './routes/api'
+import { Route as AppRouteImport } from './routes/app'
+import { Route as ChangelogRouteImport } from './routes/changelog'
+import { Route as DocsRouteImport } from './routes/docs'
+import { Route as FairnessRouteImport } from './routes/fairness'
+import { Route as InstallRouteImport } from './routes/install'
+import { Route as MaintainersRouteImport } from './routes/maintainers'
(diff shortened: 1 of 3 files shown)


JSONbored:
Good catch — fixed in the latest push.

You're right that the comment above the global `accuracyPct:` call site still pointed at `ownLedgerMerged`/`ownLedgerClosed`, which this PR removes. A comment referring a reader to a symbol that no longer exists is worse than no comment: it sends them hunting for context that was deliberately relocated.

Retargeted at the windowed disposition query, and **widened while there**. The old text described only *one* of the two bounds the pairing needs — same population, so the fleet fold can't inflate it. It now names both, since this PR adds the second: same retention window, so an immortal denominator can't outlive its numerator. That second bound is the actual subject of the PR, so the comment was under-describing the code even after the rename was corrected.

For the record, `ownLedgerReviewed` and `ownLedgerMinutes` are still present and untouched — different variables, both still live, feeding the minutes-saved calculation. Only the two accuracy-denominator snapshots were removed.

`typecheck` clean and the 48 `public-stats` tests still pass.

PR #9796, opened 2026-07-29
feat(retention): bound orb_signals by folding it into per-day rollups

### LINKED ISSUE #9783: orb_signals grows unbounded, and adding retention would silently corrupt two existing surfaces
## The gap

`orb_signals` is **absent from `RETENTION_POLICY`** (`src/db/retention.ts:16-70`), while both its siblings have rules: `orb_pr_outcomes` at 90 days and `orb_webhook_events` at 14. Every self-host instance exports into it hourly (`src/selfhost/orb-collector.ts`), so it grows without bound.

It is also a **public data source**, and has been since before the weekly fleet trend: `computeFleetAnalytics` reads it for the `fleetAccuracy` headline on `/fairness` (`src/orb/analytics.ts:266-283`), and #9775 added a second reader for the weekly series.

## Why this is not a one-line addition

Adding `{ table: "orb_signals", column: "received_at", days: 90 }` looks obvious and would **silently change two figures that are already published**:

**1. `listFleetInstances` counts lifetime signals.** `src/orb/fleet-admin.ts:19-27`:

```sql
(SELECT COUNT(*) FROM orb_signals s WHERE s.instance_id = i.instance_id) AS signalCount
```

That is an unbounded cumulative count, presented in the owner-facing fleet view as *"every instance that has ingested signals"*. Pruning turns it into "signals in the retention window" with no change of label — the reader has no way to tell.

**2. `/v1/internal/fleet/analytics?days=` takes an arbitrary window.** `src/api/routes.ts:4686-4689` defaults to 90 but accepts anything, and `computeFleetAnalytics` filters `received_at >= cutoff` from that parameter. With a 90-day prune, `?days=365` would return 90 days of data while reporting a 365-day window — the denominator and the stated window disagreeing, which is exactly the class of defect #9676 / #9718 / #9768 have been correcting on the public surface.

This is the same shape as **#9474**: `orb_pr_outcomes` could not simply be pruned either, because it feeds a cumulative public counter. That was solved by folding expiring rows into `orb_outcome_rollups` *before* deletion, with an explicit ordering constraint documented at `retention.ts:17-23`. `orb_signals` needs the same kind of thought, not a copied line.

## Options

**A. Prune at 90 days, and make both readers honest.** Add the rule; change `signalCount` to a windowed count with a label saying so; clamp the analytics `days` parameter to the retention window (or return the effective window alongside the result so a caller can see it was capped). Smallest change that keeps every published number true.

**B. Prune at 90 days, preserving the cumulative count via a rollup.** Mirror #9474: fold expiring rows into a durable per-instance total before deleting, so `signalCount` stays genuinely lifetime. More work, and it only pays off if that count is actually load-bearing for anyone.

**C. Leave it unbounded, and say so.** Document that `orb_signals` is retained indefinitely by design because two surfaces depend on unbounded history. Honest, but the table only grows, and it is the one ingest path open to any instance before a human registers it (`src/orb/ingest.ts` stores a stranger's signals; registration only gates whether they *count*).

I lean **A**: the cumulative `signalCount` reads like a diagnostic rather than a number anyone depends on, and B's rollup is real machinery to preserve it. But C is defensible if fleet history is considered worth keeping whole — in which case the growth should at least be a known, stated choice rather than an omission.

## Note on scope

Whichever is chosen, the analytics `days` clamp is worth doing **regardless**, because the window/denominator mismatch it allows is already reachable today for any `days` value larger than the data actually retained upstream.

JSONbored:
Closes #9783

## The problem

`orb_signals` had no `RETENTION_POLICY` rule and grew without bound — one row per PR per self-host instance, forever.

It could not simply be added to the policy, because two live surfaces read history the raw rows carry:

| Surface | What a plain prune would do |
|---|---|
| `listFleetInstances` → `signalCount` | Silently redefines a **lifetime** count as "signals in the retention window", with no change of label |
| `/v1/internal/fleet/analytics?days=` and the public weekly fleet trend (#9676, 8-week lookback) | Reports windows it no longer has data for — weeks collapsing to `null` as their rows age out |

That is the same window/denominator mismatch #9676 and #9793 have been correcting, so shipping it here would have been a regression, not a cleanup.

## The fix

The #9474 pattern (`orb_pr_outcomes` → `orb_outcome_rollups`), which exists for exactly this shape: **fold the aging rows into a durable aggregate in the same transaction that deletes them.**

`orb_signal_rollups` keeps the **whole confusion matrix per instance per day**, not a scalar count. The fleet numbers are not a total — `foldInstance` derives `decisionAccuracy` from `(verdict, outcome, reversal_flag, gate_reasoncode_bucket)` cells, and #9775's trend buckets the same cells by day. A scalar rollup would preserve `signalCount` and destroy every accuracy figure past the window; per-day cells preserve **both**, so a historical week reconstructs to the same number it showed while the raw rows existed.

Reads union the two sources — `computeFleetAnalytics`, `loadPublicFleetAccuracyTrend`, `listFleetInstances` — with the registered-instance trust gate applied at **read** time on both halves.

## Decisions worth reviewing

- **Unregistered instances still fold.** Discarding their history at prune time would make a later registration silently lossy, so the trust gate moved to the read side rather than the write side. Tests pin that their folded history still never moves a published number.
- **`gate_verdict` / `gate_reasoncode_bucket` are `NOT NULL DEFAULT ''`,** unlike the nullable source columns. A `PRIMARY KEY` over nullable columns does not work: SQLite does not enforce `NOT NULL` on PK columns of a rowid table and compares every `NULL` as distinct, so two folds of the same day would insert duplicate cells instead of accumulating. `''` is safe for both readers — `foldInstance` only tests `=== "merge"` / `=== "close"` / `=== "policy_action"`, which `''` fails exactly as `NULL` did.
- **Per-PR identity is genuinely dropped** (`repo_hash`/`pr_hash`). Those are per-instance HMACs that cannot be reversed or joined across instances; their only role is the ingest dedup key, which matters while a row can still be re-exported and not after it has aged out.
- **`substr(?1, 1, 10)` on the trend's rollup half.** That query bounds on a full ISO instant while `day` is `YYYY-MM-DD`, and a string prefix sorts *before* the string it prefixes — a bare `day >= ?1` drops the window's first day. `computeFleetAnalytics` does **not** need this, because its cutoff is already date-only. Both are commented, and the trend has a regression test that fails without the fix (verified by reverting it).

## Coverage

`src/db/retention.ts`, `src/orb/analytics.ts`, `src/services/public-fleet-accuracy-trend.ts` all at **100% lines / 100% branches**. The `fleet-admin.ts` change is a SQL edit inside an already-covered function.

Reaching 100% branches required the multi-slice arms, so the fold gets the same three guards the `orb_pr_outcomes` fold has: bounded slices, a run of tied timestamps still making progress (an exclusive boundary selects zero rows and spins forever), and `maxPerTable` capping one run — the last of which also asserts the unioned read is exact at the **intermediate** point, with half the cohort folded and half still raw.

## Validation

`db:migrations:check` (contiguous 0001..0202) · `db:migrations:immutable:check` · `db:schema-drift:check` · `validate:no-hand-written-js` · `dead-source-files:check` · `import-specifiers:check` · `tsc --noEmit` — all clean. 138 tests pass across the three touched suites.

No visual output touched.


JSONbored:
Good catch — that was a real gap, now fixed in efb886ac3.

**Why the fold genuinely can't carry it.** The confusion matrix survives because per-day cells *sum*. `cycleP50Ms`/`cycleP95Ms` are percentiles over individual `time_to_close_ms` values, and percentiles aren't summable — no per-day count reconstructs them. Storing a sample of durations instead would reintroduce exactly the unbounded growth this table exists to fix, so the rollup deliberately has no `time_to_close_ms` column.

**What it does instead.** Past the retention horizon the percentiles go `null`, with a new `cycleTimeObservable` flag saying why. That's the posture `gamingDetectionEligible` and `fleetFramingEligible` already set here: a surface labels the gap rather than being handed a number that quietly means something narrower than its window. Returning a real-looking p50 computed from 90 days of survivors while the caller asked for 365 is the failure you identified; `null` plus a reason is the honest answer.

The horizon comes from `retentionCutoffIsoForTable("orb_signals")` rather than a restated constant, so the window this module trusts can't drift from the one the prune enforces.

**The default window is unaffected.** `windowDays` defaults to 90 and retention is 90, so they're equal and cycle time stays fully observable — the public headline doesn't change. Only `?days=` beyond 90 nulls, which is precisely the range that had no honest answer. Existing consumers (`federated-bundle.ts`, `federated-import.ts`) already type both as `number | null` and validate with `nullableNumeric`, so nothing downstream breaks.

Four tests added, including your exact scenario (`?days=365` against a 90-day-pruned table) and one asserting `decisionAccuracy` still folds correctly over that same long window — the distinction that makes nulling cycle time acceptable rather than a general capitulation on long windows.

`src/orb/analytics.ts` and `src/db/retention.ts` remain at 100% lines / 100% branches.

PR #9804, opened 2026-07-29
fix(eval): stamp the in-Worker threshold backtest with its corpus freeze point

### LINKED ISSUE #9639: orb(eval): the threshold-backtest writer never emits `corpusChecksum

## Context

`loadPublicRulePrecision` (`src/review/public-rule-precision.ts:51`) exposes a `latestBacktestRun` block the
module header calls "the reproducibility hook (#8136's conclusion, operationalized)". It reads:

```sql
SELECT json_extract(metadata_json, '$.corpusChecksum') AS checksum, created_at FROM audit_events
 WHERE event_type IN ('calibration.threshold_backtest_run', 'calibration.logic_backtest_run')
   AND json_extract(metadata_json, '$.corpusChecksum') IS NOT NULL
 ORDER BY created_at DESC LIMIT 1
```

Two event types are in the `IN` list. Only one of them can ever satisfy the `IS NOT NULL` filter:

- `calibration.logic_backtest_run` is written by `buildLogicBacktestAuditInsertSql`
  (`scripts/backtest-logic-check-core.ts:173-190`), whose metadata is
  `{ comparison, headSha, baseSha, corpusChecksum, replayableCount, skippedCount }`. This writer runs only from
  the CI-side `scripts/backtest-logic-check.ts` CLI, and only with `--persist`.
- `calibration.threshold_backtest_run` is written by `persistThresholdBacktestRuns`
  (`src/services/threshold-backtest-run.ts:60-86`), which runs inside the Worker on every review pass whose
  diff touches a `KNOWN_THRESHOLDS` constant. Its metadata is `{ comparison, constantName }` — **no
  `corpusChecksum`, no head/base sha**.

So an ORB deployment that runs the in-Worker threshold backtest but has never had the CI logic-backtest CLI
persist a row has `latestBacktestRun: null` forever. That is not a cosmetic gap:
`buildEvalScoreRecordsFromRulePrecision` (`src/review/eval-score-records.ts:85-86`) early-returns `[]` when
`latestBacktestRun` is null, so the entire `/v1/public/eval-scores` surface — the validator-consumable
`EvalScoreRecord` feed #9266 shipped and #9265 is specified to extend — serves zero records with no
diagnostic.

The two writers are explicitly siblings: `buildLogicBacktestAuditInsertSql`'s own doc calls itself "a sibling
row to #8138's THRESHOLD_BACKTEST_EVENT_TYPE events" and says "`corpusChecksum` + the two shas are the freeze
point (#8136's reproducibility posture)". The threshold writer simply never gained the field, and the reader
was written as if it had.

`persistThresholdBacktestRuns` has everything it needs: `runThresholdBacktestAdvisory`
(`src/services/threshold-backtest-run.ts:44-54`) builds `corpusByRuleId` before calling
`backtestChangedThreshold`, and `checksumCases` (`scripts/backtest-corpus-export-core.ts:23`) is the
established canonicalized-sha256 helper the logic-backtest side already uses.

## Requirements

- `checksumCases` must become importable from the Worker runtime. Move it (and only it, with its
  canonicalization helper) into `packages/loopover-engine/src/calibration/` and re-export it from the engine
  barrel, updating `scripts/backtest-corpus-export-core.ts`, `scripts/backtest-logic-check.ts` and
  `scripts/attested-backtest-run.ts` to import it from `@loopover/engine`. Its output must be byte-identical
  for the same input — an existing corpus manifest checksum must still validate.
- `runThresholdBacktestAdvisory` must compute, per ruleId, the checksum of the corpus slice it actually
  backtested, and thread it through `ThresholdBacktestRunResult` so `persistThresholdBacktestRuns` can write it.
- `persistThresholdBacktestRuns` must write `corpusChecksum` into the audit event's metadata for every
  persisted comparison, alongside the existing `comparison` and `constantName` keys.
- Nothing about the existing `comparison`/`constantName` metadata or the `THRESHOLD_BACKTEST_EVENT_TYPE`
  string may change — `src/services/rule-calibration-trend.ts:187-189` reads `$.comparison.verdict` off these
  same rows and must keep working.

> ⚠️ Required pattern: mirror `buildLogicBacktestAuditInsertSql` in `scripts/backtest-logic-check-core.ts:173-190`
> — the same field name (`corpusChecksum`), the same `checksumCases` producer, the same position in
> `metadata_json`. What does NOT satisfy this issue: inventing a second checksum function or a differently-named
> field; removing `calibration.threshold_backtest_run` from the reader's `IN` list instead of fixing the writer;
> making `loadPublicRulePrecision` fall back to a placeholder or synthetic checksum when none is found; or
> duplicating `checksumCases` into `src/` rather than moving it into the shared engine package.

## Deliverables

- [ ] `checksumCases` lives in `packages/loopover-engine/src/calibration/`, is exported from the engine barrel,
      and every existing `scripts/` importer is updated; a test asserts its output is unchanged for a fixed
      fixture corpus.
- [ ] `persistThresholdBacktestRuns` in `src/services/threshold-backtest-run.ts` writes
      `metadata.corpusChecksum` for each comparison, asserted by a new case in the threshold-backtest-run unit
      test file.
- [ ] A test asserting `loadPublicRulePrecision` returns a non-null `latestBacktestRun` when the only persisted
      run is a `calibration.threshold_backtest_run` row — a test that fails against the current writer.
- [ ] A test asserting `buildEvalScoreRecordsFromRulePrecision` emits one record per rule (not `[]`) for that
      same fixture, so the `/v1/public/eval-scores` regression is pinned end to end.
- [ ] `metadata.comparison` and `metadata.constantName` are still present and unchanged on the persisted event,
      asserted by an existing or new case.

All Deliverables above are required in a single PR. A PR that satisfies only some of them — for example moving
`checksumCases` into the engine without wiring it into `persistThresholdBacktestRuns` — does not resolve this
issue.

## Test Coverage Requirements

This repo enforces 99%+ Codecov patch coverage, branch-counted. `coverage.include` covers both `src/**/*.ts`
and `packages/loopover-engine/src/**/*.ts`, so `src/services/threshold-backtest-run.ts`,
`src/review/public-rule-precision.ts`, `src/review/eval-score-records.ts` AND the moved `checksumCases` in
`packages/loopover-engine/src/calibration/` are all measured and gated at the same 99%+ branch-counted bar —
the move does not exempt it, and every branch of the moved function needs coverage in its new home. Both arms
of every changed conditional need a test. The byte-stability test in Deliverable 1 and the
`loadPublicRulePrecision` regression test in Deliverable 3 are both mandatory.

## Expected Outcome

An ORB deployment that has run a single in-Worker threshold backtest publishes a real `latestBacktestRun`
freeze point, and `/v1/public/eval-scores` emits `EvalScoreRecord`s instead of an empty array — so the
validator-consumable feed reflects the deployment's actual backtest history rather than only CI-side logic runs.

## Links & Resources

`src/review/public-rule-precision.ts:88-105`, `src/services/threshold-backtest-run.ts:56-86`,
`src/review/eval-score-records.ts:85-118`, `scripts/backtest-logic-check-core.ts:164-200`,
`scripts/backtest-corpus-export-core.ts:23`. Related: #9265, #9266, #9215.

JSONbored:
Closes #9639

## The seam

`loadPublicRulePrecision` selects the latest backtest run across two event types, filtered on `corpusChecksum IS NOT NULL`. Only one of the two writers ever satisfied that filter:

| Writer | Runs where | Wrote `corpusChecksum`? |
|---|---|---|
| `buildLogicBacktestAuditInsertSql` (`scripts/backtest-logic-check-core.ts`) | CI-side CLI, only with `--persist` | yes |
| `persistThresholdBacktestRuns` (`src/services/threshold-backtest-run.ts`) | in-Worker, every review pass touching a `KNOWN_THRESHOLDS` constant | **no** |

A deployment whose backtest history is entirely in-Worker therefore had `latestBacktestRun: null` forever — and `buildEvalScoreRecordsFromRulePrecision` early-returns `[]` on that, so the entire `/v1/public/eval-scores` feed served zero records with no diagnostic. Both halves were individually correct; the defect lived only in the seam, which is why the end-to-end test in Deliverable 4 matters more than the unit ones.

## The move, not a copy

`checksumCases` moves into `packages/loopover-engine/src/calibration/` so the Worker can reach it. Copying it into `src/` would have been the worse failure: a manifest frozen by one implementation and validated by the other disagrees for no visible reason, and two CLIs (`backtest-logic-check.ts`, `attested-backtest-run.ts`) re-validate a manifest against this function *before trusting its cases* — so a drifted copy doesn't fail loudly, it silently rejects every corpus ever exported.

`scripts/backtest-corpus-export-core.ts` re-exports it, so its existing importers and its own test are unaffected.

**`backtest-checksum.test.ts` pins the digest of a fixed fixture** — `d980dbee…`, computed from `origin/main`'s pre-move implementation, not from the new one. If that constant ever has to be edited to make the test pass, the freeze point has drifted and every manifest on disk is invalidated. The fixture's keys are deliberately unsorted so the canonicalization is exercised too, and separate cases assert case-order and value changes *do* move the digest.

`node:crypto` rather than Web Crypto because `checksumCases` must stay synchronous (it is called from pure, non-async manifest builders); the Worker runs with `nodejs_compat`, and `attester.ts`, `attestation-envelope.ts`, `backtest-split.ts` and `counterfactual-fixtures.ts` in that same directory already import `createHash` on exactly that basis.

## Details worth reviewing

- **Checksummed over the corpus that was scored**, not re-read at persist time. A second read could return a different corpus (a new override landing mid-run), and a freeze point that doesn't match the cases behind the verdict invites a reader to re-derive different numbers and conclude the published ones are wrong. There's a test that adds an override between the run and the persist and asserts the digest doesn't move.
- **Keyed by ruleId.** Each rule is backtested against its own corpus; one checksum for the whole run would stamp a different rule's freeze point onto this one.
- **A failed corpus read still records a freeze point** (the empty corpus's). That's a true statement about what was backtested, and it keeps a read failure distinguishable from an old deployment that never wrote the field. The *reader* is what declines to publish against `EMPTY_CORPUS_CHECKSUM` — that stays the single gate.
- **`corpusChecksum` is omitted, never nulled,** when a direct caller supplies no map, so the reader's `IS NOT NULL` filter remains the one place that decides usability.
- **`comparison` / `constantName` untouched** — `rule-calibration-trend.ts` reads `$.comparison.verdict` off these rows. Asserted.

## Deliverables

- [x] `checksumCases` in the engine, exported from the barrel + a subpath export, all three `scripts/` importers rewired, byte-stability test on a fixed fixture
- [x] `persistThresholdBacktestRuns` writes `metadata.corpusChecksum`, asserted in `threshold-backtest-run.test.ts`
- [x] `loadPublicRulePrecision` returns non-null `latestBacktestRun` when the only persisted run is a `threshold_backtest_run` — **fails against the old writer** (verified by reverting the one line)
- [x] `buildEvalScoreRecordsFromRulePrecision` emits one record per rule for that same fixture — also fails against the old writer
- [x] `comparison`/`constantName` still present and unchanged

Each of the three regression tests was re-run with the writer line removed; all three fail, so none of them passes vacuously.

## Scope note

This fixes **self-host and any deployment that executes reviews**. It does not by itself make `api.loopover.ai/v1/public/eval-scores` non-empty: hosted review execution is retired (`src/index.ts` acks-and-drops review-execution jobs), so `persistThresholdBacktestRuns` never runs there. The hosted read-path gap is a separate defect — the public corpus for `ai_consensus_defect` currently publishes 460 real cases with a re-derivable checksum while the records array stays empty — and gets its own issue and PR.

## Validation

`tsc --noEmit` · `engine-parity:drift-check` · `dead-source-files:check` · `import-specifiers:check` · `validate:no-hand-written-js` · `manifest:drift-check` — clean. 85 tests pass across the affected suites.

Coverage on every changed measured file — `backtest-checksum.ts`, `threshold-backtest-run.ts`, `public-rule-precision.ts`, `eval-score-records.ts` — **100% lines, 100% branches**. The one `v8 ignore` is the sort comparator's `0` arm, unreachable because `Object.entries` yields each key once; the reachable arms are covered by the property-order test.


JSONbored:
Both correct — fixed in 84b6dd93b.

**The subpath import.** You're right that this reversed the file's own documented rationale, and the consequence is worse than style: `scripts/replay-runner/Dockerfile` constructs `@loopover/engine` from three named `dist/calibration/*.js` files rather than `npm ci`-ing it, so the barrel import would simply have failed to resolve in the container — and had it resolved, it would have widened a trusted computing base that is measured inside a TEE. Now uses `@loopover/engine/calibration/backtest-checksum`, the subpath export this PR already adds.

**The Dockerfile.** `backtest-checksum.js` added to the COPY list. Worth noting it costs nothing in the dimension the comment there actually cares about: compiled, its entire runtime closure is one `node:crypto` import — the `BacktestCase` import is type-only and erased — so the file count goes up by one while the dependency footprint doesn't move.

```
$ grep -n "^import" packages/loopover-engine/dist/calibration/backtest-checksum.js
20:import { createHash } from "node:crypto";
```

**`replay-runner-manifest:check` caught the drift**, exactly as the Dockerfile comment says it will. Regenerated manifest committed alongside.

Also merged current main in and resolved a conflict with #9811 in `test/unit/eval-score-records.test.ts` — both PRs appended a `describe` block; kept both. The two changes compose: #9811 made the commitment fall back per rule when no run is persisted, and this PR makes the in-Worker writer persist one; the "run still wins" precedence #9811 added means this PR's end-to-end test still asserts the persisted checksum.

Engine suite: 889 pass. Affected host suites: 58 pass. `replay-runner-manifest:check`, `import-specifiers:check`, `dead-source-files:check`, `engine-parity:drift-check` all clean.

PR #9821, opened 2026-07-29
feat(gate): choose provider, model, effort and self-consistency runs — per repo, and escalated on guarded paths

JSONbored:
## The asymmetry this fixes

Effort already existed — `CLAUDE_AI_EFFORT` → `claude --effort`, `CODEX_AI_EFFORT` → `model_reasoning_effort`, levels `low|medium|high|xhigh|max`. But **only as a global env var**, while `provider` and `model` were already per-repo manifest fields. Self-consistency runs had exactly the same gap.

So the two knobs that most directly control *review depth* were the two a repo couldn't set. (I'd earlier told the maintainer no effort knob existed at all — that was wrong; it exists, it just wasn't reachable from config.)

## What's added

```yaml
gate:
  aiReview:
    provider: anthropic          # existed
    model: claude-sonnet-5       # existed
    effort: medium               # NEW — per-repo parity with the env var
    selfConsistencyRuns: 0       # NEW — per-repo parity with the env var

  guardrailEscalation:           # NEW — used INSTEAD when a PR hits hardGuardrailGlobs
    effort: high
    selfConsistencyRuns: 3
    # provider / model optional
```

**Precedence resolves per field independently**: escalation > per-repo > global. An escalation setting only `effort` still inherits the repo's model and the global provider — the common request is *"same model, think harder"*, not *"swap the model out"*.

## Why escalation, not just per-repo knobs

A `hardGuardrailGlobs` hit currently changes exactly one thing: it suppresses auto-merge and queues a human. It buys **no extra analysis** — a PR touching `.github/workflows/**` gets the same single-pass, same-model, same-effort review as a README typo. On the production ORB that was **74 distinct PRs held in 14 days**, on repos whose guardrail lists had already been narrowed twice. Automation replaced by a maintainer queue, with scrutiny unchanged.

Now a guarded path can buy more scrutiny instead. Manual review remains the fallback: if the escalated review isn't clean, the PR still holds.

## Safety

- Every layer unset ⇒ `null` throughout ⇒ **byte-identical** to today for every existing repo and self-hoster.
- `selfConsistencyRuns: 0` is a real value, not "unset" — a truthiness check there would silently re-enable multi-run reviews on a repo that deliberately turned them off. Pinned by a test.
- The escalation block is **inert** when the PR touches no guarded path. Pinned by a test.
- `isGuardrailHit` is the shared matcher (`src/signals/change-guardrail.ts`), never a second copy free to drift.
- Escalations log `ai_review_guardrail_escalated`, and `describeReviewEscalation` renders the one-line "why" for the panel and decision record — an auto-merge on a guarded path should be able to say what extra scrutiny it got.

## Tests

9 resolver cases (escalation applies / is inert off-path / per-field precedence / per-repo parity / runs=0 / all-absent / the three description shapes) plus the manifest parse, round-trip and resolve coverage. **1118 pass** across the touched files.

`tsc`, `dead-source-files`, `docs:drift`, `manifest:drift`, `engine-parity:drift`, `ui:openapi:check`, `import-specifiers` all green — including the drift guards that forced the `.loopover.yml.example` documentation and the alias rows for the nested keys.

## Follow-up

`effort`/`model`/`provider` currently reach the provider invocation through `options.claudeEffort`, which already outranks the env var; the remaining wiring for BYOK provider swaps on escalation is small and can follow. The ORB's own configs stay untouched here — flipping `guardrailEscalation` on for the three repos is a server-side config change once this ships in an image.


JSONbored:
## The blocker is stale — it reviewed a commit from 93 minutes before the fix

```
reviewed head : 353127c2d  @ 12:03:37 UTC   ← what this review actually saw
fix landed    : a04ea9724  @ 13:36:06 UTC   ← "consume the resolved model/effort in ai-review.ts itself"
current head  : e98bbb2c2                   ← 6 commits later
```

The original flag was **correct and useful** — the first implementation put the wiring in the orchestration only, so `src/services/ai-review.ts` genuinely never read `reviewKnobs.effort`/`.model`, and its own doc comment overclaimed. That was fixed in `a04ea9724`.

On the current head, `src/services/ai-review.ts` folds them into the `AiRunCorrelation` — which *is* what becomes the provider's per-call options at this module's dispatch:

```ts
claudeModel:  input.reviewKnobs?.model  ?? input.claudeModel  ?? undefined,
claudeEffort: input.reviewKnobs?.effort ?? input.claudeEffort ?? undefined,
codexModel:   input.reviewKnobs?.model  ?? input.codexModel   ?? undefined,
codexEffort:  input.reviewKnobs?.effort ?? input.codexEffort  ?? undefined,
// + ollamaModel / openaiModel / openaiCompatibleModel / anthropicModel, same precedence
```

Same `??` precedence the orchestration applies, so applying it in both places is idempotent — and this module is now self-sufficient for any other caller that passes `reviewKnobs`.

## On `provider` — the reviewer is right, and it's deliberate

`provider` is **not** consumed in `ai-review.ts`, and shouldn't be. It governs **BYOK key selection**, which happens at the orchestration *before* this module runs: an escalated provider that disagrees with the stored key drops that key, so the review can't run one provider's name against another's credential. It rides on the correlation for logging/replay only.

The doc comment now states exactly that, per field and per location, instead of the vague *"the model/effort/provider fields ride through to the provider invocation"* that earned the original flag.

## Verification of the consumption side

Rather than assert precedence, there's an end-to-end test that captures the options the provider actually receives for a guarded-path PR with `guardrailEscalation: { effort: "xhigh", model: "…" }`:

- `claudeEffort === "xhigh"`, `claudeModel === "escalated-model-x"`, `codexEffort === "xhigh"`, `anthropicModel === "escalated-model-x"`
- plus a self-consistency case: env runs **off**, escalation `selfConsistencyRuns: 3`, a guarded changed file → 1 primary + 2 rotated-exemplar calls
- plus the inert invariant: same escalation block, **unguarded** file → zero escalated runs

Engine-side coverage also went **21 → 0** uncovered lines (the engine's own c8 run measures `--all` over `dist/`, which the root vitest report doesn't satisfy — worth knowing for any future engine change).

## Why this review didn't refresh itself

```
github_app.ai_review_one_shot_reuse: reused the last published AI review
                                     instead of spending a fresh call
```

Fired repeatedly over the last ten minutes. The one-shot cadence pins the published review to the first reviewed head to control spend, so it will keep reporting on `353127c2d` until the re-run checkbox is ticked (which sets `forceAiReview` and bypasses both the cache and the manual-review freeze).

Not a defect in this PR — but it's the mechanism behind several "the review says something that's no longer true" reports, so worth naming here.

PR #9823, opened 2026-07-29
fix(test): carry #9813's new argument and cache-key fragment into the six stale assertions (#9822)

### LINKED ISSUE #9822: ci: main is red — #9813's new ignoredCheckRuns argument left six assertions stale
## What

`main` is currently **red**: six tests fail on `origin/main` (`9ec808c47`) with no PR involved. Every PR opened against it inherits the failure, so `validate` fails for work that is itself correct.

```bash
npx vitest run test/unit/queue.test.ts test/unit/agent-approval-queue.test.ts
```

| Test | File |
|---|---|
| surfaces stale optional CI after the stale-CI cap | `test/unit/queue.test.ts` |
| REGRESSION (#selfhost-ci-verification gate review): a swallowed branch-protection read failure never writes the fail-open aggregate through to the durable cache | `test/unit/queue.test.ts` |
| threads the repo's expectedCiContexts into the accept-time live CI re-check | `test/unit/agent-approval-queue.test.ts` |
| unions branch-protection contexts into the accept-time live CI re-check | `test/unit/agent-approval-queue.test.ts` |
| falls back to expectedCiContexts when the accept-time branch-protection read fails | `test/unit/agent-approval-queue.test.ts` |
| (plus the `ciRequiredContextsKey` expectation in the second queue case) | `test/unit/queue.test.ts` |

## Why

#9813 (`gate.ignoredCheckRuns`) made two shape changes and did not carry them into these expectations:

1. `fetchLiveCiAggregatePreferGraphQl` gained an **8th** parameter, `ignoredCheckRuns`. The assertions still list seven arguments.
2. The live-fact cache key gained an `|ign:` fragment (`ci-resolution.ts` builds `…|adv:…|ign:…`). One expectation still ends at `|adv:`.

Two of the accept-path assertions also still expect `undefined` for `advisoryCheckRuns` where the settings resolver now supplies `null` — the resolver returns the column's value rather than an absence, and `toHaveBeenCalledWith` distinguishes the two.

**The production code is correct.** Only the expectations are stale, which is why nothing is broken at runtime and why this went unnoticed: #9813's own CI ran before the later merges that made both call paths exercise the new argument.

## Fix

Update the six expectations to the real signature and key shape. No source change.

JSONbored:
Closes #9822.

**`main` is red on its own.** Six tests fail at `9ec808c47` with no PR involved, so every PR opened against it inherits a `validate` failure for work that is itself correct — including #9812 and #9815, which is how I found this.

```bash
npx vitest run test/unit/queue.test.ts test/unit/agent-approval-queue.test.ts
```

## Cause

#9813 (`gate.ignoredCheckRuns`) made two shape changes and did not carry them into the expectations:

1. **`fetchLiveCiAggregatePreferGraphQl` gained an 8th parameter**, `ignoredCheckRuns` (`src/queue/ci-resolution.ts:206` passes it; `src/github/backfill.ts:3434` declares it). Four assertions still list seven arguments.
2. **The live-fact cache key gained an `|ign:` fragment** — `ci-resolution.ts` now builds `…|adv:…|ign:…`. One expectation still ends at `|adv:`.

Two accept-path assertions additionally expected `undefined` for `advisoryCheckRuns` where the settings resolver supplies `null`: the resolver returns the column's value rather than an absence, and `toHaveBeenCalledWith` distinguishes the two. Note the two call paths legitimately differ — the queue processor's direct call passes `undefined`, the settings-resolved accept path passes `null` — so the expectations differ too, deliberately.

## Scope

**Expectations only — no source change.** The production code is correct, which is why nothing is broken at runtime, and why this went unnoticed: #9813's own CI ran before the merges that made both call paths exercise the new argument.

## Validation

`npx vitest run test/unit/queue.test.ts test/unit/agent-approval-queue.test.ts test/unit/ci-resolution.test.ts test/unit/pr-detail-durable-cache.test.ts` → **347 passed**, from 6 failing before.

### DIFF
--- a/config/examples/loopover.full.yml
+++ b/config/examples/loopover.full.yml
@@ -290,6 +290,25 @@ gate:
     - name: Contributor trust
       appSlug: example-security-app
 
+  # Check-runs to IGNORE ENTIRELY (#9810) — the stronger sibling of advisoryCheckRuns above. Same
+  # spoof-resistant { name, appSlug } matching, but a matched run is treated as if it did not exist: it never
+  # gates CI, never counts as "still running", and — unlike advisory — never routes the PR to a manual-review
+  # hold either. Its conclusion is surfaced informationally only.
+  #
+  # Use this when a check's verdict carries no signal for YOUR repo while OTHER checks from the same app stay
+  # meaningful. The motivating case: a vendor app publishes both a real security scan AND a heuristic
+  # contributor-trust score. The scan is worth gating on; the trust score fails for perfectly good
+  # contributors, and listing it under advisoryCheckRuns still converts every one of their otherwise-clean PRs
+  # into a manual review — automation replaced by a queue of human decisions, and contributors left wondering
+  # whether they are being judged fairly. Ignoring the trust check keeps the scan's protection and drops the
+  # noise. If BOTH lists name the same check, ignore wins (it is the stronger, more explicit intent).
+  #
+  # List of { name, appSlug }, or omit. Default: not configured (byte-identical behavior for every repo that
+  # doesn't opt in). Config-as-code only — no DB column or dashboard toggle.
+  ignoredCheckRuns:
+    - name: Contributor trust
+      appSlug: example-security-app
+
   # Promote a confident AI-judgment-only finding (one the reviewer itself placed under "Blockers", never
   # a "Nit") into a real, deterministic gate blocker instead of leaving it advisory (#3907). Only matters
   # for repos already running the registry content lane (see contentLane below) — content/registry repos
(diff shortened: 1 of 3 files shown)


JSONbored:
Superseded — the identical fix landed in #9829 (merged as `9d7f56727`), which is why this now shows conflicting: the same three files were already corrected on `main`.

Two sessions found #9813's fallout independently and filed within ~35 minutes of each other (#9822 / #9828). Same diagnosis, same three causes, same file set. Closing this one as the redundant half; nothing here is lost.

PR #9831, opened 2026-07-29
fix(test): repair main — #9813 changed three contracts and updated only some callers

JSONbored:
## main is red

#9813 changed three things and updated only some of their callers:

| change | missed caller |
|---|---|
| `ignoredCheckRuns` added as an **8th arg** to `fetchLiveCiAggregate*` | `agent-approval-queue.test.ts` ×3, `queue.test.ts` — `toHaveBeenCalledWith` pinned the 7-arg shape |
| `\|ign:` fragment added to the durable CI-cache key | `queue.test.ts` asserted `\|adv:` only |
| `ignoredCheckRuns` documented in `.loopover.yml.example` | `config/examples/loopover.full.yml` must match it from the `WHERE IT LIVES` marker onward |

Six failures, on `main`, independent of any open PR — confirmed by checking out `origin/main` clean and reproducing.

## My error, and the specific process failure

I verified #9813 by running the **touched test files**, not the full suite. All three broken files assert on contracts that PR changed while living elsewhere in the tree, so none of them ran before merge. Targeted runs cannot catch a contract change's blast radius — that is exactly what the full suite is for.

## The fixes

- Add the 8th argument to the four call-shape assertions, each commented with what it is.
- Update the cache-key assertion to `\|adv:\|ign:`, and rewrite the comment to explain *why* both fingerprints are folded in (a config change to either must invalidate the cached aggregate rather than serve a stale verdict) — the old comment named only the advisory half.
- Port the `ignoredCheckRuns` block into `config/examples/loopover.full.yml` so the two examples are byte-identical from the marker onward, which is what the test actually requires.

No production code changes — these are test/doc contracts catching up with a shipped change.

## Verification

Full unit suite: **1314 files, 25,209 tests passed**, 3 skipped, zero failures. `tsc` clean.

One thing worth recording: `mcp-cli-completion-spec` and `mcp-cli-response-validation` fail on a bare `vitest run` unless `npm run build:mcp` has been run first — they exercise the built CLI. CI builds it as part of `test:ci`; a local full-suite run needs it explicitly, or you get three failures that look real and are not.

### DIFF
--- a/config/examples/loopover.full.yml
+++ b/config/examples/loopover.full.yml
@@ -290,6 +290,25 @@ gate:
     - name: Contributor trust
       appSlug: example-security-app
 
+  # Check-runs to IGNORE ENTIRELY (#9810) — the stronger sibling of advisoryCheckRuns above. Same
+  # spoof-resistant { name, appSlug } matching, but a matched run is treated as if it did not exist: it never
+  # gates CI, never counts as "still running", and — unlike advisory — never routes the PR to a manual-review
+  # hold either. Its conclusion is surfaced informationally only.
+  #
+  # Use this when a check's verdict carries no signal for YOUR repo while OTHER checks from the same app stay
+  # meaningful. The motivating case: a vendor app publishes both a real security scan AND a heuristic
+  # contributor-trust score. The scan is worth gating on; the trust score fails for perfectly good
+  # contributors, and listing it under advisoryCheckRuns still converts every one of their otherwise-clean PRs
+  # into a manual review — automation replaced by a queue of human decisions, and contributors left wondering
+  # whether they are being judged fairly. Ignoring the trust check keeps the scan's protection and drops the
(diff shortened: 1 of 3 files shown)


JSONbored:
Superseded by #9829, which landed the same three fixes first (the 8th-arg assertions, the `|adv:|ign:` cache key, and the `config/examples/loopover.full.yml` sync).

Verified against `origin/main` at 9d7f56727 before closing: full unit suite **1314 files, 25,209 tests passed**, zero failures. Nothing from this branch is still needed.

Worth carrying forward from the post-mortem regardless of which PR fixed it: the root cause was verifying #9813 with targeted test-file runs instead of the full suite, so three files asserting on the contracts it changed never ran before merge. Also note `mcp-cli-*` needs `npm run build:mcp` before a bare `vitest run` or it reports three failures that aren't real.

Issue #9844, opened 2026-07-29
orb(anchor): the default Rekor shard was never deployed, and the failure it causes does not name the URL


JSONbored:
> ⚠️ Definition of Done: this issue must be completed in full, in a single PR.

## Problem

`DEFAULT_REKOR_SHARD_BASE_URL` in `src/review/ledger-anchor-rekor.ts` is
`https://log2026-1.rekor.sigstore.dev`. **That host does not exist.** Verified from a live self-host
deployment:

```
log2026-1.rekor.sigstore.dev  -> DNS: NXDOMAIN
log2025-1.rekor.sigstore.dev  -> DNS: 34.160.235.130  POST /api/v2/log/entries: 400   (endpoint live)
```

Sigstore shards Rekor v2 annually as `log<year>-<rev>.rekor.sigstore.dev`, and the 2026 shard has not been
deployed — 2025-1 is still the live one. The constant's own comment says it is "the current shard as of when
this was written", but it was written ahead of a shard that never shipped.

So **every operator who enables anchoring without overriding the env var gets a failed anchor**, forever,
with no successful anchor ever published.

## Second, compounding problem

The recorded failure is unhelpful. This was a real, published anchor row from that deployment:

```json
{"seq":1626,"keyId":"6b6490126ad44b51","backend":"rekor","status":"failed","error":"fetch failed"}
```

`"fetch failed"` is Node's generic message. It does not name the URL, so an operator cannot distinguish
"the shard hostname does not resolve" from "my egress is blocked" from "the shard is down" — the three have
completely different fixes. Diagnosing it required probing DNS from inside the container by hand.

That matters more here than in an ordinary code path: #9271 made anchoring failures *publicly* visible
precisely so "anchoring has been failing for a week" is checkable by anyone. A published failure that does
not say what failed only half-delivers that.

## Requirements

- Default to the shard that is actually live, with a comment recording that the year-shard pattern means this
  WILL need updating and that the env var is the intended override.
- The recorded `error` must name the shard URL that was attempted, for the thrown-error path (the `!response.ok`
  path already includes the status and body, so it is already actionable).
- Do not change the request shape, the timeout, or the `recordLedgerAnchorAttempt` contract — the raw caught
  value is deliberately passed through so the persistence layer stays the single place that normalizes an
  unknown error (see the existing comment).

## Deliverables

- [ ] `DEFAULT_REKOR_SHARD_BASE_URL` points at the live shard.
[...]

PR #9849, opened 2026-07-29
docs(selfhost): document ledger anchoring where operators actually look

### LINKED ISSUE #9848: docs(selfhost): .env.selfhost.example never mentions ledger anchoring, the one feature self-host uniquely can run
## Problem

`.env.selfhost.example` documents provider selection, telemetry, and the rest of the self-host surface, but
says nothing about `LOOPOVER_LEDGER_ANCHOR_*`. Anchoring is the one capability a self-host deployment
uniquely *can* run — the hosted deployment retired review execution, so its `decision_ledger` is empty and
there is nothing there to anchor — and the sample file an operator actually copies never mentions it.

The variables are documented in `self-hosting-configuration.mdx`, but nobody reads a docs table to discover
a feature they don't know exists.

## What this cost, concretely

Provisioning anchoring on a live instance (1,626 ledger rows) surfaced three separate stumbling blocks that
the sample file is the natural place to prevent:

1. **All-or-nothing.** Setting one half makes anchoring silently never run. Preflight flags it, but only if
   you already knew to set anything.
2. **The PEM must be single-line with `\n` escapes.** A real multi-line value breaks Docker's `env_file`
   parsing. (`pemToBytes` handles the escapes; nothing says so where an operator would look.)
3. **The Rekor shard rotates annually** and the built-in default can only ever lag. When it does, anchors
   record `status: failed` and the operator has no pointer toward the override that fixes it (#9844).

## Fix

A section in `.env.selfhost.example` covering: what anchoring buys over the plain hash chain, that it is
self-host territory and the ledger is per-instance, `npm run ledger:anchor-keygen`, the all-or-nothing
constraint, the `\n`-escape requirement, the shard override with the live value, and the git backend.

## Deliverable

- [ ] Section added; `docker-compose-env-example-parity` and `selfhost:config-lint` still pass.

JSONbored:
Closes #9848

`.env.selfhost.example` covered provider selection and telemetry but never mentioned `LOOPOVER_LEDGER_ANCHOR_*` — despite anchoring being the one capability a self-host deployment uniquely *can* run. The hosted deployment retired review execution, so its `decision_ledger` is empty and there is nothing there to anchor.

The variables were documented in `self-hosting-configuration.mdx`, but nobody reads a docs table to discover a feature they don't know exists.

## Written from actually doing it

Provisioning anchoring on a live instance (1,626 ledger rows) hit three things in sequence, each of which this file is the natural place to prevent:

- **All-or-nothing.** Setting one half makes anchoring silently never run. Preflight catches it — but only once you knew to set anything at all.
- **The PEM must be single-line with `\n` escapes.** A real multi-line value breaks Docker's `env_file` parsing. `pemToBytes` handles the escapes, but nothing said so anywhere an operator would look.
- **The Rekor shard rotates annually.** When the default lags, anchors record `status: failed` and there was no pointer toward the override that fixes it (#9844, fixed separately).

Also states plainly that the ledger is per-instance — only your decisions, never other operators' — since that was the first question asked when turning it on.

## Validation

`docker-compose-env-example-parity` and `selfhost-preflight` suites pass (44 tests). `selfhost:config-lint` clean. Comments-only addition; every line is commented out, so no default behaviour changes.

### DIFF
--- a/.env.selfhost.example
+++ b/.env.selfhost.example
@@ -111,3 +111,40 @@ LOOPOVER_REVIEW_ENRICHMENT=false
 # The one way to disable it is the explicit air-gap flag below; see .env.example's Orb section for
 # full privacy detail.
 # ORB_AIR_GAP=true                            # air-gapped/OFFLINE deployments only: compute locally, never send
+
+# =============================================================================
+# 6. Decision-ledger anchoring — optional, and the one place self-host has that
+#    the hosted deployment does not
+# =============================================================================
+# Every verdict this instance persists appends to a hash-chained ledger, which already makes
+# tampering WITHIN the chain detectable. Anchoring additionally publishes a signed checkpoint of the
+# chain's tip somewhere you do not control, so rewriting history back past a published anchor means
+# forging a signature at an external log too -- not just editing your own database.
+#
+# This is genuinely self-host territory: the hosted deployment does not execute reviews, so its
+# decision_ledger is empty and there is nothing there to anchor. Your instance's ledger is YOURS --
+# only your decisions, never other operators'.
+#
+# Generate both halves with the encodings the runtime expects:
+#
+#   npm run ledger:anchor-keygen
+#
+# OPT-IN, but ALL-OR-NOTHING: setting only one half makes anchoring silently never run (preflight
+# flags this, see src/selfhost/preflight.ts). Leave both unset to keep anchoring off.
+# LOOPOVER_LEDGER_ANCHOR_KEYS=              # public half: the JSON array, served by /v1/public/decision-ledger/anchor-key
+# LOOPOVER_LEDGER_ANCHOR_PRIVATE_KEY=       # SECRET, PKCS8 PEM. Must be ONE line, with the newlines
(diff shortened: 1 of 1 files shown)


JSONbored:
Already fixed — that review ran against `60d900aed`, which I force-pushed over with `de8394059` before it posted.

You're right that it was exactly the pitfall the section warns about: a `\n` inside my heredoc became a real newline, splitting the comment and leaving a bare uncommented line plus trailing whitespace. The `changes` job caught it, which is how I noticed.

On the current head:

```
trailing-whitespace lines: NONE
uncommented lines in section: NONE

'# LOOPOVER_LEDGER_ANCHOR_PRIVATE_KEY=       # SECRET, PKCS8 PEM. Must be ONE line, with the newlines'
'                                            # written as literal backslash-n escapes: a real multi-line'
'                                            # value breaks Docker'\''s env_file parsing.'
```

Rewritten as prose across three properly-commented lines rather than embedding an escape sequence in a comment that then has to survive every tool between here and the file. All checks green.

Issue #9850, opened 2026-07-29
ledger: /verify reports ok:false — content_mismatch at seq 5 on a 3-day-old chain


JSONbored:
## Observed on a live self-host instance

```
GET /v1/public/decision-ledger/verify
{"ok":false,"checked":4,"tipSeq":1639,"totalCount":1639,"prunedRecords":0,
 "break":{"kind":"content_mismatch","atSeq":5,
          "recordId":"record:JSONbored/loopover#8921@a2446db8e808529d19803ada464cc9d605a89638"}}
```

Verification stops at the first break, so **4 of 1,639 rows are checked and the endpoint publicly reports
`ok: false`.** Anyone running the documented verification against this deployment sees a chain break.

## Why this is probably not tampering

The chain is only three days old — `seq 1 @ 2026-07-26T13:40:09Z`, `seq 1639 @ 2026-07-29T14:01:03Z` — and
`prunedRecords: 0`, so the legitimate pruned-preimage explanation does not apply. The instance has been
tracking beta images across that window (currently `orb-v3.6.0-beta.6`).

The likely cause is that the decision-record preimage changed between the release that wrote seq 5 and the
release verifying it now. Recomputing the digest over the current shape yields a different value than the one
committed at write time, and the content check correctly reports a mismatch.

## Why that matters more than one bad row

If the record preimage is not stable across versions, then **every schema change silently invalidates the
content check for every row written before it** — and the published `/verify` endpoint starts reporting
`ok: false` on an untampered ledger. The chain-structure guarantees (sequence, prevHash, rowHash) still hold;
it is specifically the content re-check that breaks.

That inverts what the surface is for. A verifier cannot distinguish "someone edited a record" from "the
software changed shape", which is exactly the discrimination the ledger exists to provide.

## What needs deciding

- Is the record preimage intended to be version-stable? If yes, this is a bug in whatever changed it, and
  the fix is a versioned preimage (a `schemaVersion` in the digest input, with the verifier recomputing under
  the row's own version) rather than a re-hash of history.
- If it is NOT intended to be stable, `/verify` needs to say so — the same way `prunedRecords` already
  carves out legitimately-unre-checkable rows — so a shape change does not read as tampering.

## Reproduce

Any deployment whose ledger spans a release that altered the decision-record shape. Confirmed above on a
3-day-old chain.

## Links

`src/review/decision-record.ts`, `src/review/ledger-anchor.ts`, the `/v1/public/decision-ledger/verify`
[...]

Issue #9123, opened 2026-07-27
orb(trust): the decision record is never published — contributors get a 12-char prefix of a document no endpoint returns, and 51 preimages are already destroyed


JSONbored:
## Problem

The content-addressed decision record is persisted but **never published**. A contributor is shown a
12-character prefix of a SHA-256 over a document they cannot obtain, cannot recompute, and cannot dispute.

`grep -rln "decision_records" src/` returns exactly three files — `decision-record.ts`,
`risk-control-wire.ts`, `salvageability-wire.ts`. **No API route, no MCP tool, and no export path ever
returns `record_json`.** The only published artifact is `renderDecisionRecordSection`
(`src/review/decision-record.ts:148-159`), which truncates everything it shows:

```ts
149:  const short = (digest: string): string => digest.slice(0, 12);
156:  `- **record**: \`${short(recordDigest)}\` (schema v${record.schemaVersion}, head \`${record.headSha.slice(0, 7)}\`)`
```

and omits `decidedAt`, `baseSha`, `salvageability`, `repoFullName`, and `pullNumber` entirely.

The module's own stated purpose (`:6-9`) is that *"'the bot closed me' becomes 'clause X of ruleset
abc123 closed me' — inspectable, arguable, and stable under challenge."* The human-readable clause is
delivered. The cryptographic half is decorative: there is nothing to fetch, nothing to recompute, and
therefore nothing to argue with.

**Compounding — the preimage is destroyed on re-gate.** `persistDecisionRecord` upserts:

```ts
121:  ON CONFLICT(id) DO UPDATE SET action = excluded.action, ..., record_json = excluded.record_json
```

`id` is keyed on head SHA, so a re-gate at the *same* head (sweep re-gate, CI-completion re-review)
overwrites the earlier record body while `appendDecisionLedger` appends a second chain row for the *old*
digest. The comment claims supersessions are *"visible history, not silent replacement"* — but only the
digest survives; the superseded content is gone. Measured on the live instance (`edge-nl-01`,
2026-07-26): **51 of 184 chain rows already reference a digest that no longer matches the current record
body.** Those 51 are permanently unreconcilable against any content — a correct fix for #9078 would be
structurally unable to check them.

## Trigger

A contributor is auto-closed, reads `**record**: a1b2c3d4e5f6` in the panel, and asks to see the record
behind that digest. There is no endpoint. If the PR was re-gated at the same SHA, the row no longer holds
what was decided the first time.

## Impact

[...]

PR #9863, opened 2026-07-29
feat(ledger): declare unrecoverable rows instead of failing forever

JSONbored:
Refs #9850 · #9857 has merged; this branch was rebuilt directly on main and now applies cleanly on its own.

## The problem this closes

Rows whose record preimage is genuinely gone — the 83 rows (seq 5–257) left by the record-overwriting `UPDATE` that #9123 replaced — can never be reconciled. Rewriting those records so the digests match would be **exactly the tampering this ledger exists to detect**.

#9857 stopped one such row from blinding the rest of the chain. It did not solve the remaining problem: `/verify` reports `ok: false` forever, and a permanently-red endpoint is one nobody reads — a worse public signal than a green one carrying an explicit, counted exclusion.

## The mechanism

`LOOPOVER_LEDGER_CONTENT_WAIVER`, format `<fromSeq>-<toSeq>:<reason>`. Four properties make this a disclosure rather than a cover-up. Each is enforced in code and pinned by a test:

| Property | Why it matters |
|---|---|
| **Bounded by seq, not time** | A date boundary *drifts* — as the clock advances it silently swallows new damage. A seq range is fixed forever; only an explicit edit widens it. (The retention cutoff gets to be time-based because it tracks a published policy that genuinely moves. This does not.) |
| **Both ends required** | An open-ended waiver is a blanket exemption wearing a range's clothes. |
| **A reason is mandatory** | You cannot waive silently, and the text is published alongside the count. |
| **Content only** | The content check runs *after* sequence/predecessor/row_hash have passed, so a waived row that is mis-chained **still fails**. There is a test for exactly this. |

## Deliberate design choices

- **Counted separately, never folded in.** `waivedContentMismatches` is its own field; `contentMismatches` stays 0 when everything is declared. "83 rows are excused" must never render as "83 rows are fine".
- **Published even on a clean chain.** The declaration shouldn't require a failure to become visible — a reader can see what's excluded before anything goes wrong.
- **Per-deployment, not in the codebase.** This is one operator's damage. Hardcoding a seq range into the product would assert something about every operator's ledger.
- **Fails closed.** A malformed value waives *nothing* rather than widening the exclusion on a typo — and preflight surfaces it, because a set-but-ineffective value is the dangerous one (same reasoning as the half-configured-anchoring check it sits beside).

## Tests

Eleven cases. The load-bearing ones:

- a mismatch **outside** the range still fails — the bound is real, not paperwork
- a waiver **never excuses a structural break** inside its own range
- a malformed waiver waives nothing and the failure is still reported
- reason and both bounds are mandatory; descending/zero-based ranges rejected
- waived and unwaived counts stay separate when both are present

## Validation

`tsc --noEmit` · `selfhost:config-lint` · `docs:drift-check` · `import-specifiers:check` · `dead-source-files:check` — clean. 95 tests across decision-record, preflight and the public ledger routes.

Every line and branch I changed in `decision-record.ts` and `preflight.ts` is covered — verified against the coverage JSON restricted to the changed ranges. The one uncovered branch in the file (line 700) is outside them and pre-existing.

OpenAPI description updated, `apps/loopover-ui/public/openapi.json` regenerated in the same commit, and `.env.selfhost.example` documents the format and the reasoning.

## For your instance

Once this ships, the declaration would be:

```
LOOPOVER_LEDGER_CONTENT_WAIVER=5-257:rows written before the #9123 record-overwrite fix
```

I have **not** set it on edge-nl-01 — that's a claim about your ledger's integrity, and it should be your call to publish, not mine to apply.


JSONbored:
Rebuilt on current `main` now that #9857 has merged — the conflicts were both sides carrying that change (squashed on main, a commit here). Rather than resolve 13 conflict hunks across four files, I reset to main and cherry-picked the waiver commit alone; it applied clean.

Re-verified after the rebuild: `tsc --noEmit`, 113 tests across decision-record / preflight / public-ledger-routes, and `ui:openapi:check`, `selfhost:config-lint`, `docs:drift-check`, `import-specifiers:check`, `dead-source-files:check` all clean.

Issue #9860, opened 2026-07-29
Generate or guard the remaining hand-maintained lists — three shipped red-main incidents traced to one class


JSONbored:
## Why

Three separate incidents in a single day, all the same shape: **a correct change whose blast radius was tracked by a hand-maintained list that nobody re-derived.**

| incident | the hand-maintained thing | how it failed |
|---|---|---|
| #9813 → main red | test assertions pinning a call arity + a cache-key literal + the two `.loopover.yml` example copies | 8th arg added; three assertions never updated; six failures on main |
| #9815 → main red | miner-UI mock relying on an inferred shape | `ChatActionDispatchResult` closed into a union; the mock widened and stopped satisfying it; `ui:typecheck` isn't in the root `typecheck` |
| #9848 (found, unshipped) | `turbo.json`'s `//#typecheck` `inputs` | four statically-imported paths unhashed → stale-cache risk; the comment's own stated reason for excluding one was **false** |

The last one is the tell: the comment said *"a snapshot … not a structural guarantee — re-run the same grep before trusting this list again."* Nobody re-runs a grep on request. That list is now computed by `scripts/check-turbo-typecheck-inputs.ts` (#9853), which is the pattern this issue asks to apply more broadly.

## What already has a guard (do not redo)

- `MIGRATION_REBASELINE` → `db:migrations:immutable:check`
- `no-hand-written-js` allowlist → `validate:no-hand-written-js`
- codecov type-only ignores → `test/unit/codecov-ignore-type-only.test.ts`
- turbo typecheck inputs → `turbo-inputs:check` (#9853)

## Candidates to audit

1. **The four parallel MCP command lists** — `STDIO_TOOL_DESCRIPTORS`, `CLI_COMMAND_SPEC`, `printHelp`, and the README block are maintained by hand and already documented as having "bitten once". Partial parity checks exist (`mcp-cli-completion-spec`); a single generated projection would remove the class.
2. **`packages/loopover-mcp`'s ~85 hand-mirrored zod shapes** — the comments admit they mirror the remote server's. Now that `@loopover/contract` is published and shipping 12 shared modules, most could be imported instead of re-declared.
3. **`test:ci`'s ~50-command chain in `package.json`** — order and membership are hand-kept; a check that every `scripts/check-*.ts` is actually wired in would catch a checker that lands and is never run.
[...]

PR #9932, opened 2026-07-30
chore(engine): one maintainer-association predicate, with a guard against the ninth copy

JSONbored:
Toward #9860 (candidate class: hand-maintained lists). Follow-up to #9877.

## What verifying #9877's merge turned up

#9877 consolidated **five** spellings of `["OWNER", "MEMBER", "COLLABORATOR"]`. Checking that the merge landed intact, I grepped for the literal again and found three more — so the real count was **eight**:

| Site | Note |
|---|---|
| `packages/loopover-engine/src/scoring/pending-pr-scenarios.ts` | Its own comment: *"Duplicated here (rather than imported) because this package cannot reach into `src/`; keep the two in sync by hand."* That reason expired the moment #9877 moved the definition **into** the engine. |
| `src/services/contributor-evidence-graph.ts` | Spelled with `.toUpperCase()` |
| `src/services/outcome-calibration.ts` | **Found by the new guard, not by me** — on its first run |

**Two of the eight had already drifted**: one compared case-sensitively, one case-insensitively. That is how a copy stops being a copy, and it is the argument for a checker rather than another sweep.

## The one that isn't membership

`packages/loopover-engine/src/signals/engine.ts`'s `strongestAssociation` depends on the list's **order** (OWNER > MEMBER > COLLABORATOR), not on membership. It iterates the shared constant now and says so in a comment, so the ordering dependency is explicit rather than implied by a re-typed literal.

## The guard

`scripts/check-maintainer-association-copies.ts`, wired into `test:ci`. Per #9860's own bar — *compute the fact, don't remember it* — it greps for the three-value maintainer subset rather than listing known offenders, and fails with the import to use:

```
Found 1 re-typed maintainer-association literal(s) (#9860):
  src/services/outcome-calibration.ts:79 — const MAINTAINER_AUTHOR_ASSOCIATIONS = new Set([...]);

Import from packages/loopover-engine/src/settings/author-association.ts instead:
  isMaintainerAuthorAssociation(association)  — membership, case-insensitive
  MAINTAINER_AUTHOR_ASSOCIATIONS             — the list, in precedence order
  classifyAuthorAssociation(association)     — maintainer | contributor | unknown
```

It deliberately does **not** flag `z.enum` declarations of GitHub's full eight-value vocabulary (`CONTRIBUTOR`, `FIRST_TIMER`, `MANNEQUIN`, `NONE`, …). Those are a wire schema, not this predicate, and the contract legitimately declares that enum four times — conflating the two would be its own mistake. That exclusion is tested.

## Behaviour note

`pending-pr-scenarios.ts` and `outcome-calibration.ts` were case-**sensitive**; the shared predicate is case-insensitive and trims. GitHub always sends upper-case, so this is a widening with no practical change — and it removes a latent mismatch where a lower-cased stored value would have been missed by two of the eight and caught by the other six.

## Validation

- 6 new unit tests on the guard itself, against a synthetic tree so both halves are pinned: what it must catch (any quoting/spacing), and what it must not (the full-vocabulary enum, the definition module). One asserts it passes on the **real** repository, which is the point of wiring it into `test:ci`.
- 280 unit tests across the touched modules pass; the engine's own suite is 911 pass / 0 fail.
- `maintainer-associations:check`, `engine-parity:drift-check`, `dead-exports:check`, `dead-source-files:check`, `import-specifiers:check`, `docs:drift-check`, `db:migrations:check`, `tsc --noEmit`, `git diff --check` all clean.


JSONbored:
The blocker is correct, and the miss was in the guard rather than only in the migration. Fixed in the latest push — thank you, this was the right catch.

## What was wrong

`MAINTAINER_TRIPLE` only matched comma-separated array literals, so it could never have found

```ts
function isMaintainerAssociation(value: string | null | undefined): boolean {
  return value === "OWNER" || value === "MEMBER" || value === "COLLABORATOR";
}
```

which was sitting in `signals/engine.ts` with **14 call sites** — in the same file whose `strongestAssociation` I *did* migrate. A guard that recognises one spelling of the thing it guards is a hand-maintained list with extra steps, which is exactly what #9860 says to reject.

## What changed

Detection is now **"all three named within a 4-line window"** — form-agnostic, and it survives a formatter wrapping a chain across lines.

Widening it immediately surfaced **four more genuine copies** that nothing had flagged, none of which I had found by reading:

| File | Form |
|---|---|
| `signals/issue-quality-report.ts` | identical `===`/`\|\|` chain |
| `signals/predicted-gate-engine.ts` | identical chain |
| `services/control-panel-roles.ts` | identical chain |
| `services/recommendation-outcomes.ts` | same predicate inside `isMaintainerLane` |

So the real count was **thirteen**, not eight. All are migrated.

## One site deliberately not migrated

`settings/command-authorization.ts` names all three but is **not** this predicate:

```ts
if (args.commenterAssociation === "OWNER" || args.commenterAssociation === "MEMBER") roles.push("maintainer");
if (args.commenterAssociation === "COLLABORATOR") roles.push("collaborator");
```

`COLLABORATOR` is a *separate role* there. Collapsing it into the maintainer predicate would grant collaborators maintainer-only commands — a privilege escalation, not a cleanup. It is allowlisted in `ALLOWED_DISTINCT_SEMANTICS` with that reason, following `check-regate-sort-key.ts`'s `ALLOWED_OMISSIONS` idiom. That is an exception list with stated reasons, not a list of known duplicates awaiting migration — the distinction #9860 draws.

## Also fixed

The guard printed the *window's first line*, which is often a closing brace or blank, instead of the line that actually names an association. Its output was unreadable in exactly the case it was built for.

## Validation

- 11 guard tests (up from 6), including the `===`/`||` chain, a formatter-wrapped chain, the allowlisted different-semantics case, and one asserting the reported line is the one naming an association. The windowing trade-off — two hits inside one window collapse to one finding — is now stated in a test rather than left implicit.
- Engine suite 911 pass / 0 fail; 124 tests across the touched modules pass.
- `maintainer-associations:check`, `engine-parity:drift-check`, `dead-exports:check`, `import-specifiers:check`, `docs:drift-check`, `tsc --noEmit`, `git diff --check` all clean.

PR #9935, opened 2026-07-30
feat(ledger): let an operator declare historical failed appends, so /verify can go green again

### LINKED ISSUE #9933: A historical failed-append leaves /verify permanently red with no way to declare it
The public ledger verifier on the production ORB reports `ok:false` and cannot be made honest-green, because one of its two findings has no declaration mechanism.

## Measured on the production ORB (orb-v3.7.0-beta.1, 2069 ledger rows)

**1. 83 content mismatches — declarable today.** All in `seq 5..257`, all in `schemaVersion` 2 and 3; every one of the 1,569 v5 rows verifies clean. I recomputed the digests offline with the repo's own `canonicalJson`/`contentDigest` and confirmed the count and range exactly, then brute-forced every subset of null-valued keys per row (the obvious "stored as null, hashed as undefined" hypothesis) — **no subset reproduces the committed digest for any of the 83**. Their preimages are genuinely unrecoverable, which is precisely what `LOOPOVER_LEDGER_CONTENT_WAIVER` (#9850) exists for.

**2. 231 unchained records — NOT declarable.** `decision_records` rows with no `decision_ledger` entry: the failed-append signature. Spanning `2026-07-04T15:18:19Z` .. `2026-07-24T18:25:38Z`, and **nothing since 07-24**, so this is historical residue from a durability bug that has stopped, not an ongoing fault. The chain itself is intact — the tip chains correctly (`2069.prev_hash == 2068.row_hash`) and the scan reports no `sequence_gap`, `predecessor_mismatch`, or `row_hash_mismatch` anywhere.

## The gap

The content waiver deliberately never applies to chain findings — correctly, since a waived row that is mis-chained must still fail. But that means an `unchained_record` has no declaration path at all. The result is a public integrity endpoint stuck at `ok:false` forever, for a known, bounded, already-fixed historical cause. That is the worst possible state for an integrity signal: permanently red is indistinguishable from newly-broken, so a REAL failed append tomorrow is invisible.

## Proposed fix

A declared, bounded, disclosed exclusion for unchained records, mirroring `LOOPOVER_LEDGER_CONTENT_WAIVER` exactly in spirit:

- Bounded by `created_at`, not seq — an orphan has no ledger row and therefore no sequence number.
- Both bounds and a reason mandatory; malformed fails CLOSED and is surfaced by preflight, same as #9850.
- Waived orphans counted and published separately (`waivedUnchainedRecords` + `unchainedWaiver`), never folded into the real count, and always disclosed even on a clean chain.
- Genuinely structural findings (`sequence_gap` / `predecessor_mismatch` / `row_hash_mismatch` / `short_tail`) stay unwaivable. A truncated tail is exactly the attack this check defends, so only the INTERIOR `unchained_record` kind is declarable.

That restores the property that matters: red means something new is wrong.

JSONbored:
Closes #9933

## The problem, measured on the production ORB (orb-v3.7.0-beta.1, 2069 rows)

`/v1/public/decision-ledger/verify` reports `ok:false` and **cannot be made honest-green**, because one of its two findings has no declaration mechanism.

**83 content mismatches — already declarable.** All in `seq 5..257`, all in `schemaVersion` 2 and 3; every one of the 1,569 v5 rows verifies clean. I recomputed the digests offline with the repo’s own `canonicalJson`/`contentDigest` to confirm the count and range exactly, then brute-forced every subset of null-valued keys per row (the obvious "stored as null, hashed as undefined" hypothesis) — **no subset reproduces the committed digest for any of the 83**. Genuinely unrecoverable preimages, which `LOOPOVER_LEDGER_CONTENT_WAIVER` (#9850) already covers.

**231 unchained records — not declarable.** `decision_records` rows no ledger row ever vouched for, between `2026-07-04` and `2026-07-24`, **none since**. Historical residue from a durability bug that has stopped, against a structurally perfect chain (the tip chains correctly; no `sequence_gap`, `predecessor_mismatch` or `row_hash_mismatch` anywhere).

The content waiver deliberately never applies to chain findings — correct, since a waived row that is mis-chained must still fail. But that leaves the endpoint permanently red for a known, bounded, already-fixed cause, and **permanently red is indistinguishable from newly-broken**, so a real failed append tomorrow would be invisible.

## The fix

`LOOPOVER_LEDGER_UNCHAINED_WAIVER`, format `<fromIso>..<toIso>|<maxRecords>|<reason>`, mirroring #9850’s four properties. Two deliberate differences, both argued in the parser’s doc comment:

- **Time-bounded, not seq-bounded.** An orphan is *defined* by having no ledger row, so it has no sequence number to name. #9850’s rule against time bounds targets boundaries that **move** — a relative window swallows new damage as the clock advances. Both bounds here are absolute instants, so the interval is exactly as fixed as a seq range.
- **A declared `maxRecords`**, which closes what a seq range gets for free: a time interval could come to contain more orphans than were counted. Exceeding the declared count makes the **whole** waiver stop applying, so a new (or backdated) failed append cannot hide behind an old declaration.

Pipe-delimited because ISO-8601 instants contain colons.

## A truncated tail stays unwaivable

Writing that invariant test caught a real flaw in my own first implementation: it excluded the waiver window in SQL *before* distinguishing an interior orphan from a `short_tail`, which silently waived the truncated-tail signature — precisely the attack the doc promised was never waivable. Every waiver predicate now carries the interior test, so the count and the break search agree on which rows are covered.

## Tests

`parseLedgerUnchainedWaiver`: bounded parse, pipes preserved inside a reason, mandatory reason, both bounds required, descending range rejected, non-positive/fractional/`12abc` counts rejected, fail-closed on malformed.

Applied behaviour: a declared orphan verifies clean and is **disclosed** (`waivedUnchainedRecords` + `unchainedWaiver`, never folded into silence); one more orphan than declared makes the whole waiver lapse; an orphan outside the window is still reported; a `short_tail` inside the window is still reported; no waiver is byte-identical to before; a malformed waiver waives nothing. Plus preflight coverage for the set-but-unparseable case.

Changed-line coverage measured at **zero uncovered statements and zero uncovered branches**.


JSONbored:
@loopover review

Issue #9939, opened 2026-07-30
The manual-review label is a one-way latch: nothing lifts it when the hold that applied it clears


JSONbored:
Observed live on #9935 (orb-v3.7.0-beta.1).

The PR was `MERGEABLE` / `mergeStateStatus: CLEAN`, CI fully green, and a **fresh escalated review on the current head returned zero findings**. It still would not merge:

```
agent.action.merge   | denied | manual-review label "manual-review" is present on the live PR — merge not executed
agent.action.approve | denied | manual-review label "manual-review" is present on the live PR — approve not executed
```

The label had been applied several passes earlier, when an `ai_consensus_defect` fired against a **stale head** (`fb4ffe52`, pre-rebase). The finding was legitimate at the time — the branch carried generated-schema drift picked up from a `main` that was itself stale. A rebase removed it, `api-schemas.ts` went to a zero diff against `main`, and the re-review came back clean. None of that lifted the label, so the PR sat blocked until a human removed it — in the exact autonomy mode that is supposed to have no human in it.

## Why it is a latch

`agent-actions.ts` clears stale disposition labels (`readyToMerge` / `migrationCollision` / `changesRequested`) once the disposition resolves, but deliberately excludes this one:

> Intentionally do NOT remove `manualReview` here: that same label is also the live maintainer safety hold/freeze, and the planner has no provenance bit proving it was only a stale bot disposition rather than a human-applied hold. Only a maintainer removing the label should lift it.

That reasoning is **correct** and must not be reverted — the label is overloaded. A human applying `manual-review` to freeze a PR must have that respected, and auto-removal without provenance would silently override a deliberate human hold, which is far worse than the current stall.

The gap is the missing provenance, not the caution.

## Fix

Record who applied it, then let only the bot lift what the bot applied:

- Persist the provenance when the PLANNER adds `manualReview` — head SHA plus the hold reason, alongside the existing per-PR markers (`visualCaptureRetryPendingSha` and friends).
- On a later pass: if the label is live **and** provenance says the bot applied it **and** the disposition no longer holds it for manual review, remove it and say why in the audit trail.
- No provenance ⇒ never auto-removed. A human-applied label, or one predating this change, keeps exactly today’s behaviour. That is the property that makes this safe.
- The removal must be scoped to the reason that applied it, so a label applied for reason A is not lifted by reason B clearing.

[...]

PR #9943, opened 2026-07-30
fix(gate): do not lift the manual-review hold while another hold is live

JSONbored:
Follow-up to #9942, which landed the #9939 fix while this was in flight. **This PR no longer duplicates it** — it was reset onto `main` and now fixes one gap in the released version.

## The gap

#9942 guards the release on `manualHoldReason === null`, and its comment states the case cannot arise:

> Guarded on `manualHoldReason === null`, i.e. NO reason wants a hold this pass — which is why this cannot lift a label applied for reason A just because reason B cleared.
>
> No "is an add already planned this pass?" guard: every site that ADDS this label requires a live hold … and all of those set `manualHoldReason` … A guard here would be an arm no input can reach.

`manualHoldReason` covers only the guardrail hit, unverified CI, an `action_required` conclusion, and a not-review-good verdict. **Every would-MERGE hold fires on `reviewGood` with a `success` conclusion — precisely when `manualHoldReason` is null.**

## Verified against the merged planner, before changing anything

Driving `planAgentMaintenanceActions` from `main` (35fceede) with the label present, provenance set, and one hold live:

| live hold | release planned on main? |
| --- | --- |
| `migrationCollisionHold` | **YES** |
| `unlinkedIssueMatchHold` | **YES** |
| `priorityEligibilityHold` | **YES** |
| `unlinkedIssueMatchClose` | **YES** |
| unstable merge state | **YES** |

It is removed once and not re-added on the next pass (with `review_state_label` acting, the re-add sites are skipped in favour of the sibling disposition label). So this is not a flap — it is a one-way strip.

**Why that matters:** the manual-review label is the exact thing the executor checks to deny `merge` and `approve`. Removing it while the hold still stands removes the enforcement for a hold nothing has resolved. Whether a merge actually follows depends on the rest of the plan — I am not claiming these PRs merged — but the block that was supposed to stop them is gone.

## The fix

One named predicate listing every reason that would ADD the label, so a hold added later cannot silently fail to suppress the release:

```ts
const noManualReviewHoldWanted =
  manualHoldReason === null &&
  input.migrationCollisionHold === undefined &&
  input.unlinkedIssueMatchHold === undefined &&
  input.priorityEligibilityHold === undefined &&
  input.unlinkedIssueMatchClose === undefined &&
  !mergeableStateUnstable &&
  !heldForManualReview &&
  !mergeTerminallyBlocked;
```

Everything else in #9942 is untouched — the provenance column, the SHA-agnostic release, the human-applied-label refusal.

## Validation

- 7 new tests: one per live hold, plus **the counterweight** — with nothing holding, the label is still released, so the one-way latch #9942 fixed does not come back. Plus the no-provenance refusal.
- Mutation-verified: reverting the predicate to `manualHoldReason === null` fails **exactly those 5** and leaves the counterweight passing, so they are pinning the real behaviour and the guard is not merely over-broad.
- `test/unit/agent-actions.test.ts`: 343 passing.
- Full checker sweep + suite results appended below once they finish.

### DIFF
--- a/src/settings/agent-actions.ts
+++ b/src/settings/agent-actions.ts
@@ -1286,12 +1286,30 @@ export function planAgentMaintenanceActions(input: AgentActionPlanInput): Planne
   //
   // `manualReviewLabelAppliedSha` is the missing provenance. Non-null means the PLANNER applied it, so the
   // planner may take it back; null means a human applied it (or it predates the column) and it is left
-  // strictly alone. Guarded on `manualHoldReason === null`, i.e. NO reason wants a hold this pass -- which is
-  // why this cannot lift a label applied for reason A just because reason B cleared. Deliberately not scoped
-  // to the recorded head: a rebase that resolves the cause is the single most common way a hold goes stale,
-  // and refusing to lift it there would leave the exact #9935 case unfixed.
-  if (
+  // strictly alone. Deliberately not scoped to the recorded head: a rebase that resolves the cause is the
+  // single most common way a hold goes stale, and refusing to lift it there would leave the #9935 case unfixed.
+  //
+  // `manualHoldReason === null` alone is NOT "nothing wants a hold". That ternary covers the guardrail hit,
+  // unverified CI, an action_required conclusion and a not-review-good verdict -- but every one of the
+  // would-MERGE holds in sections 1c-1g below fires on `reviewGood` with a SUCCESS conclusion, which is
+  // precisely when manualHoldReason is null. Releasing on it alone therefore stripped the label while a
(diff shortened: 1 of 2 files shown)


JSONbored:
Validation complete on the final tree:

- Full `test/unit` + `test/integration`: **25,741 passed, 0 failures.**
- Full sweep of every `check-*` script: the only two failures are the pre-existing red-main ones fixed in #9944 (`dead-exports:check`, `replay-runner-manifest:check`), both reproduced on a clean checkout of `origin/main` and neither caused by this branch.

PR #9956, opened 2026-07-30
feat(review): regenerate a public-safe summary when the narrative is withheld

### LINKED ISSUE #9809: feat(review): regenerate a public-safe summary when the narrative is withheld — a real summary on every review
## Problem

When every sentence of a review's narrative trips the public-safety sanitizer (routine for PRs touching this project's own scoring/gate code — an honest sentence says "score"/"ranking"), the published panel has no real summary. #9806 fixed the misattribution (it no longer claims a provider failure, and withheld-blockers vs clean-result get distinct honest sentences), but the reader still gets a fixed placeholder rather than a genuine summary.

Maintainer requirement: **a real review summary every time** — including clean, merge-ready PRs, where the summary is the merge-readiness indicator. Measured volume: 35 PRs held `ai_review_inconclusive` in 14 days; the withheld-narrative subset of those is the target.

## Design

When `composeAdvisoryNotes` reaches the withheld-narrative branch (parse succeeded, nothing survived sanitization), make **one additional small LLM call**: rewrite the already-produced narrative for a public audience with a prompt that names the forbidden vocabulary classes, then run the rewrite through the same `toPublicSafeBySentence` sanitizer. Publish it only if it survives; otherwise keep #9806's honest fixed sentence as the floor.

Properties:
- **Never-echo preserved**: the regenerated text passes the identical sanitizer gate before publication; the fixed sentence remains the fail-safe.
- **Bounded cost**: fires only on the withheld branch (~2-3/day observed), one small completion each.
- **No new config**: pure behavior improvement; self-hosters get it with the image.

## Scope

1. `regeneratePublicSafeSummary(review, sanitizer)` in `src/services/ai-review.ts` — prompt, single attempt, sanitize, fall back.
2. Wire into the withheld branch from #9806; keep the fixed-sentence fallback and the withheld-blockers wording rule (a withheld blocker must never read as clean).
3. Telemetry: `ai_review_summary_regenerated` (success/fallback) so the regeneration rate is observable.
4. Tests: regeneration survives the sanitizer → published; regeneration itself trips the sanitizer → fixed sentence; blockers-withheld case never reads clean.

JSONbored:
Closes #9809.

When every sentence of a review's narrative trips the public-safety sanitizer — routine for a PR touching this project's own scoring/gate code, where an honest sentence says "score" or "ranking" — the reader got a fixed placeholder rather than a real summary. #9806 made that placeholder *honest*; this makes it a *summary*.

On the withheld branch only, the narrative is rewritten for a public audience in one small completion, then held to the **identical** sanitizer.

## The prompt is not the boundary

The rewrite is not trusted because it was asked nicely. It goes through the same `toPublicSafeBySentence` gate, and anything that does not survive is discarded for the fixed sentence. So this can only improve the text a reader sees — it **cannot widen what is publishable**. A prompt-injected or careless rewrite fails closed, as does a provider error and the model's own `CANNOT_SUMMARIZE` opt-out.

Mutation-verified: replacing `toPublicSafeBySentence(trimmed, options)` with `trimmed` fails the suite.

## One conflict in the issue, resolved toward safety

The issue asks for a real summary **every time**, *and* that a withheld blocker **never read as clean**. For a withheld blocker those pull against each other — the rewrite would be describing findings it cannot name.

**Blockers keep their fixed sentence**, structurally rather than by trusting the prompt. Prose reading clean over a blocker the model actually raised is the one outcome that could green-light a PR. Also mutation-verified: letting the regenerated text through on that branch fails the suite.

If you'd rather have B (regenerate for blockers too, with the prompt required to state that findings exist) or C (regenerate *and* append the fixed blocker sentence), both are small follow-ups — but only this version makes the never-green-light property independent of prompt compliance.

## Where the IO lives

The provider call is at the **single caller**, not inside `composeAdvisoryNotes`. The issue's design put it in the composer, but that function is pure and synchronous; making it async would push `await` through its call site and cost the testability that makes the sanitizer's behaviour verifiable at all.

`narrativeWasWithheld()` lets the caller decide whether to spend the call *before* composing, and computes it the same way the composer does — reproducing the sanitizer's decision by other means would let the two drift, and a mis-fire would spend a completion on **every** review rather than the ~2-3/day this targets.

## Telemetry

`loopover_ai_review_summary_regenerated_total{outcome=published|fallback}`, registered in `DEFAULT_METRIC_META` like its siblings so it renders with HELP/TYPE. **Both** outcomes are counted — an all-fallback result must not look identical to the feature never firing.

## Validation

- 7 new tests; both safety properties mutation-verified (the blocker guard and the sanitizer-is-authority rule each fail the suite when inverted).
- Full sweep of **all 36** `check-*` scripts: none failing. `selfhost:validate-observability` green.
- Full suite: 25,771 passing with one failure I caused and fixed — my provider-path cases stubbed global `fetch` and never restored it, which leaked into `selfhost-metrics.test.ts` in a full run while passing in isolation. Second commit unstubs; re-run in flight and I'll post the result.

### DIFF
--- a/src/selfhost/metrics.ts
+++ b/src/selfhost/metrics.ts
@@ -133,6 +133,7 @@ export const DEFAULT_METRIC_META: readonly (readonly [string, MetricMeta])[] = [
   ["loopover_ai_review_non_cacheable_total", { help: "AI reviews skipped by cacheability rules.", type: "counter" }],
   ["loopover_ai_review_force_bypass_total", { help: "AI review cache force-bypass events.", type: "counter" }],
   ["loopover_ai_review_inconclusive_total", { help: "AI review inconclusive outcomes.", type: "counter" }],
+  ["loopover_ai_review_summary_regenerated_total", { help: "Public-safe summary regenerations on the withheld-narrative branch, by outcome (#9809): `published` when the rewrite survived the sanitizer, `fallback` when the fixed sentence was kept. Both are counted -- an all-fallback result must not look the same as the feature never firing.", type: "counter" }],
   ["loopover_ai_review_unpublishable_blocker_total", { help: "AI reviews where a reviewer named a real blocker whose title could not be published, so the verdict held instead of passing (#9460).", type: "counter" }],
   ["loopover_ai_review_onmerge_clamped_total", { help: "AI review on-merge mode clamp events.", type: "counter" }],
   ["loopover_ai_review_model_fallback_total", { help: "AI review model fallback attempts by primary and fallback model.", type: "counter" }],
--- a/src/services/ai-review.ts
+++ b/src/services/ai-review.ts
@@ -2120,6 +2120,56 @@ export async function callAiProvider(
 
 /** Run the maintainer's BYOK frontier model for the advisory write-up. Never throws; the review is null on
  *  any error and `failure` names the reason (timeout/http_error/exception) for the audit trail. */
+/** Token ceiling for the rewrite. A summary paragraph, not a second review -- this call exists to rephrase
+ *  text the model already produced, so a small budget is both sufficient and the cost control. */
+const REGENERATED_SUMMARY_MAX_TOKENS = 400;
+
+/**
+ * Rewrite an already-produced review narrative for a PUBLIC audience, then hold it to the same sanitizer
+ * (#9809).
+ *
+ * Fires only when `narrativeWasWithheld` is true -- every sentence of the original tripped the public-safety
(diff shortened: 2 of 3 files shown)


JSONbored:
Full suite re-run after the fetch-unstub commit: **25,772 passing, 0 failures** (1,326 files). The `selfhost-metrics.test.ts` cross-file failure is gone — it was my stub leaking, as described above.

Combined with the 36/36 checker sweep, this branch is green end to end.

PR #9970, opened 2026-07-30
fix(selfhost): count the Orb's own decision ledger in published totals

### LINKED ISSUE #9963: selfhost(stats): an Orb reports totals.handled = 0 while its ledger holds 2,123 verdicts
Found by turning on `LOOPOVER_PUBLIC_STATS` on `edge-nl-01` to test #9940's option C. Caught immediately by the public verifier (#9723) and reverted within minutes — **the flag is back off and nothing contradictory was left published.**

## The contradiction

```
$ npx -p @loopover/mcp loopover-verify --base-url https://shots.loopover.ai
FAIL  Published headline stats agree with the ledger-derived parity rollups
      parity rollups report 2123 verdict(s) over their window, exceeding the all-time handled count of 0
```

On the same deployment, in the same payload:

- `reviewParity.verdicts` = **2123** — read from `decision_records`, matching the ledger's ~2,200 rows
- `totals.handled` = **0**

An Orb that has decided 2,123 pull requests reporting zero handled is not a rounding difference; one of the two reads is looking at the wrong place.

## Why it matters beyond cosmetics

This is the deployment that actually performs reviews. If self-hosters enable public stats, they publish a headline of `0` next to a parity block describing thousands of verdicts — and `accuracyPct` derives from the same `handled`/auto-action population, so it is suspect too. That is worse than publishing nothing, which is why the flag was reverted rather than left on.

## Likely cause

`totals.handled` is built from `audit_events` (`github_app.pr_public_surface_published`, `agent.action.*`) joined to the `pull_requests` cache and folded with `orb_pr_outcomes`, while `reviewParity` reads `decision_records` directly. On this Orb the **primary store is Postgres**, not Cloudflare D1 — so the audit/outcome side of the stats query is very likely reading a store that is empty here, while the `decision_records` side reads the populated one.

If that is right, every `totals.*`-derived figure (`handled`, `reviewed`, `merged`, `closed`, `accuracyPct`, `minutesSaved`) is wrong in self-host mode, and only the `reviewParity` block is correct.

## Acceptance

- On a self-hosted Orb with a populated ledger, `totals.handled` reflects reality rather than 0.
- `loopover-verify --base-url <orb>` reports the stats-parity claim as PASS.
- A regression test pins that `totals.handled` and `reviewParity.verdicts` cannot disagree by orders of magnitude on the same payload — the invariant the verifier checks.
- Until then, `LOOPOVER_PUBLIC_STATS` on an Orb stays off; see #9940 for why the two surfaces are separate.

JSONbored:
## The contradiction

An Orb published, in one payload:

- `reviewParity.verdicts` = **2123** (from `decision_records`)
- `totals.handled` = **0**

## Root cause — not the one the issue guessed

The issue supposed `totals.*` reads D1 while the Orb's primary store is Postgres. It does not: both reads hit the same store on that box. The real split is **which tables**, and the answer is that neither of `totals`' two sources exists on a self-hosted Orb.

| Source | What it reads | On the Orb |
| --- | --- | --- |
| own-ledger | `audit_events` published-surface, scoped to `LOOPOVER_PUBLIC_STATS_REPOS` | **skipped** — the allowlist is unset, and `getPublicStats` skips those queries outright when it is empty |
| fleet fold | `orb_pr_outcomes` for **other** registered installations | **empty** — an Orb is not the aggregator |
| the actual ledger | `decision_records` | 2,123 rows, **not a source at all** |

`LOOPOVER_PUBLIC_STATS_REPOS` is a frozen snapshot of the repos the old central App used to process. It is a hosted-Worker concept; no self-hoster has a reason to fill it in. Confirmed directly against the deployment — the Orb sets `LOOPOVER_REVIEW_REPOS` (three repos) and does not set `LOOPOVER_PUBLIC_STATS_REPOS` at all. So every `totals.*` figure was structurally zero on the deployment that does the work, while `reviewParity` — which has no allowlist — reported reality.

## The fix

`decision_records` becomes a third disposition source, emitted in the same shape the published-surface query produces so `totals`, `byProject` and the ordering all fold it identically. It is deliberately evaluated **outside** the allowlist branch: an empty allowlist is precisely the case it exists to cover, so gating it there would reproduce the bug.

The three sources stay **disjoint by construction** rather than reconciled afterwards — the discipline the file header already states for the first two. Two anti-joins do it, each excluding exactly what another source counts:

- the published-surface set, **but only where that repo is allowlisted**, because that is the exact condition under which the first query counts a pair. Excluding an un-allowlisted published pair would drop it from every source at once.
- the registered-install outcomes the fleet fold counts, matched on the same `(repo, pr_number)` key its own anti-join uses, and on the same `registered = 1` population.

Verdicts are counted per **distinct** PR, so a re-evaluated PR (which appends rows) does not inflate the headline.

On the hosted Worker `decision_records` is empty by design (review execution retired), so **its published figures do not move at all** — pinned by a test.

### Deliberately not changed

`accuracyPct` still pairs `windowedMerged`/`windowedClosed` with `totals.reversed`. Those share the numerator's bounds by design (own-ledger, inside the retention window); widening the denominator with ledger rows whose reversals are not in that population is exactly the #7449 error the file warns against. On the Orb it therefore reports `null` — honest — and the UI already prefers `fleetAccuracy`.

## Tests

Against a **real migrated D1**, not a SQL-shape stub: the bug was a real query returning nothing for a real reason, and a stub asks the wrong question and answers it happily.

The parity invariant is asserted by calling the verifier's own `checkStatsParity` — the function that printed the original FAIL — rather than restating it, so it cannot drift from the tool that decides whether production is publishing a contradiction.

Every guard is **mutation-tested**; each fails, baseline restored green:

| Mutation | Result |
| --- | --- |
| Ledger source gated behind the allowlist (the original bug) | 4 failed |
| Published-surface anti-join dropped | 1 failed |
| Fleet anti-join dropped | 1 failed |
| Anti-join matches unregistered installs too | 1 failed |
| Counts ledger rows instead of distinct PRs | 1 failed |

The fleet anti-join had **no** covering test on the first pass; the two `orb_pr_outcomes` cases were added specifically because that mutation survived.

## Acceptance criteria

- [x] On a self-hosted Orb with a populated ledger, `totals.handled` reflects reality rather than 0.
- [x] `loopover-verify` reports the stats-parity claim as PASS (pinned via the verifier's own checker).
- [x] A regression test pins that `totals.handled` and `reviewParity.verdicts` cannot disagree — both directions, including the failing one.

## Verification

- `npx tsc --noEmit` clean; static checker sweep green
- `npx vitest run test/unit` — **1299 files, 25323 tests passed**

`LOOPOVER_PUBLIC_STATS` remains **off** on the Orb; enabling it is a separate operator decision once this ships.

Closes #9963

### DIFF
--- a/src/review/public-stats.ts
+++ b/src/review/public-stats.ts
@@ -320,6 +320,62 @@ export const PUBLISHED_PR_KEYS = `
   WHERE event_type = 'github_app.pr_public_surface_published' AND instr(target_key, '#') > 0
                   AND length(target_key) - length(replace(target_key, '#', '')) = 1`;
 
+/**
+ * #9963: the deployment's OWN decision ledger as a disposition source, in the same shape the published-surface
+ * query above produces, so everything downstream (totals, byProject, the sort) folds it identically.
+ *
+ * Why this exists. `totals` had exactly two sources -- the allowlisted `audit_events` published-surface snapshot
+ * and the registered-installs fleet fold -- and BOTH are hosted-Worker concepts. On a self-hosted Orb
+ * `LOOPOVER_PUBLIC_STATS_REPOS` is unset (it is a frozen snapshot of the repos the old central App used to
+ * process, not a list any self-hoster has a reason to fill in), so the own-ledger queries are skipped entirely;
+ * and `orb_pr_outcomes` holds what OTHER registered installations reported, which on an Orb is nobody. Both
+ * sources are therefore structurally empty there, while `decision_records` -- the anchored ledger this
+ * deployment actually writes a row to per verdict -- held 2,123 rows. The result was a published
+ * `totals.handled: 0` beside a `reviewParity.verdicts: 2123` computed from that same ledger: not a rounding
+ * difference, an entire population missing. Every `totals.*` figure inherited it.
+ *
+ * DISJOINT BY CONSTRUCTION, so the three sources can be added rather than reconciled -- the same discipline the
+ * file header states for the first two. Both anti-joins below are exclusions of pairs another source already
+ * counts:
+ *   • the published-surface set, but ONLY where that repo is allowlisted, because that is the exact condition
+ *     under which the first query counts a pair. Excluding an un-allowlisted published pair would drop it from
+ *     every source at once;
+ *   • the registered-install outcomes the fleet fold counts (`getOrbGlobalStats`'s own population, matched on
+ *     the same (repo, pr_number) key its anti-join uses).
+ *
+ * On the hosted Worker `decision_records` is empty by design (review execution is retired there), so this adds
+ * nothing and that deployment's numbers do not move at all.
+ */
+function ledgerDispositionsSql(allowlistPlaceholders: string): string {
+  // The allowlist arm is omitted entirely when there is no allowlist: `IN ()` is a syntax error in both
+  // dialects, and with no allowlist the published-surface query contributes nothing to exclude anyway.
+  const publishedExclusion =
+    allowlistPlaceholders === ""
+      ? ""
+      : `AND NOT EXISTS (
+             SELECT 1 FROM audit_events ae
+              WHERE ae.event_type = 'github_app.pr_public_surface_published'
+                AND ae.target_key = dr.repo || '#' || dr.number
+                AND LOWER(dr.repo) IN (${allowlistPlaceholders})
+           )`;
+  // `||` for concatenation and NOT EXISTS both mean the same thing in SQLite and Postgres and need no
+  // translation -- the same basis on which getOrbGlobalStats already builds its target_key anti-join.
+  return `SELECT dr.repo AS project,
+                 COUNT(*) AS reviewed,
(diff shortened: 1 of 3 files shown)


JSONbored:
Superseded by #9967, which landed the same root-cause analysis (the `LOOPOVER_PUBLIC_STATS_REPOS` gate, not a D1/Postgres split) and a better fix.

Two things #9967 gets right that this PR did not:

- **`byProject` stays gated on the allowlist.** This PR populated it from the ledger, which would have published repo names that were never opted in. #9967's framing is the correct one — the privacy control is about *naming repos*, and `totals` carries no repo identity while `byProject` does.
- **Ledger-as-fallback rather than a third summed source.** Using the ledger only when nothing is allowlisted removes the double-counting question by construction, instead of answering it with two anti-joins that then have to be kept correct.

Closing rather than rebasing: the remaining delta is not the fix, it is the cross-surface regression test #9963's acceptance asked for, which is tracked separately and lands on its own.

Issue #9962, opened 2026-07-30
fairness: a corpusChecksum is published for a rule whose corpus is not downloadable


JSONbored:
Split out of #9940, which turned out to be a separate (topological) problem.

## The defect

`api.loopover.ai` publishes an eval-score record for `ai_consensus_defect` carrying a `commitments.corpusChecksum`, but `/v1/public/eval-corpus?rule_id=ai_consensus_defect` serves nothing usable for that rule. The commitment therefore points at bytes no reader can obtain — which is exactly what a commitment must never do.

Reproducible with the public verifier:

```
$ npx -p @loopover/mcp loopover-verify
SKIP  Every corpusChecksum matches a downloadable corpus, and no scored record commits to an empty one
      1 commitment(s) published, but no corresponding corpus is downloadable -- nothing could be rehashed
```

## Why this should not be possible

`buildPublicCorpusCommitments` (`src/review/public-eval-corpus.ts`) already implements #9805's rule and looks correct on inspection — it omits a rule whose corpus is empty or truncated:

```ts
const corpus = await loadPublicEvalCorpus(env, ruleId, nowMs);
if (corpus.caseCount === 0 || corpus.truncated) continue;
commitments.set(ruleId, corpus.checksum);
```

So the commitment path and the serving route **disagree about the same rule**. Candidates:

1. The two read different windows or different `nowMs`, so the builder sees cases the route no longer does.
2. The route applies a filter the builder does not (or vice versa) — the provenance exclusion and the rule-history read bound are both places they could diverge.
3. The route is failing for a reason that degrades to empty (`loadPublicEvalCorpus` is fail-safe by design, so a read error is indistinguishable from a genuinely empty corpus at the call site).

Candidate 3 is the most likely and the most concerning: a failed read silently produces the same result as "no cases", which means a transient D1 problem could publish a commitment while the corpus route serves nothing — the exact state observed.

## Acceptance

- A `corpusChecksum` is published only when the corpus route can actually serve the bytes it commits to, for the same rule and window.
- The two paths derive that decision from ONE computation rather than two that can disagree.
- If a read failure is why a corpus is empty, that is distinguishable from a genuinely empty corpus at the point the commitment is decided.
- `loopover-verify` reports the corpus claim as PASS against production.

Issue #9940, opened 2026-07-30
fairness: the anchored ledger lives on the ORB, which is not publicly reachable — the public API serves an empty one


JSONbored:
## What

Running the new public verifier (#9723) against production returns two SKIPs, not passes:

```
PASS  Every eval-score record's recordDigest recomputes from its own contents   1 record(s) recomputed exactly
SKIP  Every corpusChecksum matches a downloadable corpus ...                    1 commitment(s) published, but no corresponding corpus is downloadable
SKIP  The current signed ledger checkpoint verifies offline ...                 no signed checkpoint published (anchor signing not configured, or the ledger is empty)
PASS  Published headline stats agree with the ledger-derived parity rollups     all-time handled=12882, windowed parity verdicts=0
```

Both SKIPs are facts about the deployment, not verifier limitations.

## 1. No signed checkpoint is published

`/v1/public/decision-ledger/anchor-payload` 404s. Per its handler that means either
`LOOPOVER_LEDGER_ANCHOR_KEYS` / `LOOPOVER_LEDGER_ANCHOR_PRIVATE_KEY` are unset, or the ledger tip is 0.

Consequence: the entire anchoring story (#9267 epic) is currently unverifiable by an outsider. The
machinery ships and is tested, but nothing in production is actually signed, so "the ledger's tip really
was X at time T" cannot be checked by anyone.

## 2. The one committed corpus is not downloadable

The single published eval-score record commits a `corpusChecksum` for `ai_consensus_defect`, but
`/v1/public/eval-corpus?rule_id=ai_consensus_defect` returns nothing usable — so the commitment covers
bytes no reader can obtain. Per #9805 a commitment is supposed to be OMITTED when the corpus is empty or
truncated, rather than published pointing at nothing; that omission does not appear to be happening here,
or the corpus read is failing separately.

## Acceptance

- `npx -p @loopover/mcp loopover-verify` against production reports PASS (not SKIP) for the anchor
  checkpoint and corpus commitment claims.
- If a corpus genuinely cannot be published for a rule, that rule's record publishes no `corpusChecksum`
  at all, so the verifier reports it as absent rather than as an uncheckable commitment.

JSONbored:
## Correcting my own diagnosis — it is not a missing secret

I filed this saying anchoring was "not configured". **That is wrong.** Anchoring is configured and fully working — just not on the deployment the public verifier points at.

Verified directly on both:

[...]

PR #9973, opened 2026-07-30
ci(verifiability): verify our own public claims nightly, as an outsider

### LINKED ISSUE #9724: ci(verifiability): nightly anonymous walkthrough run — fail loudly on any public-path regression
## Problem

Public verification endpoints and docs can regress silently (host renames, auth changes, empty exports). The published claims should be enforced by CI, not by outside readers.

## Scope

- Scheduled workflow (nightly) that runs the one-command verifier (sibling sub-issue) against production, with no repo credentials — an anonymous outsider's view
- On failure: opens or updates a single tracking issue with the failing step and response excerpts; notifies via the existing alerting path (PagerDuty)
- On recovery: closes the tracking issue automatically
- Badge or status line on the docs walkthrough page reflecting the last green run timestamp

## Acceptance

A deliberately broken staging config (e.g., auth on the corpus export) turns the job red within one cycle and produces an actionable issue; restoring it goes green and closes it.

JSONbored:
## What this is

A scheduled job that is the stranger the docs invite. It runs the verifier **published on npm** against **production**, from a runner holding **no credentials of any kind**, and turns a regression in the public verification path into our problem within a cycle instead of an outside reader's discovery.

Both design choices are load-bearing:

- **The published tool, not this tree.** `npx -p @loopover/mcp` resolves the npm release rather than building from the checkout. What is asserted is "an outsider succeeds today", and an outsider has the published CLI — so a fix that is merged but unpublished must *not* turn this green. #9962 was exactly that: the shipped verifier asked `/v1/public/eval-corpus?rule_id=` while the route reads `ruleId`, every corpus fetch 400'd, and the tool reported production as unverifiable while production was fine. Both sides were individually tested and individually correct. Only running the real pair catches it.
- **No token on the verify step.** Job permission is `contents: read`; `issues: write` and the token exist only on the step that files the tracking issue, after verification finishes. A verifier running with repo credentials is not reproducing an outsider's view and could pass on access a stranger does not have.

Both surfaces are checked (#9940): `api.loopover.ai` for aggregates and eval-score records, `shots.loopover.ai` for the anchored ledger. Checking one leaves the other free to regress silently.

## Self-verification found a false-green

Per the agreed approach, I pointed the real verifier at a deliberately wrong base URL rather than staging a broken config. It found a bug in my first draft of the decision logic:

```
$ loopover-verify --base-url https://api.loopover.ai/definitely-not-the-api --json
skip  record-digests      /v1/public/eval-scores unavailable (HTTP 404)
skip  corpus-commitments  /v1/public/eval-scores unavailable (HTTP 404)
skip  anchor-checkpoint   no signed checkpoint published, and the ledger size is unknown
skip  stats-parity        /v1/public/stats unavailable (HTTP 404)
exit=0
```

Four legitimate skips and a clean exit. Judged claim-by-claim that looks entirely healthy — **the monitor would have reported green against a completely dead endpoint**, which is the precise failure this job exists to prevent.

The verifier is right to call them skips: from inside a single claim, "this surface is disabled" and "this host is wrong" are indistinguishable. The *monitor* has the context the claim lacks — production is supposed to be serving these — so **a run that positively verified nothing is now red**.

The floor is **one** passing claim, not a proportion, because the two surfaces are legitimately lopsided. Measured against production right now:

| Surface | pass | skip |
| --- | --- | --- |
| `api.loopover.ai` | record-digests, corpus-commitments, stats-parity | anchor-checkpoint (ledger empty **by design**, #9940) |
| `shots.loopover.ai` | anchor-checkpoint | the three stats/eval claims (publishes no stats) |

"Most claims pass" would fail the Orb every healthy night.

### End-to-end result

| Input | Verdict |
| --- | --- |
| Both real production surfaces | `VERIFY_OK=true` — *4 passed, 0 failed, 4 skipped* |
| Wrong base URL + healthy Orb | `VERIFY_OK=false` — names **which** surface is dead; the healthy one still passes |

Restoring the URL goes green, which is what closes the tracking issue — the recovery half of the acceptance criterion.

## Other false-greens rejected

- a run that produced **no readable report** (CLI crashed, malformed JSON, npx could not resolve the package): nothing failed because nothing *ran*;
- a **non-zero exit with no failing claim** — two contradictory signals, and the comfortable one is wrong;
- an **unreadable capture file**, and a non-numeric exit code, which is treated as failure rather than an optimistic zero.

Skips alone never file an issue — the API's ledger skips nightly by design, and a monitor that files an outage on day one is muted by week one — but they are always *listed*, since a new skip beside a failure is often its cause.

## Tracking issue lifecycle

One issue, found by exact title: filed on failure, **commented on** while it persists, closed automatically on recovery. Comments rather than body rewrites, because which claim failed on which night is the useful history. The filing step runs before the step that fails the run, so a red job always leaves the issue behind — a monitor that notices a problem and then drops it is worse than none.

## Why the logic is a script

`scripts/verify-public-claims-report.ts` is pure and unit-tested (**30 tests**), because the judgement of what counts as a regression must not be discovered to be wrong on the night it first fires. It also caught an output-interleaving bug: the workflow wrapping the script's stdout in a heredoc while the script separately appended its scalars buried `VERIFY_OK` inside the body block — the workflow would have read an empty verdict, which is not `"true"`, and filed an outage on a healthy night. The script now writes all outputs itself, in one ordered write, and refuses to emit a body containing the delimiter.

## Docs

A status line and badge on the walkthrough page, which is the surface the issue asks for: green means an outsider succeeded within 24 hours, red means we already know and there is an issue naming the broken claim. The badge links to every run including the failures.

## Not included — needs your call

`workflow_dispatch` takes `api-base-url`, `orb-base-url` and `dry-run` inputs, so the job can be pointed anywhere on demand.

**PagerDuty is not wired.** The issue asks to notify "via the existing alerting path (PagerDuty)", but that path is Alertmanager on the Orb — there is no PagerDuty secret in GitHub Actions, and no workflow references one. Adding it means introducing a new repository secret, which is an operator decision rather than something to slip into this PR. The tracking issue plus the red run is the notification today. Happy to add the routing step as a follow-up if you want the secret provisioned.

## Verification

- `actionlint`, `lint:composite-actions`, `typecheck`, `docs:drift-check`, `dead-source-files:check`, `dead-exports:check`, `import-specifiers:check`, `coverage-boltons:check`, `fixture-clock-races:check`, `checkers-wired:check`, `validate:no-hand-written-js`, `ui:typecheck`, `ui:test` — all green
- `npx vitest run test/unit` — **1299 files, 25347 tests passed**

Closes #9724

### DIFF
--- a/.github/workflows/verify-public-claims.yml
+++ b/.github/workflows/verify-public-claims.yml
@@ -74,21 +74,29 @@ jobs:
           # never consults git's credential config.
           persist-credentials: false
 
-      # AUDITED (#9724). A supply-chain scan flags this action for credential-file access, citing
-      # `__tests__/authutil.test.ts`. Checked at this exact SHA (v7.0.0) rather than waved through:
+      # DELIBERATELY NO actions/setup-node. This job needs a Node RUNTIME, not this repository's toolchain:
+      # it runs a published CLI through npx and one checked-in script, and builds nothing. The runner image
+      # already provides both node and npm, and three workflows here (ci.yml, mcp-release-watch.yml,
+      # package-release-watch.yml) already rely on that.
       #
-      #   • the entrypoint is `dist/setup/index.js`; `__tests__/` is never executed by the action;
-      #   • the credential writer is `auth.configAuthentication(registryUrl)`, and `src/main.ts` calls it
-      #     ONLY inside `if (registryUrl)`.
+      # Dropping it removes the last action that can write credentials. `actions/setup-node` writes an
+      # authenticated .npmrc via `auth.configAuthentication(registryUrl)` -- gated behind `if (registryUrl)`
+      # in its `src/main.ts`, so it was unreachable while that input stayed unset, but "unreachable because
+      # nobody has added one line yet" is a weaker property than "not present". For a job whose entire claim
+      # is that it holds no credentials, not present is the one worth having.
       #
-      # So the capability is real and unreachable here, because `registry-url` is not set below. That is a
-      # property of this file, not of the action -- adding `registry-url:` would make this job write an
-      # authenticated .npmrc while still claiming to hold no credentials -- so it is asserted in
-      # test/unit/verify-public-claims-workflow.test.ts rather than left to this comment.
-      - name: Setup Node
-        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
-        with:
-          node-version-file: .nvmrc
+      # The tradeoff is that the Node version is now the runner's rather than .nvmrc's, so it is ASSERTED
+      # rather than assumed. `--experimental-strip-types` needs 22.6+; if an image ever regresses below that,
+      # this fails immediately with the fix in the message instead of the nightly quietly breaking.
+      - name: Require a Node that can strip types
+        run: |
+          node -e '
+            const [major, minor] = process.versions.node.split(".").map(Number);
+            if (major > 22 || (major === 22 && minor >= 6)) process.exit(0);
+            console.error("::error::Node " + process.versions.node + " cannot run --experimental-strip-types (needs 22.6+). Re-add actions/setup-node with node-version-file: .nvmrc.");
+            process.exit(1);
+          '
+          echo "node $(node --version), npm $(npm --version)"
 
       # No `npm ci`: this step must not depend on the repository's own dependency tree, because a stranger has
       # none. `npx -p @loopover/mcp` fetches the published package exactly as the docs tell a reader to.
--- a/.github/workflows/verify-public-claims.yml
+++ b/.github/workflows/verify-public-claims.yml
@@ -1,0 +1,157 @@
+# Nightly anonymous verification (#9724, epic #9722).
+#
+# LoopOver publishes fairness claims and tells a stranger they can check them without asking permission. This
+# job IS that stranger, on a schedule. It runs the verifier PUBLISHED ON NPM against PRODUCTION with no
+# credentials of any kind, so a regression in the public verification path is caught by us within a cycle
+# instead of by an outside reader whenever they happen to look.
+#
+# THE PUBLISHED TOOL, NOT THE ONE IN THIS TREE. `npx -p @loopover/mcp` deliberately resolves the release on npm
+# rather than building from the checkout. What is being asserted is "an outsider succeeds today", and an
+# outsider has the published CLI -- so a fix that is merged but unpublished must NOT turn this green. #9962 is
+# exactly that failure: the shipped verifier asked `/v1/public/eval-corpus?rule_id=`, the route reads `ruleId`,
+# every corpus fetch 400'd, and the tool reported production as unverifiable while production was fine. Both
+# sides were individually tested and individually correct. Only running the real pair catches it.
+#
+# NO CREDENTIALS ON THE VERIFY STEP. The job-level permission is `contents: read`, and the verify step is given
+# no token in its environment at all -- `issues: write` and the token live only on the step that files the
+# tracking issue, after verification is finished. A verifier that runs with repo credentials is not reproducing
+# an outsider's view, and could pass on access a stranger does not have.
+#
+# TWO SURFACES, because they hold different things (#9940): `api.loopover.ai` serves aggregate stats and
+# eval-score records, while the anchored decision ledger lives on the Orb at `shots.loopover.ai`. Checking only
+# one leaves the other free to regress silently, so both are verified and reported independently.
+name: Verify Public Claims
+
+on:
+  schedule:
+    # 04:12 UTC. Off the hour on purpose: the top of the hour is the busiest slot on GitHub's shared scheduler
+    # and the most likely to be delayed or dropped, which for a nightly is a silently skipped check.
+    - cron: "12 4 * * *"
+  workflow_dispatch:
+    inputs:
+      api-base-url:
+        description: "Base URL for the aggregate/eval surface"
+        required: false
+        default: "https://api.loopover.ai"
+      orb-base-url:
+        description: "Base URL for the anchored decision ledger"
+        required: false
+        default: "https://shots.loopover.ai"
+      dry-run:
+        description: "Report the verdict without filing, updating or closing the tracking issue"
+        required: false
+        default: "false"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: verify-public-claims
+  cancel-in-progress: false
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      issues: write
+    env:
+      API_BASE_URL: ${{ inputs.api-base-url || 'https://api.loopover.ai' }}
+      ORB_BASE_URL: ${{ inputs.orb-base-url || 'https://shots.loopover.ai' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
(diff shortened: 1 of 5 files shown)


JSONbored:
**On the red `Superagent Supply Chain Scan`** — investigated rather than waved through.

It reports "1 actionable risk across 2 changed dependencies". This PR changes **no** dependency files (`git diff origin/main...HEAD --name-only` over `package.json`/`package-lock.json` is empty); the two "dependencies" are the actions a *new workflow* necessarily references. Both are pinned to the SHAs this repo already standardises on — `actions/checkout@3d3c42e` (40 other uses) and `actions/setup-node@8207627` (31 other uses) — so there is no drift to correct. The same check failed on #9957 and that merged, so it is non-blocking here.

The scan did surface something real, though, and it is fixed in `7970a20b8`: the first draft ran the report script via `npx --yes tsx`, pulling an **unpinned TypeScript runner** from the registry at job time. That was pure added supply-chain surface for a script that is ours and already checked out. It now runs under `node --experimental-strip-types`, the same invocation `npm run actionlint` uses, which removes the fetch entirely.

That leaves exactly **one** intentional unpinned fetch — `npx -p @loopover/mcp loopover-verify` — and pinning it would defeat the job: resolving whatever an outsider actually gets from npm today is the property being asserted. A pinned verifier would keep passing after the published one broke, which is the #9962 failure mode this job exists to catch.

JSONbored:
**Both Superagent findings triaged.**

**Security Scan (P2) — real, and fixed in `17e212c3a`.** The finding was precise: *"actions/checkout persists GitHub token to disk, contradicting the 'no credentials on verify step' design."* It was right. `actions/checkout` defaults to `persist-credentials: true`, which writes an authenticated `http.extraheader` into `.git/config` — so withholding the token from the verify step's `env:` while leaving a usable credential on disk was theatre, and my claim that the step runs with no credentials was false. Now `persist-credentials: false`. Nothing in the job needs it: the only step that talks to GitHub is the tracking-issue step, which uses `gh` with an explicit `GITHUB_TOKEN` in its own environment, and `gh` reads that variable directly rather than git's credential config.

**Supply Chain (HIGH) — not actionable.** It flags `actions/setup-node@8207627` for `threat-filesystem-read`, and the cited location is inside the action's **own test suite**:

```
setup-node-8207627.../__tests__/authutil.test.ts
```

That is a test fixture exercising npmrc handling, not behaviour this workflow invokes. The pin is the SHA **31 other workflows in this repo already use**, and this PR changes no dependency manifest — the "2 changed dependencies" are simply the two actions any new workflow must reference. There is nothing to remediate short of unpinning from the repo standard, which would be worse.

`validate-code` will stay red here until #9976 lands — that is the repo-wide `collections/*` breakage where a `node_modules` cache hit skips `postinstall`, unrelated to this PR and currently failing every PR.

JSONbored:
**Status: everything passes except `Superagent Supply Chain Scan`.** `validate-code`, `validate`, `codecov/patch`, `codecov/project`, `Workers Builds` and `Superagent Security Scan` are all green.

The **Security** finding is fixed — it was real, and it was mine: `actions/checkout` defaults to `persist-credentials: true`, writing an authenticated header into `.git/config`, which made this job's "no credentials on the verify step" claim false. Now `persist-credentials: false`.

The remaining **Supply Chain** failure is a false positive I do not think should be fixed in code:

```
HIGH: actions/setup-node@8207627 — Detects access to sensitive files (credentials, configs, keys)
Location: setup-node-8207627.../__tests__/authutil.test.ts
```

The cited path is a **test fixture inside the action's own repository**, exercising npmrc handling — not behaviour this workflow invokes. The pin is the SHA **31 other workflows here already use**, and this PR changes no dependency manifest: the "2 changed dependencies" are simply `actions/checkout` and `actions/setup-node`, which *any* new workflow must reference. So this fires for every new workflow added to the repo, regardless of content.

### I looked at removing the flagged dependency, and decided against it

Dropping `actions/setup-node` would silence the scan. The job only needs Node to run `npx -p @loopover/mcp` and one `node --experimental-strip-types` script, and three workflows here (`ci.yml`, `mcp-release-watch.yml`, `package-release-watch.yml`) already run `npx` on the runner's ambient Node.

But `--experimental-strip-types` needs Node ≥ 22.6, and **every workflow here that uses it pairs it with `setup-node`** — there is no precedent for running it on the ambient runner Node. Removing the pin would make this nightly the first to depend on the runner image's Node version, untested, for a job that runs unattended at 04:12 UTC. Trading a scanner false positive for a nightly that might silently stop running is a bad trade, and dodging a scanner is not a good enough reason to make the workflow less deterministic.

### Options

- **A (recommended):** accept/suppress this finding in Superagent — it is a third-party test fixture at the repo-standard pin, and it will recur on every new workflow.
- **B:** I drop `setup-node` and add an explicit Node-version precondition that fails loudly if the runner is too old. Removes the finding; accepts a first-of-its-kind dependency on the runner image.
- **C:** fold the nightly into an existing workflow that already references these actions. Silences the scan, but couples an independent schedule and permission set to an unrelated workflow — I would not recommend it.

Happy to do B if you would rather the board be clean; otherwise this is ready to merge as-is.

JSONbored:
## Adjudicated at the pinned SHA, and hardened where it actually mattered

**The finding names a real capability that is unreachable in this workflow.** Verified against `actions/setup-node` at `820762786026740c76f36085b0efc47a31fe5020` rather than assumed:

- The SHA is the official **v7.0.0** tag.
- `action.yml` entrypoint is `dist/setup/index.js` (+ `dist/cache-save/index.js`). The cited `__tests__/authutil.test.ts` is a **test fixture** and is never executed by the action.
- The credential writer is `auth.configAuthentication(registryUrl)`, and `src/main.ts` calls it **only** inside a guard:
  ```ts
  const registryUrl: string = core.getInput('registry-url');
  if (registryUrl) {
    auth.configAuthentication(registryUrl);
  }
  ```
  This workflow does not set `registry-url`, so the `.npmrc`-writing path is not reached.

So there is no vulnerability to fix in the action or the pin. But the audit surfaced something that **is** worth fixing: the safety is a property of *this file*, not of the action. One innocuous line — `registry-url:` — would make a job that claims to hold no credentials write an authenticated `.npmrc`, and nothing would have noticed.

That is not hypothetical. The same class already shipped in this very file and was caught in review: `actions/checkout`'s `persist-credentials` default left a usable token in `.git/config` while the job asserted it had none.

### What changed (`a69c16c5f`)

The four credential paths are now **asserted against the shipped workflow** instead of described in a comment:

| Guarantee | Mutation | Result |
|---|---|---|
| checkout sets `persist-credentials: false` | set it back to `true` | 1 failed |
| setup-node has no `registry-url` | add one | 1 failed |
| the verify step has **no env at all** | add a token | 2 failed |
| no step runs `npm ci` | add one | 1 failed |

Asserted explicitly rather than absent-or-false, because `persist-credentials` defaults to true and an absent-or-false check would pass on the broken default. And "no env at all" rather than "no token", because an allowlist of forbidden names misses the next secret someone adds. The suite also asserts it found a verify job with steps, so it cannot pass vacuously if the file is renamed.

The audit is recorded on the step itself, pointing at the test rather than asking the next reader to trust the comment.

### The scan will still show red

Nothing here changes what it reports: it flags the two actions any new workflow must reference, and the HIGH is a test fixture in a third-party repo at the SHA 31 other workflows already pin. I stand by not removing `setup-node` — `--experimental-strip-types` needs Node ≥ 22.6, every workflow here that uses it pairs with `setup-node`, and making this nightly the first to depend on the runner image's Node version would trade a scanner false positive for a job that might silently stop running. Options A/B/C from my previous comment still stand; **A** remains my recommendation.

PR #9988, opened 2026-07-31
style(ui-kit): format chart.test.tsx, which left ui:lint red on main

JSONbored:
**`main`'s `ui:lint` has been failing since #9977 merged, and every PR inherits it.**

`packages/loopover-ui-kit/src/components/chart.test.tsx` — the tests I added in #9977 — was never run through prettier. `ui:lint` runs `format:check` across the ui-kit workspace and is part of `test:ci`:

```
[warn] src/components/chart.test.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
npm error workspace @loopover/ui-kit@1.3.1
npm error command sh -c prettier --check .
```

Confirmed against `origin/main`, not just locally.

## My regression, and the same cause as last time

I ran a checker subset on #9977 — `typecheck`, `dead-exports:check`, `test:ui-kit-pack`, the ui-kit suite — and not `ui:lint`, which is the one that would have caught it. That is the second time this session a subset sweep has let something through to main; the first was an integration test I hadn't run.

## The change

Prettier reflow only. No assertions, names, or behaviour altered — the 10 tests still pass, and the diff is entirely line wrapping of JSX props and multi-line call arguments.

Verified: `ui:lint` passes (warnings only, all pre-existing `react-refresh/only-export-components` notices), and `vitest run src/components/chart.test.tsx` is 10/10.

### DIFF
--- a/packages/loopover-ui-kit/src/components/chart.test.tsx
+++ b/packages/loopover-ui-kit/src/components/chart.test.tsx
@@ -10,7 +10,9 @@ import { describe, expect, it, vi } from "vitest";
 // asserting against a mock rather than against `chart.tsx`.
 vi.mock("recharts", async (importOriginal) => ({
   ...(await importOriginal<typeof import("recharts")>()),
-  ResponsiveContainer: ({ children }: { children?: React.ReactNode }) => <div data-testid="responsive">{children}</div>,
+  ResponsiveContainer: ({ children }: { children?: React.ReactNode }) => (
+    <div data-testid="responsive">{children}</div>
+  ),
 }));
 
 import {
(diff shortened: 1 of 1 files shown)


JSONbored:
**On the red `Workers Builds: loopover-ui`** — investigated, and it is not this PR.

It is Cloudflare-side, not GitHub Actions. The GitHub-side `Build UI preview artifact` job **passes** on this PR; the failing check comes from the `cloudflare-workers-and-pages` app.

Three things point away from our code:

1. **It never ran.** The check-run reports `started_at` and `completed_at` at the *same second* — a 0-second failure, with no log surfaced anywhere outside the Cloudflare dashboard.
2. **The exact build command succeeds locally.** I ran what Cloudflare runs — `npm run build:cloudflare`, which is `npm --prefix ../.. ci && npm --prefix ../.. run ui:build` — on this branch, from a clean install. Exit 0.
3. **It is intermittent on `main` itself**, with no code pattern: `5ee37c48b` succeeded, `d1f770670` failed, `364f5ba4a` failed then succeeded. `d1f770670` is #9984, which merged with this same check red.

So it is non-blocking and not caused by this change.

**Worth a look when you have dashboard access, though.** This is a *dashboard-configured* GitHub integration, independent of the repo's workflows — and it targets **production** (`/workers/services/view/loopover-ui/production/builds/...`) while running on pull-request branches. The repo's own `ui-deploy.yml` is deliberately `workflow_dispatch`-only, so there is a second, repo-invisible deploy path here. I cannot diagnose or reconfigure it: the Cloudflare MCP server is unauthorized in this session, and the build logs live only behind that dashboard.

Issue #9193, opened 2026-07-27
orb(dashboard): per-repo trust panel — ledger verify, anchor, decision records, calibration with coverage, attestation states


JSONbored:
### Context

Part of the ORB maintainer chat platform epic. The verification stack recently became real: the public ledger-verify endpoint (the #9120 fix), the external anchor with verify returning the chain hash (the #9122 fix), and published, digest-stable decision records (the #9123 fix) have all landed; the attested-backtest path (#8537/#8538) and per-tenant corpus export (#8539) are in flight. This issue renders that stack **as a product surface**: a per-repo trust panel inside the maintainer dashboard.

⚠️ **Blocked by the unified tenant verification contract issue (this epic's milestone-42 sibling)** — the panel renders that contract; do not invent a parallel taxonomy here. Do not start until it closes.

### Requirements

1. A per-repo trust panel presenting, from the verification contract's matrix: live ledger-verify status (real call, not cached decoration), the latest external-anchor reference, decision-record lookup by digest (paste a digest from a PR comment → see the full published record), and the calibration/fairness summary — accuracy always **with coverage and confidence interval**, plus the certified-guarantee line where a guarantee is active.
2. When attestation envelopes exist for a run (#8537), render the envelope reference and its verification path per #8538; absent envelopes render as "not yet attested", never as an error.
3. The "not verifiable" boundary from the contract (live-gate execution, per #9141's recorded decision) is stated in-panel, in the contract's language — honesty is the feature.
4. Every panel figure links to its public artifact (endpoint or docs walkthrough) so a skeptical maintainer can leave the dashboard and reproduce the number.
5. Chat integration: the grounding backend's ledger/calibration tools and this panel present identical numbers from identical sources — one implementation, two renderings.

### Deliverables

- The trust panel in `apps/loopover-ui`'s app area (ui-kit components), fed by existing public/protected endpoints — no new verification mechanisms.
- Tests: render states (verified / anchor-stale / record-missing / not-yet-attested), figure-formatting rules (coverage+interval always present with accuracy), link integrity to the contract's artifacts.

### Expected outcome

A maintainer can see — and independently re-derive — exactly how much of LoopOver's decision history is verified, from inside the product.

JSONbored:
## Design-pass addendum (2026-07-28) — presentation treatments for this panel

From the frontend design pass (system filed as #9568; prototype validates before implementation):

[...]

PR #9989, opened 2026-07-31
feat(proof): the public per-repo proof page, the last piece of #9569

### LINKED ISSUE #9569: Public proof page /proof/:owner/:repo + README badge — the shareable verification surface
### Context

The verification stack is live — public ledger-verify, external anchor with chain hash, digest-stable published decision records — and #9193 renders it as the in-app per-repo trust panel. This issue is the **public twin**: an unauthenticated, shareable per-repo proof page plus a README badge linking to it. Maintainer trust becomes distribution: every proof page markets a capability competitors without a verifiable ledger cannot ship. The public repo-quality page and badge endpoints are the existing pattern to follow (`apps/loopover-ui/src/routes/repos.$owner.$repo.quality.tsx`, the public badge routes).

### Requirements

1. **`/proof/:owner/:repo` public page:** decisions count, calibration — always with coverage and confidence interval, never a bare scalar — live ledger-verify status (real call, not cached decoration), latest external-anchor reference, and sample published decision records. Styled with the receipt/evidence treatment from the design-language issue.
2. **README badge:** an SVG badge endpoint linking to the proof page; cache-safe.
3. **One implementation, two renderings:** identical figures and sources as the in-app trust panel (#9193); every figure links to its public artifact so the number can be independently re-derived.
4. **Privacy boundary:** published/public data only; no wallet, hotkey, reward, trust-score, or private-ranking data anywhere.
5. **Honest boundary states:** not-yet-anchored and no-published-records render as neutral states, not errors; the "not verifiable" boundary is stated in the verification contract's language.
6. Decide and record: whether a maintainer can opt a repo out of the public proof page, and the default.

### Boundaries

- Read-only over existing public endpoints — no new verification mechanisms or data paths.
- Feature-flagged off by default, matching the epic's siblings.

### References

Relates to #9193 (in-app twin), #9183, #4935 (public status page), and the design-language issue.

JSONbored:
The API half already shipped. `/v1/public/repos/:owner/:repo/proof`, the badge SVG, the feature flag and the per-repo opt-out are all on `main`, all built on `buildProofSummary`. What was missing was the page the issue is named for — there was no `apps/loopover-ui/src/routes/proof.*`.

## One implementation, two renderings

This component **renders and computes nothing**. Every figure comes from the endpoint, which is the same composition the in-app trust panel (#9193) reads, so the two cannot disagree about a number. The only arithmetic is multiplying an already-computed rate by 100 for display — a percentage derived here would be a second implementation free to drift from the first, undermining the exact property the page exists to demonstrate.

## The states are the substance

Each is somewhere a plausible implementation says something untrue, so each is tested:

| State | What the page does | Why |
|---|---|---|
| Below the sample floor | Shows the decision **count**, no rate | "7 decisions, too few to claim a rate" is honest. Hiding the count with the figure, or printing `0%`, are not. |
| Not yet anchored | Neutral | A repo without an anchor is not failing — the chain is still self-verifying. |
| Empty ledger | Neutral: "a new repository, not a failing one" | Same. |
| **Broken** ledger | Stated as a problem, naming the row **and the kind** | The one state that must not be softened; the kind of break is the actionable half. |
| Unavailable | "not a claim that anything is wrong" | Could-not-check is not evidence of a fault. |
| Opted out (404) | **Empty** state, not an error | Different ARIA role, different meaning. Telling an opted-out repo something broke invites hunting a fault that does not exist. |
| Accuracy | Never without its denominator and Wilson interval | The difference between a claim someone can argue with and marketing. |

Two further details:

- **The boundary statement is rendered from the payload**, not written in this component, so a screenshot or embed cannot shed the caveat while keeping the numbers.
- **Digests are head-and-tail with the full value in `title`** — a truncated digest with no way back to the whole cannot be checked against anything.

## Mutation testing

| Mutation | Result |
|---|---|
| Publish a rate below the sample floor | 1 failed |
| Hardcode the boundary instead of reading the payload | 1 failed |
| Turn the 404 into an error instead of empty | 1 failed |

Baseline restored green.

## Issue requirements

1. `/proof/:owner/:repo` page — **this PR**
2. README badge SVG endpoint — already on main
3. One implementation, two renderings — shared `buildProofSummary`
4. Privacy boundary — allowlisted shape in `proof-summary.ts`
5. Honest boundary states — table above
6. Opt-out decision recorded — `isProofPageEnabledForRepo`

## Verification

- `ui:typecheck`, `typecheck`, `docs:drift-check`, `dead-source-files:check`, `dead-exports:check`, `import-specifiers:check`, `coverage-boltons:check`, `ui-derived-types:check` green
- Full `ui:test` suite green; 10 new tests
- `routeTree.gen.ts` regenerated via `ui:build` and committed
- Raw Tailwind text sizes converted to the repo's `text-token-*` design tokens, which `ui:lint`'s `no-restricted-syntax` rule enforces

**Note:** `ui:lint` currently fails on `main` for an unrelated reason — `chart.test.tsx` formatting, fixed in #9988. Not caused by this PR.

Closes #9569

### DIFF
--- a/apps/loopover-ui/src/components/site/public-proof-page.test.tsx
+++ b/apps/loopover-ui/src/components/site/public-proof-page.test.tsx
@@ -1,0 +1,177 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { PublicProofPage, type ProofSummary } from "./public-proof-page";
+
+// #9569: the public proof page.
+//
+// What is worth testing here is not that cards render — it is the JUDGEMENT the page encodes, because every
+// one of these states is a place where a plausible implementation says something untrue:
+//
+//   • a repo below the sample floor must show its decision COUNT and no rate. Hiding the count with the
+//     figure, or printing 0%, both misrepresent "we have 7 decisions, too few to claim a rate".
+//   • not-yet-anchored and empty-ledger are NEUTRAL. A page that renders a new repo as an error is lying in
+//     the more damaging direction than one that says nothing.
+//   • a genuinely BROKEN ledger is the one state that must read as a problem, and must name where it broke.
+//   • an opted-out repo (404) is an empty state, not an error state — different ARIA role, different meaning.
+//   • the boundary statement comes from the PAYLOAD. Hardcoding it here would let a screenshot or embed shed
+//     the caveat while keeping the numbers.
+
+const summary = (over: Partial<ProofSummary> = {}): ProofSummary => ({
+  schemaVersion: 1,
+  repoFullName: "acme/widgets",
+  decisionCount: 128,
+  accuracy: {
+    state: "published",
+    accuracy: 0.964,
+    decided: 112,
+    confirmed: 108,
+    interval: { lo: 0.912, hi: 0.987 },
+  },
+  ledger: {
+    state: "verified",
+    tipSeq: 128,
+    totalCount: 128,
+    checkedAt: "2026-07-31T12:00:00.000Z",
+  },
+  anchor: {
+    state: "anchored",
+    backend: "rekor",
+    seq: 128,
+    rowHash: "a".repeat(64),
+    at: "2026-07-30T00:00:00.000Z",
+  },
+  sampleRecords: [
+    {
+      pullNumber: 42,
+      action: "merge",
+      reasonCode: "gate_pass",
+      decidedAt: "2026-07-30T10:00:00.000Z",
+      recordDigest: "b".repeat(64),
+    },
+  ],
+  boundary:
+    "This page proves what was decided and that the record is intact. It does not prove the decisions were correct.",
+  ...over,
+});
+
+/** Renders with fetch stubbed to one response, then waits for the query to settle. */
+async function renderPage(response: { status: number; body?: unknown }): Promise<void> {
+  vi.stubGlobal("fetch", async () =>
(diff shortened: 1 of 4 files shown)


JSONbored:
**On the red `Workers Builds: loopover-ui`** — investigated, and it is not this PR.

It is Cloudflare-side, not GitHub Actions. The GitHub-side `Build UI preview artifact` job **passes** on this PR; the failing check comes from the `cloudflare-workers-and-pages` app.

Three things point away from our code:

1. **It never ran.** The check-run reports `started_at` and `completed_at` at the *same second* — a 0-second failure, with no log surfaced anywhere outside the Cloudflare dashboard.
2. **The exact build command succeeds locally.** I ran what Cloudflare runs — `npm run build:cloudflare`, which is `npm --prefix ../.. ci && npm --prefix ../.. run ui:build` — on this branch, from a clean install. Exit 0.
3. **It is intermittent on `main` itself**, with no code pattern: `5ee37c48b` succeeded, `d1f770670` failed, `364f5ba4a` failed then succeeded. `d1f770670` is #9984, which merged with this same check red.

So it is non-blocking and not caused by this change.

**Worth a look when you have dashboard access, though.** This is a *dashboard-configured* GitHub integration, independent of the repo's workflows — and it targets **production** (`/workers/services/view/loopover-ui/production/builds/...`) while running on pull-request branches. The repo's own `ui-deploy.yml` is deliberately `workflow_dispatch`-only, so there is a second, repo-invisible deploy path here. I cannot diagnose or reconfigure it: the Cloudflare MCP server is unauthorized in this session, and the build logs live only behind that dashboard.

Issue #9157, opened 2026-07-27
orb(selfhost): a malformed CRON_INTERVAL_MS spins the scheduler at 1ms and nothing validates any numeric env var — PORT and the GitHub cache TTL fail the same way


JSONbored:
## Problem

Numeric environment variables are read with bare `Number(...)` and validated nowhere. A wrong **unit** or
**type** — the most common operator mistake — produces `NaN`, and Node coerces `setTimeout(fn, NaN)` and
`setInterval(fn, NaN)` to a **1 ms** delay.

`src/server.ts:1172`:

```ts
const intervalMs = Number(process.env.CRON_INTERVAL_MS ?? 120_000);
```

`delayToNextWallClockBoundaryMs` (`src/selfhost/cron-alignment.ts:10-13`) is `nowMs % intervalMs` then
`intervalMs - msIntoCycle`, so NaN in → NaN out, and both the one-shot `setTimeout` and the follow-on
`setInterval` (`src/server.ts:1193-1195`) fire at 1 ms. `intervalMs = 0` takes the same path
(`x % 0` → NaN).

**Trigger:** an operator writes `CRON_INTERVAL_MS=2m`, `=120s`, `=120_000`, or `=0` to "disable" the
cron. Result: ~1000 scheduled ticks per second instead of one per two minutes, each running the full
`worker.scheduled()` fan-out — sweep dispatch, ops-alerts, reconciliation, registry refresh. Queue
flooded with duplicate maintenance jobs, GitHub REST budget burned, AI spend follows.

Same unguarded shape elsewhere:

- `src/server.ts:958` — `PORT` NaN → binds a random port, breaking the compose healthcheck.
- `src/server.ts:729` — `GITHUB_CACHE_TTL_SECONDS` NaN → `Math.max(0, NaN)` = NaN → `> 0` is false → the
  GitHub response cache **silently disables itself**, quietly multiplying API usage.

`preflightEnv` (`src/selfhost/preflight.ts:156-227`) checks only `REDIS_URL`, the GitHub App pair,
`DATABASE_URL`, and secret strength. It validates no numeric var's type or range, so nothing stops boot.

## Impact

A single-character config mistake produces a runaway that burns rate limit and AI spend, with no boot-time
error and no alert naming the cause. The cache case is worse in a way — it degrades silently and looks
like ordinary load.

## Requirements

[...]

PR #10098, opened 2026-07-31
fix(selfhost): route the last two bare-Number() env knobs through parsePositiveIntEnv + preflight

### LINKED ISSUE #10056: selfhost(config): route the two remaining bare-Number() env knobs through parsePositiveIntEnv and preflight

## Context

`parsePositiveIntEnv` (`src/selfhost/queue-common.ts:1103`) is this repo's contract for reading a numeric
self-host env knob: a missing value takes the fallback silently, and a **supplied** value that is non-finite
or out of range takes the fallback **with a structured warn line** (`warnEnvKnobRejected`,
`src/selfhost/queue-common.ts:1090`). `positiveInteger` in `src/selfhost/preflight.ts:259` is the boot-time
half of the same contract, and its own header states the reasoning:

```
// #9157: a bare `Number(process.env.X ?? default)` at a call site turns a wrong-unit/wrong-type operator
// mistake ("2m", "120s", "120_000", a fraction, a negative number) into NaN — and NaN silently becomes a
// runaway ... or a silent feature disable (a `> 0` gate on NaN is always false), with no boot-time signal
// either way.
```

`CRON_INTERVAL_MS`, `PORT` and `GITHUB_CACHE_TTL_SECONDS` were converted to that contract on both halves
(`src/selfhost/preflight.ts:381`–`src/selfhost/preflight.ts:383`; `src/server.ts:817`, `src/server.ts:1165`,
`src/server.ts:1396`). Two self-host numeric knobs were left on the bare-`Number()` form and are the only
remaining instances in `src/`:

```ts
// src/server.ts:1579
const forceReleaseAfterMs = Number(process.env["LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS"] ?? "");
const drainPromise = backend.shutdown();
const drainedInTime =
  Number.isFinite(forceReleaseAfterMs) && forceReleaseAfterMs > 0
    ? await Promise.race([...])
    : await drainPromise.then(() => true);
```

```ts
// src/selfhost/ai.ts:742
export function ollamaNumCtx(): number {
  const raw = Number(process.env["OLLAMA_NUM_CTX"] ?? "");
  return Number.isFinite(raw) && raw > 0 ? Math.floor(raw) : 32_768;
}
```

Both are `env.SOMETHING` reads that already appear in the generated self-host env reference
(`apps/loopover-ui/src/lib/selfhost-env-reference.ts:349` for the shutdown knob), so an operator can find and
set them — and both fail in exactly the two ways #9157 named:

1. **Silent feature disable, no warn.** `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS=30s` (or `2m`, or `30_000`)
   parses to `NaN`, `Number.isFinite` is false, and the shutdown path silently takes the "wait for the drain"
   branch — i.e. the opt-in the operator just enabled does nothing, with no log line, no metric, and no
   preflight error. `parsePositiveIntEnv` would have emitted `warnEnvKnobRejected`; `positiveInteger` would
   have failed boot with a message naming the unit-suffix mistake.
2. **A fractional value is accepted rather than rejected.** `Number("0.5")` is finite and `> 0`, so
   `Promise.race` gets a 0.5 ms deadline: every shutdown loses the race, `drainedInTime` is always `false`,
   and the proactive bulk `releaseAllHeldLocksAtShutdown()` fires on **every** shutdown — the exact
   behaviour the comment immediately above it (`src/server.ts:1566`–`src/server.ts:1578`) says was removed
   because it "could claim the freed lock at t0+e and duplicate the very actuation the lock exists to
   serialize". `parsePositiveIntEnv` floors and range-checks; `positiveInteger`'s `/^\d+$/` test rejects a
   decimal point outright.

   `OLLAMA_NUM_CTX=0.5` is accepted the same way and is then `Math.floor`ed to `0`, which is sent to Ollama as
   `options.num_ctx: 0` by `ollamaContextOptions` (`src/selfhost/ai.ts:731`) — a context window of zero,
   rather than the 32768 the fallback exists to provide.

Neither var is listed in `preflightEnv` (`src/selfhost/preflight.ts:381`–`src/selfhost/preflight.ts:383`), so
neither gets the boot-time half either.

## Requirements

- `src/server.ts:1579` must read the knob via `parsePositiveIntEnv` (already imported at `src/server.ts:76`).
  Because `0`/unset must continue to mean "wait for the drain, do not force-release", the read must keep that
  meaning explicitly: use `{ min: 0, fallback: 0 }` and keep the existing `> 0` gate on the result, so unset
  ⇒ `0` ⇒ wait for the drain (byte-identical to today), while a supplied non-integer/negative value now
  produces a warn line and falls back to `0` instead of silently doing the same thing unannounced.
- `ollamaNumCtx` (`src/selfhost/ai.ts:742`) must read via
  `parsePositiveIntEnv("OLLAMA_NUM_CTX", { min: 1, fallback: 32_768 })` and return that value directly. The
  existing default of `32768` and the "supplied valid value wins" behaviour must not change.
- `preflightEnv` (`src/selfhost/preflight.ts:306`) must gain two `positiveInteger(...)` calls alongside the
  existing three, so a malformed value hard-fails boot with a clear message rather than only warning at use
  time:
  - `positiveInteger(problems, env, "LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS", 0, 24 * 60 * 60_000)` — `0`
    is a valid, meaningful value here (it is the documented "wait for the drain" default), so the minimum is
    `0`, matching how `GITHUB_CACHE_TTL_SECONDS` is treated at `src/selfhost/preflight.ts:383`.
  - `positiveInteger(problems, env, "OLLAMA_NUM_CTX", 1, 1_000_000)` — `0` is not a meaningful context
    window, so the minimum is `1`.
- Absence must remain fine for both: `positiveInteger` already returns early on a blank value
  (`src/selfhost/preflight.ts:265`) and must keep doing so — this issue only judges a value the operator
  actually set.
- Must NOT change: `parsePositiveIntEnv` or `warnEnvKnobRejected` themselves, the three existing
  `positiveInteger` calls, the shutdown sequence's drain-first ordering
  (`src/server.ts:1580`–`src/server.ts:1587`), or `ollamaContextOptions`' provider gate
  (`src/selfhost/ai.ts:731`).
- No `migrations/*.sql` file may be added or edited by this issue; it needs no schema change.

> ⚠️ Required pattern: mirror `src/server.ts:1396`'s `parsePositiveIntEnv("CRON_INTERVAL_MS", { min: ..., fallback: ... })`
> call plus its paired `positiveInteger(...)` entry at `src/selfhost/preflight.ts:381`. What does NOT satisfy
> this issue: (a) writing a new local parse helper in `src/server.ts` or `src/selfhost/ai.ts` instead of using
> the existing `parsePositiveIntEnv` — a second mechanism is what #9157 removed; (b) fixing the runtime read
> but skipping the two `preflightEnv` entries, so a unit-suffix typo still only produces a warn line the
> operator has already scrolled past; (c) changing the default of either knob; (d) a test-only PR.

## Deliverables

- [ ] `src/server.ts`'s shutdown handler reads `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS` through
      `parsePositiveIntEnv` with `{ min: 0, fallback: 0 }` and keeps the `> 0` gate that selects the
      `Promise.race` branch.
- [ ] `ollamaNumCtx` (`src/selfhost/ai.ts`) reads through
      `parsePositiveIntEnv("OLLAMA_NUM_CTX", { min: 1, fallback: 32_768 })`.
- [ ] `preflightEnv` (`src/selfhost/preflight.ts`) adds the two `positiveInteger` entries described above.
- [ ] Tests in `test/unit/selfhost-preflight.test.ts` asserting `preflightEnv` reports a problem whose `var`
      is `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS` for each of `"30s"`, `"30_000"`, `"0.5"` and `"-1"`, and
      `ok: true` for each of unset, `""`, `"0"` and `"30000"`; and the same table for `OLLAMA_NUM_CTX` with
      `"0"` additionally reported as a problem and `"1"` accepted.
- [ ] A regression test named for this bug at `test/unit/selfhost-ai.test.ts` asserting
      `ollamaNumCtx()` returns `32768` (not `0`) when `OLLAMA_NUM_CTX` is stubbed to `"0.5"`, and returns
      `65536` when stubbed to `"65536"`.

All Deliverables above are required in a single PR. A PR that satisfies only some of them — for example one
that converts the two runtime reads but leaves `preflightEnv` unchanged — does not resolve this issue.

## Test Coverage Requirements

This repo enforces 99%+ Codecov patch coverage, **branch-counted**. `vitest.config.ts`'s `coverage.include`
covers `src/**/*.ts` and `packages/loopover-engine/src/**/*.ts`; `src/server.ts`, `src/selfhost/ai.ts` and
`src/selfhost/preflight.ts` are all measured. `positiveInteger` (`src/selfhost/preflight.ts:259`) has three
branches per call — blank/early-return, the `/^\d+$/` reject, and the range reject — and each new var must
exercise all three plus the accepted path. The `> 0` ternary at the `src/server.ts` shutdown call site needs
both arms: a run with the knob unset (falls to the plain `await drainPromise`) and one with a valid positive
value (takes the `Promise.race` branch).

## Expected Outcome

An operator who writes `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS=30s` or `OLLAMA_NUM_CTX=0.5` finds out at
boot, from a preflight error naming the mistake, instead of silently getting a disabled opt-in (or a
zero-width context window) that looks identical to never having set the variable. Every numeric self-host env
knob in `src/` then reads through the single `parsePositiveIntEnv` contract, with no bare `Number(env.X)`
left behind.

## Links & Resources

- `src/server.ts:1579` — the shutdown-lock knob's bare `Number()`
- `src/selfhost/ai.ts:742` — `ollamaNumCtx`
- `src/selfhost/queue-common.ts:1103` — `parsePositiveIntEnv`, the contract to use
- `src/selfhost/preflight.ts:259` — `positiveInteger`, the boot-time half
- `src/selfhost/preflight.ts:381` — the three existing entries to sit beside
- `src/server.ts:1566` — the comment explaining why an always-firing bulk lock release is harmful
- `apps/loopover-ui/src/lib/selfhost-env-reference.ts:349` — the shutdown knob in the generated reference

RealDiligent:
## Summary

`parsePositiveIntEnv` is this repo's contract (#9157) for a numeric self-host env knob: a supplied non-finite/out-of-range value takes the fallback **with a structured warn**, and `positiveInteger` is its boot-time half. `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS` (`src/server.ts:1579`) and `OLLAMA_NUM_CTX` (`src/selfhost/ai.ts:742`) were the only two self-host numeric knobs still on the bare-`Number(process.env.X ?? "")` form, and both fail exactly the two ways #9157 named:

1. **Silent feature disable.** `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS=30s` (or `2m`, `30_000`) → `NaN` → `Number.isFinite` false → the shutdown path silently takes "wait for the drain", i.e. the opt-in the operator just enabled does nothing, no log/metric/preflight error.
2. **A fractional value is accepted.** `Number("0.5")` is finite & `>0`, so the shutdown `Promise.race` gets a 0.5 ms deadline every shutdown loses → the bulk `releaseAllHeldLocksAtShutdown()` fires on *every* shutdown (the behaviour the comment above it says was removed for causing double-actuation). `ollamaNumCtx` floors `0.5` to `0` — a disabled context window.

## Fix

- `src/server.ts` reads the knob via `parsePositiveIntEnv("LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS", { min: 0, fallback: 0 })` and keeps the `> 0` gate — unset ⇒ `0` ⇒ drain-first, byte-identical, while a malformed value now warns and falls back to `0` instead of doing the same thing unannounced.
- `ollamaNumCtx` reads via `parsePositiveIntEnv("OLLAMA_NUM_CTX", { min: 1, fallback: 32_768 })`.
- `preflightEnv` gains the two paired `positiveInteger(...)` entries (`min 0` for the shutdown knob, matching `GITHUB_CACHE_TTL_SECONDS`; `min 1` for `OLLAMA_NUM_CTX`), so a unit-suffix typo hard-fails boot.

Defaults, the shutdown drain-first ordering, `parsePositiveIntEnv`/`positiveInteger` themselves, the three existing `positiveInteger` calls, and `ollamaContextOptions`' provider gate are all unchanged. `src/server.ts` is in `codecov.yml`'s ignore list (entrypoint), so its change is covered by the Docker boot smoke test, not unit coverage.

## Tests

- `test/unit/selfhost-preflight.test.ts`: `preflightEnv` reports a problem for `LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS` / `OLLAMA_NUM_CTX` on `"30s"`, `"30_000"`, `"0.5"`, `"-1"` (plus `"0"` for `OLLAMA_NUM_CTX`), and `ok: true` for unset/`""`/valid integers (`"0"` accepted for the shutdown knob, `"1"` for the context window).
- `test/unit/selfhost-ai.test.ts`: `ollamaNumCtx()` returns `32768` (not `0`) for `"0.5"` and `65536` for `"65536"`.

The `"0.5"` and malformed-preflight cases fail against the current bare-`Number()` code.

Closes #10056

### DIFF
--- a/src/selfhost/ai.ts
+++ b/src/selfhost/ai.ts
@@ -9,6 +9,7 @@
 import { isStructuralProviderConfigError } from "../services/ai-review";
 import type { AiContentBlock, CombineStrategy, OnMerge } from "../services/ai-review";
 import { isConfiguredSelfHostProvider, resolveConfiguredProviderNames } from "./ai-config";
+import { parsePositiveIntEnv } from "./queue-common";
 export { assertNoLegacySharedAiEnv } from "./ai-config";
 import { wasLoadedFromFile } from "./file-sourced-secrets";
 import { getProviderCredentialResolver } from "./provider-credential-registry";
@@ -740,8 +741,9 @@ export function ollamaContextOptions(
 /** Context window requested from Ollama for review-sized prompts. Overridable because it trades GPU memory
  *  against how much of a large diff the model can actually see. */
 export function ollamaNumCtx(): number {
-  const raw = Number(process.env["OLLAMA_NUM_CTX"] ?? "");
-  return Number.isFinite(raw) && raw > 0 ? Math.floor(raw) : 32_768;
+  // parsePositiveIntEnv (not a bare Number()): a supplied non-integer/out-of-range OLLAMA_NUM_CTX warns and
+  // falls back to the default instead of silently disabling the override via NaN, matching #9157's contract.
+  return parsePositiveIntEnv("OLLAMA_NUM_CTX", { min: 1, fallback: 32_768 });
 }
 
 export function providerNameFromBaseUrl(baseUrl: string | undefined): "ollama" | "openai" | "openai-compatible" {
--- a/src/selfhost/preflight.ts
+++ b/src/selfhost/preflight.ts
@@ -381,6 +381,10 @@ export function preflightEnv(env: SelfHostPreflightEnv): SelfHostPreflightResult
   positiveInteger(problems, env, "CRON_INTERVAL_MS", CRON_INTERVAL_MIN_MS, 24 * 60 * 60_000);
   positiveInteger(problems, env, "PORT", 1, 65_535);
   positiveInteger(problems, env, "GITHUB_CACHE_TTL_SECONDS", 0, 86_400);
+  // 0 is the documented "wait for the drain" default, so min 0 (mirrors GITHUB_CACHE_TTL_SECONDS above); #10056.
+  positiveInteger(problems, env, "LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS", 0, 24 * 60 * 60_000);
+  // 0 is not a meaningful context window, so min 1.
+  positiveInteger(problems, env, "OLLAMA_NUM_CTX", 1, 1_000_000);
 
   checkLedgerAnchorConfig(problems, env);
   checkLedgerContentWaiverConfig(problems, env);
--- a/src/server.ts
+++ b/src/server.ts
@@ -1576,7 +1576,11 @@ async function main(): Promise<void> {
     // is genuinely imminent and a stranded lock is the worse outcome. Opt in with
     // LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS; unset means "wait for the drain", which is right wherever the
     // orchestrator's grace period comfortably exceeds a review (this deployment's stop_grace_period is 300s).
-    const forceReleaseAfterMs = Number(process.env["LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS"] ?? "");
+    // parsePositiveIntEnv (not a bare Number()): a supplied non-integer/out-of-range value (e.g. "30s",
+    // "30_000", "0.5", "-1") now warns and falls back to 0 — "wait for the drain" — instead of silently
+    // taking that same branch (NaN) or accepting a fractional millisecond deadline every shutdown loses (#10056).
+    // { min: 0, fallback: 0 } keeps unset ⇒ 0 ⇒ the `> 0` gate below selecting the drain-first path, unchanged.
+    const forceReleaseAfterMs = parsePositiveIntEnv("LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS", { min: 0, fallback: 0 });
     const drainPromise = backend.shutdown();
     const drainedInTime =
       Number.isFinite(forceReleaseAfterMs) && forceReleaseAfterMs > 0
--- a/test/unit/selfhost-ai.test.ts
+++ b/test/unit/selfhost-ai.test.ts
@@ -2522,6 +2522,14 @@ describe("ollama context window (#9478)", () => {
     process.env["OLLAMA_NUM_CTX"] = value;
     expect(ollamaNumCtx()).toBeGreaterThan(8_192); // must exceed the truncation-prone stock defaults
   });
+
+  it("REGRESSION (#10056): a fractional OLLAMA_NUM_CTX falls back to the default instead of flooring to 0, and a valid value still wins", () => {
+    process.env["OLLAMA_NUM_CTX"] = "0.5"; // bare Number() floored this to 0 (a disabled context window)
+    expect(ollamaNumCtx()).toBe(32_768);
+    process.env["OLLAMA_NUM_CTX"] = "65536";
+    expect(ollamaNumCtx()).toBe(65_536);
+    delete process.env["OLLAMA_NUM_CTX"];
+  });
 });
 
 describe("resolveClaudeOauthToken (#9543 — rotate the credential without recreating the container)", () => {
--- a/test/unit/selfhost-preflight.test.ts
+++ b/test/unit/selfhost-preflight.test.ts
@@ -355,6 +355,38 @@ describe("self-host environment preflight (#2080)", () => {
       expect(preflightEnv({ ...baseEnv, GITHUB_CACHE_TTL_SECONDS: "0" })).toEqual({ ok: true, problems: [] });
     });
 
+    it("rejects a malformed LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS instead of only warning at use time (#10056)", () => {
+      for (const LOOPOVER_SHUTDOWN_LOCK_RELEASE_AFTER_MS of ["30s", "30_000", "0.5", "-1"]) {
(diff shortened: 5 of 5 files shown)


RealDiligent:
Superseded by #10108 — that PR includes the 
pm run selfhost:env-reference regen that failed validate-code here (stale selfhost-env-reference.ts). Keeping one open PR for #10056.

PR #10135, opened 2026-07-31
fix(discovery-index): bound TtlCache by a max-entry cap enforced on set

### LINKED ISSUE #10029: discovery-index(cache): bound TtlCache so a never-re-read key is actually freed instead of living for the process lifetime

## Context

`TtlCache` (`packages/discovery-index/src/cache.ts:13-54`) evicts **only** on a read of the same key:

```ts
  get(key: string): V | undefined {
    const entry = this.store.get(key);
    if (!entry) return undefined;
    if (entry.expiresAt <= this.now()) {
      this.store.delete(key);
      return undefined;
    }
    return entry.value;
  }
```

`delete` is the only other removal path, and it is called from exactly one place —
`SoftClaimStore.release` (`packages/discovery-index/src/soft-claim.ts:97-99`). There is no size cap, no
background sweep, and no eviction of entries whose key is never requested again. The file's own header comment
acknowledges the "no background sweep" design and justifies it with "entries this service caches — GitHub
issue metadata — are cheap to recompute on a stale miss", which addresses freshness but not **retention**: an
entry whose key is never read a second time is never freed at all.

This is not a Worker-isolate-lifetime concern. `packages/discovery-index/src/server.ts:28-35` runs the service
as a long-lived Node process (`@hono/node-server`) with three module-level cache instances that outlive every
request:

```ts
const app = createApp({
  github: new GitHubClient({ token: githubToken }),
  resultCache: new TtlCache<DiscoveryIndexCandidate[]>(),
  policyCache: new TtlCache<AiPolicyVerdict>(),
  cacheTtlMs,
  softClaimStore: new SoftClaimStore(new TtlCache(), softClaimTtlMs),
```

All three grow without bound from caller-controlled key space:

- `resultCache` keys on `scopeCacheKey(query)` — `JSON.stringify({repos, orgs, searchTerms})`
  (`packages/discovery-index/src/discovery-query.ts:129-135`) — i.e. arbitrary caller-supplied strings, with a
  full `DiscoveryIndexCandidate[]` as the value. Every distinct scope any miner ever queries is retained.
- `policyCache` keys on `repoFullName`, one entry per repo ever seen.
- The soft-claim cache keys on `${repoFullName}#${issueNumber}` (`soft-claim.ts:58-60`), one entry per issue
  ever claimed; only an explicit `action: "release"` frees it, and a client that claims and never releases
  leaves it resident forever.

Every caller shares one bearer secret (`DISCOVERY_INDEX_SHARED_SECRET`,
`packages/discovery-index/src/app.ts:59-68`, `:96-104`), so there is no per-caller identity to bound growth
by; the IP rate limiter (60/min, `packages/discovery-index/src/rate-limiter.ts:15-16`) bounds the *rate* of new
keys, not the total. At 60 distinct scopes/minute the result cache accumulates unboundedly across a container
lifetime while `cache.size` never falls, because nothing re-reads those keys.

## Requirements

- `TtlCache` MUST accept a maximum entry count and MUST enforce it in `set`. When the cap would be exceeded,
  it MUST first drop every already-expired entry, and if that is not enough, evict the oldest-inserted entry
  (insertion order, which a `Map` already preserves) until the store is at or below the cap.
- The cap MUST be a constructor parameter with a named exported default constant in
  `packages/discovery-index/src/cache.ts`, so each of the three call sites in
  `packages/discovery-index/src/server.ts:30-33` can pass its own value.
- `packages/discovery-index/src/server.ts` MUST pass an explicit cap for all three instances (`resultCache`,
  `policyCache`, and the `SoftClaimStore`'s cache). `SoftClaimStore`'s constructor
  (`packages/discovery-index/src/soft-claim.ts:78-82`) MUST keep its current signature — pass the already-capped
  `TtlCache` in, do not add a cap parameter there.
- `get`'s existing lazy expiry-on-read MUST be preserved exactly: an expired entry is still deleted and still
  returns `undefined`, and `size` still counts not-yet-lazily-evicted entries (this is what
  `test/unit/discovery-index/cache.test.ts:20-23` already asserts).
- `getOrCompute`'s contract MUST be unchanged: a live cached value short-circuits, otherwise `compute()` runs
  and its result is cached.
- Do NOT add a timer, `setInterval`, or any background sweep — the service must stay free of process-level
  timers; eviction happens on `set`.
- Do NOT change the TTL values or the env-var resolution in `packages/discovery-index/src/server.ts:21-26`.

> ⚠️ Required pattern: bound the store inside `TtlCache.set` using the `Map`'s own insertion order, keeping the
> class free of timers — the same "small, self-contained, no new dependency" shape the file's header comment
> commits to. What does NOT satisfy this issue: (a) a background `setInterval` sweeper in `server.ts`;
> (b) swapping in an external LRU dependency — this package deliberately carries no cache dependency (see the
> file header); (c) capping only `resultCache` and leaving `policyCache` and the soft-claim cache unbounded;
> (d) a test-only PR that asserts today's unbounded behaviour.

## Deliverables

- [ ] `TtlCache` in `packages/discovery-index/src/cache.ts` takes a max-entry cap and enforces it in `set`.
      Exact expectation: with a cap of 2, `set("a",…); set("b",…); set("c",…)` leaves `cache.size === 2`,
      `cache.get("a") === undefined`, and `cache.get("c")` returning the value just set.
- [ ] Expired-first eviction: with a cap of 2 and `"a"` already expired, `set("c",…)` evicts `"a"` and leaves
      `"b"` live — asserted in `test/unit/discovery-index/cache.test.ts`.
- [ ] An exported default cap constant in `packages/discovery-index/src/cache.ts`, passed explicitly at all
      three construction sites in `packages/discovery-index/src/server.ts:30-33`.
- [ ] A test in `test/unit/discovery-index/cache.test.ts` pinning the unchanged behaviours: lazy
      expiry-on-read still deletes and returns `undefined`; `getOrCompute` still returns the cached value
      without calling `compute` a second time.
- [ ] A regression test at `test/unit/discovery-index/cache.test.ts` named for this bug (e.g.
      `"REGRESSION: a key that is never re-read must not survive past the entry cap"`) that writes cap+50
      distinct keys without ever reading them back and asserts `cache.size` never exceeds the cap.

All Deliverables above are required in a single PR. A PR that satisfies only some of them — for example one
that adds the cap to `TtlCache` but leaves `packages/discovery-index/src/server.ts` constructing all three
instances with the parameter defaulted/omitted — does not resolve this issue.

## Test Coverage Requirements

This repo enforces 99%+ Codecov patch coverage, **branch-counted**. `vitest.config.ts`'s `coverage.include`
covers `src/**/*.ts` and `packages/loopover-engine/src/**/*.ts` — `packages/discovery-index/src/cache.ts` is
**not** in `coverage.include`, so `codecov/patch` does not gate this path. The tests are still mandatory as
written above: `test/unit/discovery-index/cache.test.ts` already exists and is part of `npm run test:ci`, and
the deliverables are only verifiable through it.

Both arms of every new branch must be exercised regardless: store-at-cap vs under-cap in `set`, expired-entry
eviction sufficient vs insufficient to get under the cap, and the cap-parameter-provided vs defaulted paths.
`packages/discovery-index/src/server.ts` is a process entrypoint and is not unit-imported (see its own header
comment), so assert the constructor wiring by reading the exported constant, not by importing `server.ts`.

This change is NOT in `packages/loopover-engine/src/**`, so the dual-upload engine-coverage rule does not apply.

## Expected Outcome

The discovery-index service's three in-memory caches have a hard entry ceiling, so a long-running container
serving many distinct query scopes, repos, or soft-claims reclaims memory instead of retaining every key it has
ever seen for the process lifetime. Lazy expiry-on-read, TTL semantics, and `getOrCompute` behave exactly as
before.

## Links & Resources

- `packages/discovery-index/src/cache.ts:13-54` — `TtlCache`: eviction only on same-key read, no size cap
- `packages/discovery-index/src/server.ts:28-35` — the three long-lived module-level cache instances
- `packages/discovery-index/src/discovery-query.ts:129-135` — `scopeCacheKey`, caller-controlled key space
- `packages/discovery-index/src/soft-claim.ts:58-60`, `:77-99` — one cache entry per claimed issue, freed only
  by an explicit release
- `packages/discovery-index/src/app.ts:59-68` — the single shared bearer secret (no per-caller identity)
- `test/unit/discovery-index/cache.test.ts:15-40` — the existing lazy-expiry assertions that must keep passing

bitfathers94:
fix(discovery-index): bound TtlCache by a max-entry cap enforced on set

Without a cap, a key that is never re-read (e.g. a scope no caller repeats,
or a soft-claim that's never released) stays resident in the long-lived
module-level caches for the whole process lifetime. set() now drops
already-expired entries first, then evicts the oldest-inserted entry until
the store is back at or under the cap.

Closes #10029

### DIFF
--- a/packages/discovery-index/src/cache.ts
+++ b/packages/discovery-index/src/cache.ts
@@ -3,17 +3,24 @@
 // (src/selfhost/redis-cache.ts) live in the main Cloudflare Worker package and aren't importable from a
 // separate npm workspace package — there's no in-repo precedent for a plain in-memory keyed-with-expiry
 // cache, so this is net new, kept deliberately small (lazy expiry check on read, no background sweep;
-// entries this service caches — GitHub issue metadata — are cheap to recompute on a stale miss).
+// entries this service caches — GitHub issue metadata — are cheap to recompute on a stale miss). A
+// constructor-supplied max-entry cap, enforced in `set`, bounds retention for keys that are never re-read.
 
 interface Entry<V> {
   value: V;
   expiresAt: number;
 }
 
+/** Default max-entry cap for a TtlCache with no explicit cap passed to its constructor. */
+export const DEFAULT_CACHE_MAX_ENTRIES = 5_000;
+
 export class TtlCache<V> {
   private readonly store = new Map<string, Entry<V>>();
 
-  constructor(private readonly now: () => number = Date.now) {}
+  constructor(
+    private readonly now: () => number = Date.now,
+    private readonly maxEntries: number = DEFAULT_CACHE_MAX_ENTRIES,
+  ) {}
 
   /** Returns the cached value, or undefined if absent or expired (an expired entry is evicted on read). */
   get(key: string): V | undefined {
@@ -26,7 +33,24 @@ export class TtlCache<V> {
     return entry.value;
   }
 
+  /** Bounds the store to `maxEntries` before inserting a NEW key: first drops every already-expired entry,
+   *  then (if still at cap) evicts the oldest-inserted entries -- the Map's own insertion order -- until
+   *  there's room. Overwriting an already-present key never grows the store, so it skips this entirely. */
+  private evictForCapacity(key: string): void {
+    if (this.store.has(key) || this.store.size < this.maxEntries) return;
+    const now = this.now();
+    for (const [storedKey, entry] of this.store) {
+      if (entry.expiresAt <= now) this.store.delete(storedKey);
+    }
+    while (this.store.size >= this.maxEntries) {
+      const oldestKey = this.store.keys().next().value;
+      if (oldestKey === undefined) break;
+      this.store.delete(oldestKey);
+    }
+  }
+
   set(key: string, value: V, ttlMs: number): void {
+    this.evictForCapacity(key);
     this.store.set(key, { value, expiresAt: this.now() + Math.max(0, ttlMs) });
   }
 
--- a/packages/discovery-index/src/server.ts
+++ b/packages/discovery-index/src/server.ts
@@ -7,7 +7,7 @@
 import { serve } from "@hono/node-server";
 import type { AiPolicyVerdict, DiscoveryIndexCandidate } from "@loopover/engine";
 import { createApp } from "./app.js";
-import { TtlCache } from "./cache.js";
+import { DEFAULT_CACHE_MAX_ENTRIES, TtlCache } from "./cache.js";
 import { DEFAULT_CACHE_TTL_MS } from "./discovery-query.js";
 import { GitHubClient } from "./github-client.js";
 import { captureUnhandledPostHogError, flushDiscoveryIndexPostHog, initDiscoveryIndexPostHog, resolvePostHogEnvironment, shutdownDiscoveryIndexPostHog } from "./posthog.js";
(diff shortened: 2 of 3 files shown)


JSONbored (commented):
Failing CI, closing.

PR #10268, opened 2026-07-31
ci: run the contract api-schemas drift check in CI, not only in the local gate

### LINKED ISSUE #10237: contract: api-schemas.ts is half-generated — regenerating it deletes live discovery-route schemas, and the check is not wired into CI
## Problem

`npm run contract:api-schemas:check` fails on current `main`, and **running the generator to fix it destroys working code**.

`packages/loopover-contract/src/api-schemas.ts` opens with:

```
// GENERATED by scripts/gen-contract-api-schemas.ts -- do not edit.
```

but it currently contains hand-maintained content the generator does not produce. Regenerating removes it:

```diff
-import { FindOpportunitiesInput, RetrieveIssueContextInput } from "./tools/local-branch.js";
-export const FindOpportunitiesRequestSchema = z.object(FindOpportunitiesInput.shape);
-export const IssueRagRetrieveRequestSchema = z.object(RetrieveIssueContextInput.shape);
+// #9310 — request/response schemas for the two discovery routes below, mirroring the MCP tools'
+// own Zod shapes verbatim ...
```

`FindOpportunitiesRequestSchema` is live — `api-schemas.ts:2322` maps it to `POST /v1/opportunities/find`. Regenerating deletes both the symbol and its import while leaving the reference, so the file stops compiling.

## Why this is the dangerous shape

`test/unit/contract-limits-pinned.test.ts:18` already records this exact failure once before: *"left `main` failing `contract:api-schemas:check` for every PR: regenerating produced an uncompilable file."* It has recurred.

The generator is behind the #10040/#10225 discovery-route work — those schemas were added to the generated file without the generator learning to emit them, so the artifact and its generator now disagree permanently. The check correctly reports drift; the documented remedy ("run `npm run contract:api-schemas` and commit the result") is actively harmful here.

## Two contributing factors, both worth naming

1. **`contract:api-schemas:check` is in the local `test:ci` aggregate but NOT in `ci.yml`.** Nothing in CI runs it, which is why this drifts silently and is only discovered by whoever next runs the full local gate. Every other generated-artifact drift check (`db:migrations:check`, `cf-typegen:check`, `selfhost:env-reference:check`, `command-reference:check`) IS wired into `ci.yml`.
2. A separate, older drift also exists in this file: #10160 added `linkedIssueMaintainerExempt` to the repository-settings surface without regenerating. A fix for that was carried in #10233 and did not survive the merge — the PR listed 16 files, the merge commit `ba35c8bab` has 15, and `api-schemas.ts` is the missing one. Worth understanding on its own, since a silently-dropped file in a gate merge is a bigger problem than this file.

## Deliverables

- [ ] Decide the source of truth: either teach `scripts/gen-contract-api-schemas.ts` to emit the discovery-route schemas, or stop generating this file and drop the "do not edit" header. It cannot stay half-generated.
- [ ] `npm run contract:api-schemas` produces a file that **compiles** — regression-guarded, since this is the second occurrence.
- [ ] Fold in the `linkedIssueMaintainerExempt` regeneration from #10160.
- [ ] Wire `contract:api-schemas:check` into `ci.yml` alongside the other generated-artifact drift checks, so this cannot drift unobserved a third time.

JSONbored:
## What

`contract:api-schemas:check` was reachable only from the local `npm run test:ci` aggregate — **nothing in
CI ran it**. This wires it into `ci.yml` alongside the other generated-artifact drift checks.

Closes #10237

## Two of the issue's four deliverables are already resolved on main

Verified before writing anything, since the issue's central claim is now stale:

- *"`npm run contract:api-schemas:check` fails on current `main`"* — it **passes**.
- *"running the generator to fix it destroys working code"* — it no longer does. Running
  `npm run contract:api-schemas` on current main produces **an empty diff**. #10225 taught the generator to
  emit the discovery-route schemas, so the artifact and its generator no longer disagree.
- The `linkedIssueMaintainerExempt` regeneration landed with #10225 too.

What remains live is the issue's **contributing factor 1**, which is the reason both drifts went unnoticed
in the first place — and that is what this PR fixes.

## The gating

The generator has three inputs, and only two were covered by existing filters:

| input | filter |
|---|---|
| `src/openapi/schemas.ts` | `backend` ✓ |
| `apps/loopover-ui/public/openapi.json` | `ui` ✓ |
| the contract package's own modules (scanned to emit imports) | **nothing watched `packages/loopover-contract/**`** |

So a contract-only PR would have skipped the very check it most needs. This adds a `contract` filter shaped
exactly like its `engine` / `mcp` / `miner` siblings.

## The "must produce a file that compiles" deliverable

Satisfied structurally rather than with a new test, and the pairing is worth stating explicitly:

- `contract:api-schemas:check` proves **committed == generated**;
- `typecheck:packages` runs `tsc -p packages/loopover-contract/tsconfig.json`, whose `include` is `src`, so
  it compiles the generated file.

Together they prove the generated file compiles. Both now run in CI; previously the first held only if
someone happened to run the local gate.

## Verification

- `npm run contract:api-schemas:check` — passes on this base.
- `npm run contract:api-schemas` — empty diff, confirming the generator is no longer destructive.
- `npm run actionlint` — clean.

## Noted separately, not fixed here

While measuring this I found `contract:api-schemas:check` is not the exception — **22 of the 42 drift and
wiring checks reachable from `test:ci` run in no workflow at all** (measured against `main`, this one
included), including `checkers-wired:check`,
`dead-exports:check`, `typecheck-coverage:check` and `test-wiring:check`. Fixing that wholesale is a
maintainer call about CI runtime and per-check gating, not something to smuggle into this PR, so it is
filed on its own.


JSONbored:
Note on the red **Superagent Supply Chain Scan** — it is a pre-existing false positive, not something this
diff introduced.

It reports `actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9` as an *added* dependency with three
HIGH `threat-runtime-keylogging` hits in `dist/restore-only/index.js`. That pin is already on `main` at the
identical SHA (`.github/workflows/ci.yml` lines 600, 663, 728), and this PR's diff references
`actions/cache` **zero** times — it adds one `run:` step and one path filter. The scanner appears to
re-evaluate every action referenced by a changed workflow file and report them as new.

Precedent that it does not block: #10217 (`ci(ui-kit): wire ui-kit's vitest suite into ui:test/CI`) merged
with `Superagent Supply Chain Scan = FAILURE` and `Security Scan = SUCCESS`, the same pair as here.

Flagging rather than acting on it: silencing a supply-chain finding is not something to do from inside an
unrelated PR, and the heuristic matching `actions/cache`'s bundled webpack dist is worth its own look if it
is going to fail every PR that edits a workflow.

JSONbored:
Superseded by #10280, which contains this change verbatim (the Contract api-schemas drift check step, its conditional, and the `contract` path filter) plus the meta-checker that guards the whole wired-check list against drifting again. Closing rather than carrying both: this PR's only red check is a Superagent scanner false-positive on the unchanged official actions/cache v6.1.0 pin (rule `threat-runtime-keylogging`, fired for any PR touching ci.yml since 2026-07-31 ~15:28 UTC) that has no repo-side fix, while #10280's rescan already comes back neutral.

PR #10280, opened 2026-07-31
ci: run every local-gate drift check in CI, and guard the two lists against drifting again

### LINKED ISSUE #10237: contract: api-schemas.ts is half-generated — regenerating it deletes live discovery-route schemas, and the check is not wired into CI

### LINKED ISSUE #10269: ci: 22 of 42 test:ci drift checks run in no workflow — including the meta-guard that detects unwired checkers
## Problem

`npm run test:ci` is documented as the local gate, and `ci.yml` is what actually blocks a merge. They have
drifted: **22 of the 42 drift/wiring checks reachable from `test:ci` run in no workflow at all.**

They are therefore enforced only when a human happens to run the full local aggregate. On a repo whose gate
auto-merges on green CI, a check nothing in CI runs is not a gate — it is a convention.

Measured against `main` by walking `npm run` references transitively from `test:ci` and matching each
resulting `*:check` / `*drift-check` / `*:audit` script against every file in `.github/workflows/`:

| check | script |
|---|---|
| `db:migrations:immutable:check` | `scripts/check-released-migrations-immutable.ts` |
| `turbo-inputs:check` | `scripts/check-turbo-typecheck-inputs.ts` |
| `workspace-dep-ranges:check` | `scripts/check-workspace-dep-ranges.ts` |
| `control-plane:contract:check` | control-plane contract drift |
| `control-plane:openapi:check` | control-plane OpenAPI drift |
| `coverage-boltons:check` | `scripts/check-coverage-bolt-on-filenames.ts` |
| `ui-derived-types:check` | `scripts/check-ui-derived-types.ts` |
| `server-manifest:check` | `scripts/check-server-manifest.ts` |
| `dead-exports:check` | `scripts/check-dead-exports.ts` |
| `publishable-deps:check` | `scripts/check-publishable-deps.ts` |
| `fixture-clock-races:check` | `scripts/check-fixture-clock-races.ts` |
| `typecheck-coverage:check` | `scripts/check-typecheck-coverage.ts` |
| `test-wiring:check` | `scripts/check-test-coverage-wiring.ts` |
| `checkers-wired:check` | `scripts/check-checkers-wired.ts` |
| `focus-manifest-fields:check` | `scripts/check-focus-manifest-fields.ts` |
| `release-commit-parsing:check` | `scripts/check-release-commit-parsing.ts` |
| `releasable-commit-types:check` | `scripts/check-releasable-commit-types.ts` |
| `maintainer-associations:check` | `scripts/check-maintainer-association-copies.ts` |
| `release-linked-versions:check` | `scripts/check-release-linked-versions.ts` |
| `mcp:tool-reference:check` | MCP tool-reference drift |
| `mcp:client-config:check` | MCP client-config drift |
| `contract:api-schemas:check` | **being wired by #10268** |

## Why this is worth acting on rather than noting

Three of these are load-bearing in a way that makes the gap self-concealing:

- **`checkers-wired:check`** is the meta-guard from #9860 that fails when a `scripts/check-*.ts` runs
  nowhere. It itself runs nowhere in CI. So the mechanism that detects unwired checkers is unwired.
- **`db:migrations:immutable:check`** guards the invariant that a shipped `migrations/*.sql` is never
  edited. Violating it breaks boot for **every upgraded Orb** — the highest-blast-radius rule in the repo,
  and CI does not check it.
- **`test-wiring:check`** and **`typecheck-coverage:check`** exist to stop coverage and typecheck silently
  narrowing. Narrowing them is invisible precisely because these do not run.

This is not hypothetical: #10237 is the same class having already fired twice on one artifact, both times
discovered by a human running the local gate rather than by CI.

## The shape of the fix is a decision, not an obvious default

Wiring all 22 is not simply "add 22 steps". Each needs a `needs.changes.outputs.*` gate chosen from its real
inputs — the mistake #10268 had to fix for its own check was that no filter watched
`packages/loopover-contract/**` at all, so the naive gating would have skipped the exact PRs that matter.
CI runtime is also an explicit constraint here (it is what moved `selfhost.yml` post-merge).

Roughly:

**A. Wire all 22 into `ci.yml`, individually gated.** Highest fidelity, most review effort, some CI cost.
Most are sub-second node scripts, so the cost is likely dominated by job setup rather than the checks.

**B. One `drift-checks` job that runs the whole set ungated on every PR.** Cheapest to write and to keep
correct — no per-check gating to get wrong, and no filter can be missing. Pays one job's setup cost per PR.
Loses per-check path gating.

**C. Wire only the load-bearing subset** (`db:migrations:immutable`, `checkers-wired`, `test-wiring`,
`typecheck-coverage`, `dead-exports`, `publishable-deps`), leave the rest local-only, and **document** that
`test:ci` is deliberately a superset.

**D. Add a mechanical guard** that fails when a check reachable from `test:ci` is absent from every
workflow, with a reasoned allowlist — the `ALLOWED_UNWIRED` idiom `check-checkers-wired.ts` already uses.
Composes with any of A–C and is what stops the gap reopening.

**Recommendation: B + D.** B removes the per-check gating surface entirely, which is where the fidelity bugs
actually live, and one extra job on a PR is cheap next to the existing fan-out. D is what makes it stay
true. A is the "right" answer on paper but its 22 hand-chosen gates are precisely the hand-maintained list
#9860 exists to reject, and C leaves the meta-guard unguarded.

## Deliverables

- [ ] Decide between A–D above.
- [ ] Every check reachable from `test:ci` either runs in a workflow or is explicitly recorded as
      local-only with a reason.
- [ ] A mechanical guard so a newly added `test:ci` check cannot silently be CI-invisible.

## How this was found

Walking `npm run` references transitively out of `test:ci` and diffing against every workflow file, while
fixing the single instance in #10237. The single-instance fix is #10268; this is the class.

JSONbored:
## What

`npm run test:ci` (the documented local gate) and `ci.yml` (what actually blocks a merge) are two
independently hand-maintained lists, and they had drifted: **22 of the 42 drift checks reachable from
`test:ci` ran in no workflow at all.** This wires all of them and adds the guard that stops it reopening.

Closes #10269
Closes #10237

Implements **B + D** from the issue's options, with one correction found by measuring (below).

## Why it matters, concretely

- **`checkers-wired:check`** is the #9860 meta-guard that fails when a `scripts/check-*.ts` runs nowhere.
  It was itself not run by CI. The mechanism for detecting unwired checkers was unwired.
- **`db:migrations:immutable:check`** guards the rule whose violation breaks boot for *every upgraded Orb*.
- **`test-wiring:check`** / **`typecheck-coverage:check`** exist to stop coverage and typecheck silently
  narrowing — invisible precisely because they did not run.

## One ungated block, not 19 gated steps

Per-check path gating is where the fidelity bugs actually live: the sibling PR #10268 had to add an entire
`contract` path filter because *nothing* in `ci.yml` watched `packages/loopover-contract/**`. Nineteen
hand-chosen gates would be exactly the hand-maintained list #9860 exists to reject. These are all
sub-second file-reading scripts, so running them unconditionally costs far less than choosing 19 gates
correctly.

## The correction to my own recommendation

I recommended "one ungated job, no gating to get wrong". Measuring showed that is wrong for three checks,
which need git history `validate-code`'s checkout does not have (it is shallow *on purpose*). They degrade
in two different, equally bad ways:

| check | in a shallow checkout |
|---|---|
| `db:migrations:immutable:check` | exits 1, `"no orb-v* tags visible"` → **fails every PR** |
| `release-commit-parsing:check` | returns `[]` and **passes** |
| `releasable-commit-types:check` | prints `"cannot resolve …; skipping"`, **exits 0** |

The last two are the important ones: wired naively they would be **permanently green while verifying
nothing** — the exact "guards nothing while looking like a guard" failure this PR exists to remove,
reproduced at the CI layer and then read as coverage. That would have been worse than leaving them unwired.

So they run in a new **`drift-checks-history`** job with `fetch-depth: 0` + `fetch-tags`, which asserts the
history is actually present before running them rather than trusting it. The `validate` aggregator gains a
dependency on that job, since a job nothing depends on cannot fail a PR.

## The guard (D)

`scripts/check-ci-drift-checks-wired.ts` is `check-checkers-wired.ts`'s sibling on the other axis. That one
asks *"does this checker run anywhere?"* — satisfied by the local gate alone. This one asks the question
that decides merge safety: *"does CI run it?"* Both sides are **computed, never listed**, so the roster
cannot drift the way `ci.yml` drifted from `test:ci`. `ALLOWED_LOCAL_ONLY` exists and is deliberately
empty.

It found a scoping bug in itself during review: the repo names audits `-audit`, not `:audit`, so my first
pattern matched neither `ui:version-audit` nor `test:miner-deployment-docs-audit` — silently excluding them
while appearing to cover them. Fixed and pinned by a test.

## Relationship to #10268

This branch **includes #10268's commit**, because the new guard cannot pass without it —
`contract:api-schemas:check` was one of the 22. The hunks are in different regions of `ci.yml` and are
identical, so whichever lands first, the other merges cleanly.

**If #10268 merges first**, its hunk is already here and this PR is unaffected. **If this one merges
first**, #10268's diff becomes empty and the gate will auto-close it as an empty PR — which would leave
#10237 open even though its work had shipped. That is why `Closes #10237` is listed above as its own line
(a combined "Closes #A and #B" only closes the first). Either order now closes both issues correctly.

## Verification

- Full suite green: **26,715 passed / 0 failed**, 1367 files.
- `npm run ci-drift-checks-wired:check` → *all 45 local-gate drift check(s) also run in a workflow*.
- All 21 newly wired checks run clean locally; `actionlint`, `typecheck` (root + packages + ui),
  `checkers-wired:check` clean.
- **An existing test caught a real omission** — `ci-skip-draft-prs.test.ts` enumerates the aggregator's
  `needs`, so adding the job without aggregating it failed. Updated, and the new job is also asserted to
  honour the draft gate.
- **Mutation-tested:** dropping a check from the block, moving a history-dependent check into the
  unconditional block, removing the aggregator dependency, and removing the word-boundary guard from the
  checker's matcher each fail a test; the restored tree passes.
- `codecov/patch` does not grade this diff (`scripts/**` is codecov-ignored; workflows are not source).
  Tests written to the same standard regardless: 22 across the two files.


JSONbored:
Rebased onto `main` after #10266 merged, and the guard in this PR immediately paid for itself.

#10266 added `dispatch-provenance:check` to `test:ci` and — like the 22 before it — wired it into no
workflow. On the first run after the rebase:

```
check-ci-drift-checks-wired: 1 of 46 check(s) run in the local gate but in NO workflow (#10269):

  npm run dispatch-provenance:check
```

That is the 23rd instance of this class, introduced *during* the work to fix the class, by me, and caught
mechanically rather than by anyone remembering. It is now in the unconditional block.

Also refreshed: the rebase pulled in `chore(release): cut orb-v3.7.0` and the engine-and-dependents release,
which left the miner `dist/` stale against the bumped `package.json` and failed `miner-cli.test.ts`'s
`--version` assertion. Rebuilt the workspaces; that is the known stale-dist artifact, not a code defect.

Full suite re-verified after the rebase: **26,745 passed / 0 failed**.

JSONbored:
CI caught a real ordering bug in this PR, which is worth recording because it is the same class the PR is
about and my own local verification missed it.

`validate-code` failed on `mcp:tool-reference:check`:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '/home/runner/work/loopover/loopover/node_modules/@loopover/engine/dist/index.js'
```

That check resolves a **bare `@loopover/engine` specifier through node_modules to `dist/`** — a *transitive*
prerequisite. My pre-flight grep for `dist/` in the generator script found nothing because the dependency is
not in that file, and it passed locally only because my worktree happened to have a built engine. The
unconditional block runs ~100 lines before `Build engine package`.

Reproduced properly by deleting `packages/loopover-engine/dist` and re-running all 19 checks: **exactly one**
has a build prerequisite. `mcp:client-config:check`, the obvious co-suspect, does not.

Fix: `mcp:tool-reference:check` moved to its own step immediately after `Build engine package`, carrying
that build's condition **verbatim**. The gate is *derived* — "run iff the thing I need was built" — rather
than hand-picked from the check's own inputs, which is the distinction that keeps it out of the
hand-maintained-gate class the unconditional block exists to avoid.

Pinned by a test that asserts all three properties (not in the unconditional block, condition identical to
the build's, positioned after it), and mutation-tested: moving it back, diverging the two conditions, and
moving a history-dependent check into the block each fail it.

While fixing that I also found the placement test's slice ran past the step boundary into the rest of the
job, so it could not have failed. Replaced with a proper single-step extractor — otherwise this PR would
have shipped a test with exactly the "looks like a guard" defect it is meant to prevent.

Re-verified: **26,746 passed / 0 failed**, typecheck and actionlint clean, guard reports *all 46 local-gate
drift check(s) also run in a workflow*.

Separately: `Workers Builds: loopover-ui` is red here and also on #10252, and this PR touches no UI source
(only `ci.yml`, `package.json`, `scripts/**`, `test/**`). Watching whether the re-run clears it rather than
assuming.

